PR #9, opened 2025-04-08
Plotly Integration

jblake42:
This PR builds on the work done by Rui to integrate plotly into the UI service. The following changes were made:

- Seperated out the zone and disruption providers to make tools more generic
- Moved back to D3 drag handling
- Removed specific element references from providers to ensure linked plots still works
- Seperated out graph creation from tools to make tools more reusable
- Standardised context provider initialisation with data
- Added context menu to each provider to handle tool editing
- Added context menu to main plot to handle adding zones/disruptions

Through this work an issue with plotly has been found that means that the library requires dynamic loading. This leads to a slow first render of the plotly graphs, however subsequent graphs are quick to load

image


samueljackson92 (commented):
Looking real nice. Thanks Josh.

samueljackson92 (approved):

PR #13, opened 2025-04-25
Fix #11 – Eliminate duplicate zones/disruptions in dev mode

### LINKED ISSUE #11: Disruption table rendering duplicate values and "click to hide/show" not working
Disruption table is rendering all the zones in duplicate.

The "click" functionality is not working as well - if you intended to remove this just ignore this part.

https://github.com/user-attachments/assets/647765c8-7ad4-405e-9b2a-4143cff759c7

abdullah-ukaea:
This PR resolves the “Disruption table rendering duplicate values and click-to-hide not working” bug (#11).
The root cause was React 18 Strict-Mode mounting components twice during Fast-Refresh, causing our initialisation Effects to run twice and leaving extra Plotly elements on the page.

**Summary of changes:**

- React Strict-Mode mounted components twice, our effects added state on each mount and never cleaned up, so duplicates appeared on refresh/Fast-Refresh.

- zone-provider.tsx: initialise with zones.current = [...initialData]; clear in effect cleanup → no duplicate horizontal bars or table rows.

- vpsan-provider.tsx: same overwrite + cleanup pattern for disruption v-spans → no duplicate red lines/rows.

- disruption-plot.tsx: keep a plotElement reference, detach plotly_relayout, Plotly.purge(root), remove custom <g> overlay, and clean the context-menu listener on unmount.

**Result**: single set of tools after refresh, no leaked SVG groups or listeners, lower memory; production behaviour unchanged.

**Verification**: tested manual refresh and repeated Fast-Refresh on /disruption/26599; duplicates gone and console clean.

[issue11AFTER.webm](https://github.com/user-attachments/assets/a29147e1-bd9e-435a-a708-9e3b3bc328ed)

### DIFF
--- a/services/ui/src/app/components/providers/vpsan-provider.tsx
+++ b/services/ui/src/app/components/providers/vpsan-provider.tsx
@@ -72,13 +72,16 @@ export const VSpanProvider = ({categories, initialData, children} : {
 
     // Initialisation of data - this should only run once
     useEffect(() => {
-        if (!initialData) return
-
-        for (const span of initialData) {
-            spans.current.push(span)
-        }
-        triggerVSpanUpdate()
-    }, [initialData])
+        if (!initialData) return;
+    
+        spans.current = [...initialData]; 
+        triggerVSpanUpdate();
+    
+        /* cleanup runs when the first (discarded) mount unmounts */
+        return () => {
+          spans.current = [];
+        };
+      }, [initialData]);
 
     // Provides an array of the categories for the context menu
     const updateTypeItems = categories.map((category, index) => {
--- a/services/ui/src/app/disruption/components/disruption-plot.tsx
+++ b/services/ui/src/app/disruption/components/disruption-plot.tsx
@@ -127,8 +127,8 @@ export const DisruptionPlot = ({data, plotId: externalId, zoneCategories, disrup
         initGraph()
         
         return () => { // cleanup on unmount / Fast-Refresh
-            plotElement?.removeAllListeners?.("plotly_relayout") // detach relayout listener
-            ;(window as any).Plotly?.purge?.(root) //  fully purge plotly figure
+            plotElement?.removeAllListeners?.("plotly_relayout"); // detach relayout listener
+            (window as any).Plotly?.purge?.(root);                // fully purge plotly figure
             root?.querySelector(`.${plotId}-overplot`)?.remove() // remove custom overlay group
             setPlotReady(false) // reset ready state
         } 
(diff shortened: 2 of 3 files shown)


services/ui/src/app/disruption/components/disruption-plot.tsx
```diff
@@ -114,13 +118,20 @@ export const DisruptionPlot = ({data, plotId: externalId, zoneCategories, disrup
...
-                })
+                } 
+                plot.on("plotly_relayout", relayoutHandler) // attach listener so it can be removed
             })
         }
         initGraph()
         
+        return () => { // cleanup on unmount / Fast-Refresh
+            plotElement?.removeAllListeners?.("plotly_relayout") // detach relayout listener
+            ;(window as any).Plotly?.purge?.(root) //  fully purge plotly figure
```
jblake42, line 131: Please move the semi colon to the end of the line above

services/ui/src/app/components/providers/vpsan-provider.tsx
```diff
@@ -72,13 +72,16 @@ export const VSpanProvider = ({categories, initialData, children} : {
...
-        triggerVSpanUpdate()
-    }, [initialData])
+        if (!initialData) return;
+    
+        spans.current = [...initialData]; 
+        triggerVSpanUpdate();
+    
+        /* cleanup runs when the first (discarded) mount unmounts */
+        return () => {
+          spans.current = [];
```
jblake42, line 82: Since you have implemented the override rather than pushing new elements this may actually be unnecessary. I don't see a real efficiency gain to justify taking it out though

services/ui/src/app/disruption/components/disruption-plot.tsx
jblake42, line 131: Additionally, I don't believe this line currently does what you expect. If you log `window.Plotly` I think you'll find it is undefined. Please could you provide some evidence that the purge is needed - I believe overwritting the root is enough as I can't see duplicate plots being produced. If we do need to purge,  you will need to locally import the purge function similar to how it is done in line 101. Then `purge(root)` can be called directly. We just need to be careful of any resource penalties we get from doing this, so justification is needed

services/ui/src/app/disruption/components/disruption-plot.tsx
```diff
@@ -114,13 +118,20 @@ export const DisruptionPlot = ({data, plotId: externalId, zoneCategories, disrup
...
+            root?.querySelector(`.${plotId}-overplot`)?.remove() // remove custom overlay group
```
jblake42, line 132: I believe that if we carry out a purge properly, this is then not needed as it is removed in the purge. And if we decide that the purge is not needed then this will still be overwritten

jblake42 (commented):
This looks really good and fixes the issues highlighted in the related issue. There are some comments for you to review though, please reach out with any questions

services/ui/src/app/disruption/components/disruption-plot.tsx
abdullah-ukaea: I agree, I have removed the line 131 that does the purge plotly figure

abdullah-ukaea: I agree, I have fixed the semi colons on these lines

jblake42 (approved):

PR #22, opened 2025-05-12
Basic locked mode plot

rorlandoc:
Adds a spectrogram example for the locked mode analysis.

For the future we can look at:
- Letting user pick coloraxis vMin and vMax;
- Adding segmentation / thresholding capabilities;
- Linking a histogram plot to view data distribution


services/ui/src/app/locked-mode/components/locked-mode-plot.tsx
```diff
@@ -0,0 +1,457 @@
...
+        const vMax = 1E-3
+
+        const logvals = linspace(vMin, vMax, 10)
+        const logvalsMapped = logvals.map((x) => (9 / (Math.max(...logvals) - Math.min(...logvals))) * (x - Math.min(...logvals)) + 1.0)
+        const tickvals = logvalsMapped.map((x) => Math.log10(x))
+        console.log(logvals)
+        console.log(logvalsMapped)
+        console.log(tickvals)
+        console.log(logvals.map((x) => x.toExponential(1)))
+
```
jblake42, line 327: I'm not 100% on this maths for generating log scales, something seems odd but that is maybe better for someone more data focussed than me

services/ui/src/app/locked-mode/components/locked-mode-plot.tsx
```diff
@@ -0,0 +1,457 @@
...
+        const plotData: Plotly.Data[] = [{
+            name: "Saddle Coil FFT",
+            type: 'heatmap',
+            x: time.current,
+            y: freq.current,
+            z: zdata(),
+            customdata: ampl.current,
+            hovertemplate: "t: %{x:.2f}s<br>f: %{y:.2f}Hz<br>s: %{customdata:.2e}<extra></extra>",
+            coloraxis: 'coloraxis'
+        }];
```
jblake42, line 352: From the documentation have read the approach usually seems to be including `colorbar` and `colorscale` in the data and not the layout. This does seem to work despite the typing system complaining so happy for it to stay, out of curiosity are there any docs backing this approach up?

services/ui/src/app/locked-mode/components/locked-mode-plot.tsx
```diff
@@ -0,0 +1,457 @@
...
+            return
+        }
+
+        const dragElement = plot.querySelector(".drag")
+
+        if (!dragElement) {
+            console.error("Could not locate drag element to assign context menu")
+            return
+        }
+
```
jblake42, line 446: It seems you have included the context menu provider higher up in the DOM and have inlcuded this useEffect, however have not fully implemented the code needed to generate the context menu used to add zones/vspans. It should either be implemented, otherwise I don't think you need the useEffect (or even the context menu provider)

services/ui/src/app/locked-mode/components/locked-mode.tsx
```diff
@@ -0,0 +1,47 @@
...
+    ]
+
+    return (
+        <div className="flex flex-col items-center space-y-3">
+            <header className="p-6">
+                <h1 className="text-4xl font-bold text-center text-gray-900">
+                    Locked Mode demo
+                </h1>
+            </header>
+            <ContextMenuProvider menuId="locked-mode-menu">
```
jblake42, line 37: As mentioned, this may not be needed if you do not want the context menu on the plot itself?

jblake42 (commented):
I have added my reviews focussing on the UI implementation, I have not looked as much into the backend as this is more suited to Sam. All in all, it looks good, it's nice to see the use of the standardised tools (hopefully these inetgrated pretty smoothly). My concerns focus mainly on the log scale stuff, but this is probably more specific to the data science side which I am not as confident on

services/ui/src/app/locked-mode/components/locked-mode-plot.tsx
```diff
@@ -0,0 +1,457 @@
...
+import { VSpans } from "@/app/components/tools/vspans"
+import { Zones } from "@/app/components/tools/zones"
+
+type LockedModePlotProps = {
+    plotId?: string;
+    data: SpectrogramData;
+    lockedModeCategory: Category;
+}
+
+const civids = (x: number) => {
```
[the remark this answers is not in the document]
rorlandoc: Found out D3 has a built in function for this - used that instead.

services/ui/src/app/locked-mode/components/locked-mode-plot.tsx
rorlandoc: For reference this is what the math is doing:

1. Compute 10 linearly space values in desired range [vMin, vMax]:
```js
const logvals = linspace(range[0], range[1], 10)
```

2. Map ```logvals``` to [1, 10] range:
```js
const logvalsMapped = logvals.map((x) => (9 / (Math.max(...logvals) - Math.min(...logvals))) * (x - Math.min(...logvals)) + 1.0)
```
note that this is achieved by first getting the values in a [0, 1] range (by computing the ratio of the current element to the interval length), then multiplying by 9 to get [0, 9] and finally adding 1 to get the desired range. Getting the values in this range is important to ensure that log10 over the data will result in the data being within [0, 1] (because log10(1) = 0 and log10(10) = 1) thus using all the colour values in the coloraxis (min value - 0 - would be mapped to the first colour and max value - 1 - would be mapped to the last colour). 

3. Apply the log function 
```js
logvalsMapped.map((x) => Math.log10(x))
```

Hope this helps make the math clearer and the reasoning behind certain things. Now, this is a bit hacky in the sense that I am forcing 10 ticks / divisions in the coloraxis. This is not ideal but this was for proof of concept. Can and should be changed in later iterations.

PR #24, opened 2025-05-21
Subplotting

### LINKED ISSUE #16: Investiagte tooling support for plotly sub-plotting
Currently the exisiting D3 tooling does not fully support sub-plotting due to code like:
```
const marginTop = plot._fullLayout._size.t
const totalHeight = plot._fullLayout.yaxis._length
const y = marginTop - totalHeight
const height = 2 * totalHeight

const xaxis = plot._fullLayout.xaxis;
```

When using subplots there can multiple x and y axes with different names which will lead to issues.

We should investigate what is needed for our current uses and what is needed in the future to fully support sub-plotting:

- [ ] Create a plot with two sub plots in a column
- [ ] Link x-axis scrolling
- [ ] Add shared zoning to both subplots

[Subplotting docs](https://plotly.com/javascript/subplots/)

jblake42:
This PR relates to adding subplotting support to our eco-system.

### Rationale
The reason for this change relates to the fact that our initial tooling was designed to only support plots consisting of single subplots. This works well for individual traces, but falls apart when we want to start linking axes as plotly makes use of teh subplotting feature for this.

We would like tooling to be linked across subplots so all traces are correctly annotated.

### Added features
The current tooling has been improved to interrogate all plots they are assigned to in order to locate any subplots. It will then automatically render the tooling onto all subplots. This is useful as developers will not need to make any changes to code to allow subplotting other than adding additional traces to a plot, the tooling will automatically pick up the subplots.

A demo of subplotting has been added to a demo section in order to provide an example of how subplotting can be deployed.

### Future Improvements
In the future it would be useful if the plots were abstracted into a component which allows traces to be passed in, this reduces the need for developers to mess around with the plotting boilerplate just to add a new graph, especially now that there is additional subplotting logic in there as well. I cannot see this being too challenging.


samueljackson92 (changes requested):
Hi Josh, this looks good to me. However there is one small thing:

Now when I go to the `/disruption` page it does not draw the zones/lines straight away. I need to first drag the plot to make them render. 

image

To reproduce:

1. Go to the home screen `localhost:3001`
2. Choose disruption analysis
3. Enter shot `30421`

jblake42:
> Hi Josh, this looks good to me. However there is one small thing:
> 
> Now when I go to the `/disruption` page it does not draw the zones/lines straight away. I need to first drag the plot to make them render.
> image
> 
> To reproduce:
> 
>     1. Go to the home screen `localhost:3001`
> 
>     2. Choose disruption analysis
> 
>     3. Enter shot `30421`

@samueljackson92  I cannot reproduce this bug, when I follow the steps to reproduce the tooling shows.

jblake42:
> Hi Josh, this looks good to me. However there is one small thing:
> 
> Now when I go to the `/disruption` page it does not draw the zones/lines straight away. I need to first drag the plot to make them render.
> image
> 
> To reproduce:
> 
>     1. Go to the home screen `localhost:3001`
> 
>     2. Choose disruption analysis
> 
>     3. Enter shot `30421`

For some bizzare reason the range attribute of a graph does not initialize properly when the graph is first rendered (although it does work when refreshed) meaning all the tooling had almost zero height. I have found some replacement attributes that do consistently appear to initialize correctly. From testing the fix seems to have worked.

samueljackson92 (approved):
Looks good to me now!

PR #25, opened 2025-05-23
Define System Architecture & API

samueljackson92:


services/api/schemas/samples.py
```diff
@@ -0,0 +1,19 @@
...
+
+class FileSample(Sample):
+    file_name: str
+
+
+class ShotSample(Sample):
+    shot_id: int
+
+
+class UFOSample(Sample):
```
samueljackson92, line 17: I think this might be too specific.

I am thinking we have roughly two types of `Samples`:

 - Samples where they are names of files (e.g. CSV, png, parquet, mp4).
 - Samples where they index a shot number. (e.g. for UDA, SAL)

Additionally, I am wondering if we really need both a camera and a frame for UFOs? Wouldn't it make more sense to have a whole video, with all the annotated frames, and the video viewer just jumps to the first UFO?

services/api/schemas/events.py
```diff
@@ -0,0 +1,28 @@
...
+    validated: bool = False
+    
+class Point(Event):
+    time: int
+
+class Region(Event):
+    time_min: float
+    time_max: float
+    
+class UFO(Point):
```
samueljackson92, line 17: I think `UFO` `ELMRegion` and `ELM` are too specific. I think we need to stick to generic types, more like:

`TimeSeriesRegion`
`TimeSeriesPoint`
`BoundingBox`

Each may have a `label` associated with it, which could be flexible.

PR #28, opened 2025-06-06
Standard Time Series Component

jblake42:
### Rationale
This PR aims to ease the creation of time series graph through the creation of a standard component, removing the need to touch the plotly or D3 boilerplate.

### Features
This PR achieves the following:
- Creation of a component wrapper that abstracts the plot creation boilerplate from developers
- Creation of a graph requires only the plot configuration:
   - Plot data
   - Plot layout
   - Plot config (optional)
- Full automatic support for subplotting, with options exposed to control layout
- Hoisting of tooling child components outside of the plot component to ease automatic component generation in the future

### Imporvements
- It may be desired to simplify the plot configuration props by abstracting them further
- It would possibly be nice if the tooling children were not needed at all, however the feasibility of this is questionable


samueljackson92:
Looks good to me!

PR #35, opened 2025-06-19
Integrate Spectrogram View

samueljackson92:
This PR integrates the spectrogram viewer work by @rorlandoc into the new API and backend framework created by @wk9874.

This PR:
- Adds the ability to create "views" of data, in this case we create a spectrogram from a time series signal
  -  Views may take parameters, such as time, amplitude, and frequency range as well as number of points.
- Created toolbar for interactive amplitude selection
- Adds `MHD` as a new task which leads to the spectrogram view.
- Add the ability to save and load Zones

Follow up work with add the ability to save/load time points.

Screenshot 2025-06-19 at 09 53 10

### DIFF
--- a/scripts/setup.py
+++ b/scripts/setup.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from typing import Optional
 import requests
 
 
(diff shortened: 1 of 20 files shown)


samueljackson92:
@rorlandoc Sorry I broke the log scale as part of this PR. I've tried to re-enable it with them being log scaled based on the data distribution.

Yes, I swapped the colormap, but you are right that we should use a perceptually uniform one. Can we use `interpolatePlasma` instead?

samueljackson92:
Yes, I agree that it should be ultimately be a user choice. Sorry I just really didn't like the moss green-yellow, but wanted to pick something equivalent. 

I created a follow up issue for adding an option as you suggest: https://github.com/ukaea/viz-annotation/issues/37

rorlandoc (approved):

PR #45, opened 2025-06-20
Refactor & tidy up the front end

samueljackson92:
**Do not merge before #36**

This PR does not add any new functionality. Here we:

1. Rename some of the views
2. Refactor the toolbar component into a separate file
3. Add front end schemas matching the REST API.
4. Provide type annotations in quite a few places.


services/ui/src/app/components/peaks.tsx
jblake42, line 1: There are lots more type issue in here, especially `null` validation but I have already commented on this from memory

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const response = await fetch(ANNOTATIONS_URL, {
+        method: 'PUT',
+        headers: {
+        'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(annotations),
+    });
+}
+
+async function getNextSample(project_id: str) {
```
jblake42, line 19: Not python I'm afraid - `string`

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
+
+type AmplitudeSliderInfo = {
+  data: Data,
+  viewParams: any,
+  setViewParams: (viewParams: any) => void
+}
+
+export function AmplitudeSlider({data, viewParams, setViewParams}: AmplitudeSliderInfo) {
+    const onAmplitudeRangeChange = async (ampRange) => {
```
jblake42, line 75: Please type this argument

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,130 @@
...
+        viewParams.amplitude_min = Math.pow(10, ampRange.start);
+        viewParams.amplitude_max = Math.pow(10, ampRange.end);
+        setViewParams(viewParams);
+    };
+
+    let ampValues = data.amplitude.flat();
```
jblake42, line 81: As before, it is not guarenteed that amplitude is defined in the schema. You need to validate it is there for this to be type safe

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,130 @@
...
+    ampValues = ampValues.map(x => Math.log10(Math.max(x, 1e-6)));
+    const ampRangeTool = (
+        <DataRangeSlider name={'Amplitude Range'} data={ampValues} onChange={onAmplitudeRangeChange}
```
jblake42, line 84: Proper typing not present in `DataRangeSlider` - adding some might correct the missing typing for `getValueLabel` argument but if not you will need to explicitly type that

jblake42, line 84: From memory I may have already highlighted this but `Math.round` is used incorrectly

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
+export default function ToolBar({ project, sample, data, annotations, setAnnotations, viewParams, setViewParams} : ToolBarInfo) {
+  const project_id = project._id;
+  const sample_id = sample._id;
+
+
+  let tools = [];
+  if (project.task == 'ELM') {
+    const findPeaksTool = (
+        <FindPeaksTool project_id={project_id} sample_id={sample_id} data={data} setAnnotations={setAnnotations}></FindPeaksTool>
```
jblake42, line 107: The `setAnnotations` function is not the same type as the prop requires - the argument type is different. I personally think this is an issue with the `FindPeaksTool` erronously requiring `Annotation` rather than `Annotations`

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,130 @@
...
+    );
+    tools.push(findPeaksTool); 
+  } else if (project.task == 'MHD') {
+    let mhdData = data.values['mirnov'];
```
jblake42, line 111: As above - need to validate this if we keep the schema the same

services/ui/src/app/disruption/components/disruption.tsx
```diff
@@ -9,34 +9,34 @@ import { TimeSeries } from "@/app/components/plots/time-series"
...
 
-type DisruptionInfo = {
-    time: Array<number>,
-    values: Record<string, Array<number>>
-}
-
-export const Disruption = ({ data, annotations, setAnnotations}: {data: DisruptionInfo}) => {
+export const DisruptionView = ({data, annotations, setAnnotations} : DisruptionViewInfo) => {
     const disruptionCategories: Category[] = [
             { name: "Disruption", color: 'rgb(255, 0, 0)' },
```
jblake42, line 20: `name` is not part of `Category`

services/ui/src/app/disruption/components/disruption.tsx
```diff
@@ -87,27 +87,27 @@ export const Disruption = ({ data, annotations, setAnnotations}: {data: Disrupti
...
-    const updateVSpans = (newVSpans) =>  {
+    const updateVSpans = (newVSpans: Array<VSpan>) =>  {
         let spans = newVSpans.map(item => ({
                 time: item.x,
                 label: item.category.name
         }));
 
 
-        const otherAnnotations = annotations.filter(x => !TimePoint.safeParse(x).success);
+        const otherAnnotations: Annotations = annotations.filter((x: Annotation) => !TimePointSchema.safeParse(x).success);
```
jblake42, line 110: Identical to other calls. Can we abstract?

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -1,13 +1,18 @@
 'use client'
+import { Annotations, MultiVariateTimeSeriesData, TimeRegion, Zone, Category } from "@/types"
 import { ZoneProvider } from "@/app/components/providers/zone-provider"
 import { ContextMenuProvider } from "@/app/components/providers/context-menu-provider"
 import { TimeSeries } from "@/app/components/plots/time-series"
 import { Zones } from "@/app/components/tools/zones"
 import 'react-contexify/ReactContexify.css';
 import Plotly from "plotly.js-dist";
```
jblake42, line 8: I commented this previously but can be try and be a bit more specific here please

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -145,23 +54,26 @@ async function getProject(project_id: string) {
...
     return await getData(`${process.env.NEXT_PUBLIC_API_URL}/backend-api/projects/${project_id}/samples/${sample_id}/annotations`);
 }
 
+type SamplePageInfo = {
+  params: { project_id: string, sample_id: string };
+};
 
-export default function SamplePage({ params }: Props) {
+export default function SamplePage({ params }: SamplePageInfo) {
   const props = use(params);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx:66]: Same comment about `use`

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -145,23 +54,26 @@ async function getProject(project_id: string) {
...
   const project_id = props.project_id;
   const sample_id = props.sample_id;
 
-  const [project, setProject] = useState<any>(null);
-  const [sample, setSample] = useState<any>(null);
-  const [data, setData] = useState<any>(null);
-  const [annotations, setAnnotations] = useState<any>([]);
-  const [viewParams, setViewParams] = useState<any>({name: 'identity'});
+  const [project, setProject] = useState<Project>(null);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx:70]: The typing also doesn't align with the setter calls - e.g `getProject` does not return an object with type `Project`

services/ui/src/app/spectrogram/components/spectrogram.tsx
```diff
@@ -20,26 +20,27 @@ const linspace = (start: number, end: number, num: number) => {
...
         { name: "LLM", color: 'rgb(200, 100, 100)' },
     ]
 
-    const convertRegionToZone = (item) => {
+    const convertRegionToZone = (item: TimeRegion) => {
         const category = zoneCategories.find(x => x.name === item.label);
         return {x0: item.time_min, x1: item.time_max, category: category};
     };
-    const zones = annotations.map(convertRegionToZone);
+    const zones: Zone = annotations.map(convertRegionToZone);
```
jblake42, line 43: There have been lots of this style code - same comment about the types not matching

jblake42, line 43: I also think these repeated conversions could be extracted into helper functions as it seems like we are repeating the same thing many times

services/ui/src/types.ts
```diff
@@ -1,47 +1,102 @@
...
+  query_strategy: z.string(),
+  data_loader: z.string()
+}); 
+export type Project = z.infer<typeof ProjectSchema>;
+
+export const SampleSchema = z.object({
+  _id: z.string(),
+  project_id: z.string(),
+  shot_id: z.number(),
+  data: z.record(z.any())
```
jblake42, line 81: As above

jblake42 (changes requested):
This review is a first pass mainly focussing on typing. I am not confident with these PRs going into main personally. My suggestions would be to merge the other two PRs into this one once you have worked through them and then I will do another pass. I have left a lot of comments out of this PR as I think I have covered it in other PRs - I don't want them to slip through though

services/ui/src/app/disruption/components/disruption.tsx
```diff
@@ -9,34 +9,34 @@ import { TimeSeries } from "@/app/components/plots/time-series"
...
             { name: "Disruption", color: 'rgb(255, 0, 0)' },
         ]
 
-    const convertTimePointToVSpan = (item) => {
+    const convertTimePointToVSpan = (item: TimePoint) => {
         const category = disruptionCategories.find(x => x.name === item.label);
         return {x: item.time, category: category};
     };
-    const spans = annotations.filter(x => TimePoint.safeParse(x).success).map(convertTimePointToVSpan);
+    const spans = annotations.filter((x: Annotation) => TimePointSchema.safeParse(x).success).map(convertTimePointToVSpan);
```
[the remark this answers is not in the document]
samueljackson92: I think it is? But the type of `annotations` is incorrect.

samueljackson92:
@jblake42 I've had a go at addressing the issues you have highlighted. Please have a re-review and let me know if there are any further issues.

services/ui/src/app/components/peaks.tsx
```diff
@@ -1,31 +1,42 @@
...
+    const signalOptions = Object.keys(data.values).map((value, index)=> ({id: index, name: value}));
+
+    const clearPeaks = () => {
+        setAnnotations([]);
+    };
 
     useEffect(() => {
-        if (data) {
-            const time = data.values['ip'].time;
+        if (data && (signalName in data.values)) {
```
jblake42: `signalName` needs validating that it is not `null`

services/ui/src/app/components/peaks.tsx
```diff
@@ -1,31 +1,42 @@
...
             setTimeMaxDefault(tmax);
         }
-    }, [data]);
+    }, [data, signalName]);
 
     useEffect(() => {
         const fetchData = async () => {
-            if (clearPeaks) {
-                setAnnotations([]);
+            if (signalName == null && !(signalName in data.values)) {
```
jblake42: Should this be `||` -  it doesn't make sense to me as `&&`

services/ui/src/app/components/peaks.tsx
```diff
@@ -47,18 +59,24 @@ export function FindPeaksTool({ project_id, sample_id, data, setAnnotations }) {
         };
 
         fetchData();
-    }, [prominence, distance, clearPeaks, timeRange]);
+    }, [prominence, distance, timeRange, signalName]);
```
jblake42: We are missing quite a few dependencies here including `data.values` and `project_id` - it is a code smell not to include them

services/ui/src/app/components/peaks.tsx
```diff
@@ -47,18 +59,24 @@ export function FindPeaksTool({ project_id, sample_id, data, setAnnotations }) {
...
+
 
 
     return (
         <Provider theme={defaultTheme}>
             <Header>Find Peaks</Header>
             <div className='m-4'>
             <Flex direction="column">
+                <ComboBox defaultItems={signalOptions} onInputChange={setSignalName} allowsEmptyCollection={true} placeholder="None selected">
```
jblake42: `placeholder` is deprecated due to accessibility issues - we should really use the help text below but I am happy for it to stay if needed

jblake42: From my investigation it doesn't look like allowsEmptyCollection does anything - going by the definition of ComboBox the typing doesn't support it. Could you link the docs that show what the use case is please, I couldn't find it

services/ui/src/app/components/tools/dataRangeSlider.tsx
```diff
@@ -1,8 +1,16 @@
 "use client";
-import { RangeSlider} from '@adobe/react-spectrum'
+import { ViewParams } from '@/types';
+import { RangeSlider, RangeValue } from '@adobe/react-spectrum'
```
jblake42: `RangeValue` is not exported in this package - we should probably install the `@react-types/shared` package to get the proper typing

services/ui/src/app/components/tools/dataRangeSlider.tsx
```diff
@@ -1,8 +1,16 @@
...
 import { useEffect, useState } from 'react';
 
-export function DataRangeSlider({ name, data, onChange, getValueLabel}) {
+type DataRangeSliderInfo = {
+  name: string,
+  data: Array<number>,
+  onChange: (params: ViewParams) => void,
```
jblake42: The typing of this does not align with what `RangeSlider` is expecting in the `onChnageEnd` prop - I saw `NaN` show on the slider which I think was caused by this

services/ui/src/app/components/tools/dataRangeSlider.tsx
```diff
@@ -1,8 +1,16 @@
...
+  getValueLabel: (param: RangeValue<number>) => string
+}
+
+export function DataRangeSlider({ name, data, onChange, getValueLabel}: DataRangeSliderInfo) {
     const [timeMinDefault, setTimeMinDefault] = useState(null);
```
jblake42: These values are used as props in `RangeSlider` but they could be null causing typing errors and possible other issues - we should provide a default value or use `start: timeMinDefault || 0` in case it is `null`

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
+"use client";
+import { useRouter } from 'next/navigation';
+import {Provider, defaultTheme,  ButtonGroup, ToastQueue, Button } from '@adobe/react-spectrum'
+import { Annotation, Annotations, CompositeDataSchema, Data, MultiVariateTimeSeriesDataSchema, Project, Sample, SpectrogramData, SpectrogramDataSchema, SpectrogramViewParams, ViewParams } from "@/types";
+import { FindPeaksTool } from '@/app/components/peaks';
+import { DataRangeSlider } from '@/app/components/tools/dataRangeSlider';
+
+async function saveAnnotations(project_id: string, sample_id: string, annotations: Annotations) {
+    const ANNOTATIONS_URL = `${process.env.NEXT_PUBLIC_API_URL}/backend-api/projects/${project_id}/samples/${sample_id}/annotations`;
+    const response = await fetch(ANNOTATIONS_URL, {
```
jblake42: We are not using the response so there is no need to assign it to a variable

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+  project_id: string
+  sample_id: string
+  annotations: Annotation
+};
+export function NextButton({project_id, sample_id, annotations} : NextButtonInfo) {
+  const router = useRouter();
+
+  const handleClick = async () => {
+    try {
+      await saveAnnotations(project_id, sample_id, annotations);
```
jblake42, line 58: `annotations` is the wrong type - I think in `NextButtonInfo` the type should be `Annotations` rather than `Annotation`

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+      router.push(NEXT_SAMPLE_URL);
+    } catch (err) {
+      console.error('Failed to fetch data:', err);
+    }
+  };
+
+  return <Button variant="primary" onPress={handleClick} >Next</Button>
+}
+
+type SaveButtonInfo = {
```
jblake42: It seems a bit pointless defining these types when they are identical - also there is the same issue with the `annotations` typing

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+  project_id: string
+  sample_id: string
+  annotations: Annotation
+};
+export function SaveButton({project_id, sample_id, annotations}: SaveButtonInfo) {
+  const handleClick = async () => {
+    try {
+      await saveAnnotations(project_id, sample_id, annotations);
+      ToastQueue.positive(`Saved ${annotations.length} annotations!`, {timeout: 5000})
```
jblake42: `length` property does not exist but again I think this is because it should be `Annotations`

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+      console.error('Failed to fetch data:', err);
+    }
+  };
+
+  return <Button variant="primary" onPress={handleClick} >Save</Button>
+}
+
+type AmplitudeSliderInfo = {
+  data: SpectrogramData,
+  viewParams: any,
```
jblake42: I'd prefer us not to use `any` is our type definitions if we can avoid it - bit of a code smell. If we can't know the type it is better to use `unknown`

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+  setViewParams: (viewParams: any) => void
+}
+
+export function AmplitudeSlider({data, viewParams, setViewParams}: AmplitudeSliderInfo) {
+    const onAmplitudeRangeChange = async (ampRange: SpectrogramViewParams) => {
+        viewParams.amplitude_min = Math.pow(10, ampRange.start);
```
jblake42: `start` and `end` don't exist in `SpectogramViewParams`?

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+        viewParams.amplitude_max = Math.pow(10, ampRange.end);
+        setViewParams(viewParams);
+    };
+
+    let ampValues = data.amplitude.flat();
+    ampValues = ampValues.map((x: number) => Math.log10(Math.max(x, 1e-6)));
+    const ampRangeTool = (
+        <DataRangeSlider name={'Amplitude Range'} data={ampValues} onChange={onAmplitudeRangeChange} 
+        getValueLabel={val => `${Math.round(Math.pow(10, val.start)*10000)/10000} - ${Math.round(Math.pow(10, val.end)*10000)/10000}`}/>
```
jblake42: I would personally extract this rounding function which is used twice into a utility function to make it clearer what they are doing at a glance

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+  annotations: Annotations,
+  setAnnotations: (annotations: Annotations) => void,
+  viewParams: ViewParams,
+  setViewParams: (viewParams: ViewParams) => void
+}
+export default function ToolBar({ project, sample, data, annotations, setAnnotations, viewParams, setViewParams} : ToolBarInfo) {
+  const project_id = project._id;
+  const sample_id = sample._id;
+
+  let tools = [];
```
jblake42: Setting a type for this would be nice if possible, even if it is just a React component type - also it can be `const` as pushing does not count as a modification

jblake42: What is the reason for tools being an array? As far as I can see only one tool can be assigned due to `else` statement

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+  if (project.task == 'ELM') {
+    data = MultiVariateTimeSeriesDataSchema.parse(data);
+    const findPeaksTool = (
+        <FindPeaksTool project_id={project_id} sample_id={sample_id} data={data} setAnnotations={setAnnotations}></FindPeaksTool>
+    );
+    tools.push(findPeaksTool); 
+  } else if (project.task == 'MHD') {
+    data = CompositeDataSchema.parse(data)
+    let mhdData = SpectrogramDataSchema.parse(data.values['mirnov']);
```
jblake42: `const` here

jblake42: `values` is not guarenteed to be in this type so this is dangerous

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  let tools = [];
+  if (project.task == 'ELM') {
+    data = MultiVariateTimeSeriesDataSchema.parse(data);
+    const findPeaksTool = (
+        <FindPeaksTool project_id={project_id} sample_id={sample_id} data={data} setAnnotations={setAnnotations}></FindPeaksTool>
+    );
+    tools.push(findPeaksTool); 
+  } else if (project.task == 'MHD') {
+    data = CompositeDataSchema.parse(data)
```
jblake42: I don't think this schema is right - the union means that you cannot parse `data` using this

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const ampRangeTool = <AmplitudeSlider data={mhdData} viewParams={viewParams} setViewParams={setViewParams}/>
+    tools.push(ampRangeTool);
+  }
+
+  return (
+        <Provider theme={defaultTheme}>
+        <div className='h-screen text-center'>
+          <div className='p-4'>
+            <ButtonGroup>
+              <SaveButton project_id={project_id} sample_id={sample_id} annotations={annotations}/>
```
jblake42: Same `Annotation` type issue

jblake42: And the line below

services/ui/src/app/components/tools/vspans.tsx
```diff
@@ -88,6 +88,8 @@ export const VSpans = ({plotId, plotReady, forceUpdate} : ToolingProps) => {
                     const x = xaxis.p2d(newX); // The context provider stores the decimal value rather than pixel
                     d.x = x;
                     handleVSpanUpdate() // Global refresh must be triggered to update all linked plots
+                }).on("end", function(event, d) {
+                    handleVSpanDragFinish();
```
jblake42, line 92: I think I have changed my mind - the drag finish event is probably a better way to handle this than debouncing

services/ui/src/app/disruption/components/disruption.tsx
```diff
@@ -1,38 +1,57 @@
...
+  acc[curr.name] = curr.color;
+  return acc;
+}, {});
 
-    const initialDisruption: VSpan[] = [
-        { x: 0.3, category: disruptionCategories[0] }
-    ]
+const colorMapping = { ...disruptionCategoryColors, ...zoneCategoryColors };
+
+type DisruptionViewInfo = {
```
jblake42, line 54: It could be beneficial for us to create some generic types or even templated types if all that chnages is the data type - we are getting alot of repetition of similar types

services/ui/src/app/disruption/components/disruption.tsx
```diff
@@ -1,38 +1,57 @@
...
+  data: MultiVariateTimeSeriesData,
+  annotations: Annotations,
+  setAnnotations: (annotations: Annotations) => void
+};
+
+
+export const DisruptionView = ({data, annotations, setAnnotations} : DisruptionViewInfo) => {
+    const convertAnnotationToDisplayAnnotation = createAnnotationToDisplayAnnotationFunc(colorMapping);
+    const displayAnnotations: DisplayAnnotation[] = annotations.map(convertAnnotationToDisplayAnnotation);
+    const zones: Zone[] = displayAnnotations.filter((x: DisplayAnnotation) => ZoneSchema.safeParse(x).success);
```
jblake42: These safe parse checks are still not enough to ensure type safety as they can be zone or vspan data which differ

jblake42: These functions could also be something we look to pull out into (templated) utility functions

services/ui/src/app/disruption/components/disruption.tsx
```diff
@@ -83,11 +102,12 @@ export const Disruption = ({ data }: DisruptionInfo) => {
...
+    
     return (
         <div className="flex flex-col items-center space-y-3">
             <ContextMenuProvider menuId="disruption-menu">
-                <VSpanProvider categories={disruptionCategories} initialData={initialDisruption}>
-                    <ZoneProvider categories={zoneCategories} initialData={initialZones}>
+                <VSpanProvider categories={disruptionCategories} initialData={vspans} onModifyVSpan={updateVSpans}>
+                    <ZoneProvider categories={zoneCategories} initialData={zones} onModifyZone={updateZones}>
                         <TimeSeries plotId="Disruption" plotConfig={{data: plotData, layout: plotLayout}}>
                             <Zones />
```
jblake42: We are missing the `onZoneUpdate` props for both tools - this prop is not optional (and should also be renamed as it is used across multiple tools, not just zones)

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -1,13 +1,37 @@
 'use client'
+import { MultiVariateTimeSeriesData, Zone, Category, DisplayAnnotation, ZoneSchema, TimeRegionSchema, Annotation } from "@/types"
 import { ZoneProvider } from "@/app/components/providers/zone-provider"
 import { ContextMenuProvider } from "@/app/components/providers/context-menu-provider"
 import { TimeSeries } from "@/app/components/plots/time-series"
 import { Zones } from "@/app/components/tools/zones"
 import 'react-contexify/ReactContexify.css';
 import Plotly from "plotly.js-dist";
```
jblake42: Comment as before

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -1,13 +1,37 @@
...
+type ELMViewInfo = {
+    data: MultiVariateTimeSeriesData, 
+    annotations: Annotation, 
+    setAnnotations: (annotations: Annotation) => void
+};
+
+export const ELMView = ({data, annotations, setAnnotations}: ELMViewInfo) => {
+    const convertAnnotationToDisplayAnnotation = createAnnotationToDisplayAnnotationFunc(zoneCategoryColors);
+    const displayAnnotations: DisplayAnnotation[] = annotations.map(convertAnnotationToDisplayAnnotation);
+    const zones: Zone[] = displayAnnotations.filter((x: DisplayAnnotation) => ZoneSchema.safeParse(x).success);
```
jblake42: Similar comment to before regarding both making this a util function and also the safe parsing does not provide adequate type safety

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -1,13 +1,37 @@
...
+
+type ELMViewInfo = {
+    data: MultiVariateTimeSeriesData, 
+    annotations: Annotation, 
+    setAnnotations: (annotations: Annotation) => void
+};
+
+export const ELMView = ({data, annotations, setAnnotations}: ELMViewInfo) => {
+    const convertAnnotationToDisplayAnnotation = createAnnotationToDisplayAnnotationFunc(zoneCategoryColors);
+    const displayAnnotations: DisplayAnnotation[] = annotations.map(convertAnnotationToDisplayAnnotation);
```
jblake42: `annotations` is not an array, how are we mapping it? Clean up typing of `ELMViewInfo` and this will be resolved

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -1,13 +1,37 @@
...
+    { name: "ELM", color: 'rgb(233, 170, 98)' },
+    { name: "H-Mode", color: 'rgb(100, 170, 98)' },
+]
+
+const zoneCategoryColors = zoneCategories.reduce<Record<string, string>>((acc, curr) => {
+  acc[curr.name] = curr.color;
+  return acc;
+}, {});
+
+type ELMViewInfo = {
```
jblake42, line 36: Same comment regarding generic types

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -1,13 +1,37 @@
...
+    setAnnotations: (annotations: Annotation) => void
+};
+
+export const ELMView = ({data, annotations, setAnnotations}: ELMViewInfo) => {
+    const convertAnnotationToDisplayAnnotation = createAnnotationToDisplayAnnotationFunc(zoneCategoryColors);
+    const displayAnnotations: DisplayAnnotation[] = annotations.map(convertAnnotationToDisplayAnnotation);
+    const zones: Zone[] = displayAnnotations.filter((x: DisplayAnnotation) => ZoneSchema.safeParse(x).success);
+
+    const updateZones = (newZones: Array<Zone>) => {
+        updateAnnotations(setAnnotations, newZones, TimeRegionSchema);
```
jblake42: `setAnnotations` does not accept the correct type - should be fixed as above

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -1,13 +1,37 @@
...
+    }
 
     var dataTrace = {
```
jblake42: Is there a reason all of these are `var` - if there isn't please use `const`

jblake42: We should also type properly:
```
const dataTrace: Partial<Plotly.PlotData> = {
        name: 'Dalpha',
        x: data.values.dalpha.time,
        y: data.values.dalpha.values,
        mode: 'lines',
    };
```

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -168,27 +192,12 @@ export const ElmGraph = ({data, annotations, setAnnotations}) => {
...
-        setAnnotations(zones);
-    }
 
     return (
         <div className="flex space-y-3">
             <div className="flex-1 text-center items-center">
                 <ContextMenuProvider menuId="elm-menu">
-                    <ZoneProvider categories={zoneCategories} initialData={zones} onAddZone={updateAnnotations}>
+                    <ZoneProvider categories={zoneCategories} initialData={zones} onModifyZone={updateZones}>
                         <TimeSeries plotId="ELMs" plotConfig={{data: plotData, layout: plotLayout}}>
```
jblake42: The layout passed as a prop has typing issues - I beleive all that is needed is to use `Partial<Plotly.Layout>` when you define the layout above

services/ui/src/app/elms/components/elms.tsx
```diff
@@ -168,27 +192,12 @@ export const ElmGraph = ({data, annotations, setAnnotations}) => {
...
                             <Zones />
```
jblake42: `onZoneUpdate`

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -14,11 +14,11 @@ export const SampleBreadCrumbs = (info) => {
...
 
-type SampleTableParams = {
-  project_id: string;
+type SampleTableInfo = {
+  project_id: Promise<string>;
 };
 
-export const SamplesTable = ({project_id}: SampleTableParams) => {
+export const SamplesTable = ({project_id}: SampleTableInfo) => {
   const samples = getSamples(project_id);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx:40]: `project_id` is now a promise so this is incorrect

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -53,11 +53,11 @@ export const SamplesTable = ({project_id}: SampleTableParams) => {
...
 }
 
-type Props = {
+type ProjectViewInfo = {
   params: { project_id: string };
 };
 
-export default function ProjectView({params} : Props) {
+export default function ProjectView({params} : ProjectViewInfo) {
   const project_id = use(params).project_id;
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx]: This still isn't right as far as I can see. `project_id` is not a promise - You appear to have changed it in a different type definition but not here (perhaps these should be a common type)?

Also I think it should be `use(params.project_id)` although I haven't really ever used `use`

services/ui/src/app/projects/[project_id]/page.tsx
Comment (jblake42) [reply services/ui/src/app/projects/[project_id]/page.tsx:40]: In the same vein, when `project_id` is passed to `SamplesTable` below it is a promise not a string

services/ui/src/app/projects/[project_id]/page.tsx
Comment (jblake42) [reply services/ui/src/app/projects/[project_id]/page.tsx]: Reading through again I think maybe the `SamplesTable` prop needs to be a `string` and the `ProjectView` needs to be a promise. I think that would work everything out

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
-  } else if (args.project.task == 'ELM') {
-    return (<ElmGraph data={args.data} annotations={args.annotations} setAnnotations={args.setAnnotations} />);
-  } else if (args.project.task == 'MHD') {
-    return (<LockedMode data={args.data.values['mirnov']} viewParams={args.viewParams} annotations={args.annotations} setAnnotations={args.setAnnotations}/>);
-  }
-}
-
-type Props = {
-  params: { project_id: string, sample_id: string };
+type SampleViewInfo = {
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx:55]: Maybe a generic type?

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
-        <FindPeaksTool project_id={project_id} sample_id={sample_id} data={data} setAnnotations={setAnnotations}></FindPeaksTool>
-    );
-    tools.push(findPeaksTool); 
+const SampleView = ({project, data, annotations, setAnnotations}: SampleViewInfo) => {
+  if (project.task == 'disruption') {
+    data = MultiVariateTimeSeriesDataSchema.parse(data);
+    return (<DisruptionView data={data} annotations={annotations} setAnnotations={setAnnotations} />);
+  } else if (project.task == 'ELM') {
+    data = MultiVariateTimeSeriesDataSchema.parse(data);
+    return (<ELMView data={data} annotations={annotations} setAnnotations={setAnnotations} />);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: `Annotation` type issues again

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
-    };
-
-    let ampValues = mhdData.amplitude.flat();
-    ampValues = ampValues.map(x => Math.log10(Math.max(x, 1e-6)));
-    const ampRangeTool = (
-        <DataRangeSlider name={'Amplitude Range'} data={ampValues} onChange={onAmplitudeRangeChange} 
-        getValueLabel={val => `${Math.round(Math.pow(10, val.start)*10000, 2)/10000} - ${Math.round(Math.pow(10, val.end)*10000, 2)/10000}`}/>
-    );
-    tools.push(ampRangeTool);
+    data = CompositeDataSchema.parse(data)
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: Similar `CompositeSchema` parsing issues as before here

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
+    let mhdData = SpectrogramDataSchema.parse(data.values['mirnov']);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: const

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
 
-export async function getData(url) {
+export async function getData(url: string): Promise<Data> {
     const response = await fetch(url);
     const payload = await response.json();
     return payload;
 }
 
-async function getSample(project_id: string, sample_id: string) {
+async function getSample(project_id: string, sample_id: string): Promise<Sample> {
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: This function states it returns a `Promise<Sample>` yet it actually appears to return `Data`
The issue is two fold:
- The type of promise being returned `Data` vs `Sample`
- You should not return the result of an `await` if you say you are returning a promise - unless the promise is returned by another async function but that would be very complex

Comment (jblake42) [reply services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: This is the case for the other similar functions

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
-  const [project, setProject] = useState<any>(null);
-  const [sample, setSample] = useState<any>(null);
-  const [data, setData] = useState<any>(null);
-  const [annotations, setAnnotations] = useState<any>([]);
-  const [viewParams, setViewParams] = useState<any>({name: 'identity'});
+  const [project, setProject] = useState<Project | null>(null);
+  const [sample, setSample] = useState<Sample | null>(null);
+  const [data, setData] = useState<Data | null>(null);
+  const [annotations, setAnnotations] = useState<Annotations>([]);
+  const [viewParams, setViewParams] = useState<ViewParams>({name: 'identity'});
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: The set function is never used

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
 
-  const refreshData = async ( viewParams ) => {
+  const refreshData = async ( viewParams: ViewParams ) => {
     const project = await getProject(project_id);
     setProject(project);
 
     const sample = await getSample(project_id, sample_id);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: You have a naming clash here

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -1,175 +1,91 @@
...
     setSample(sample);
 
-    const annotations = await getAnnotations(project_id, sample_id);
-    setAnnotations(annotations);
+    const dbAnnotations = await getAnnotations(project_id, sample_id);
+    setAnnotations(dbAnnotations);
     
     if (project.task == 'MHD') {
       viewParams.name = 'spectrogram';
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: The line below this refers to `nperseg` this doesn't exist in the type

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -204,7 +120,7 @@ export default function SamplePage({ params }: Props) {
         <ToastContainer placement="top" />
         <SampleDataBreadCrumbs project={project} sample={sample}></SampleDataBreadCrumbs>
           <div className='flex'>
-            <ToolBar project={project} sample_id={sample_id} data={data} annotations={annotations} setAnnotations={setAnnotations} viewParams={viewParams} setViewParams={refreshData}/>
+            <ToolBar project={project} sample={sample} data={data} annotations={annotations} setAnnotations={setAnnotations} viewParams={viewParams} setViewParams={refreshData}/>
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: `project` and `sample` are the wrong types in every case in the return statement - from a cursory glance it looks like a lack of checking for `null`

services/ui/src/app/spectrogram/components/spectrogram.tsx
```diff
@@ -1,15 +1,15 @@
...
 import { ContextMenuProvider } from "@/app/components/providers/context-menu-provider"
 import { ZoneProvider } from "@/app/components/providers/zone-provider"
 import { TimeSeries } from "@/app/components/plots/time-series"
 import { Zones } from "@/app/components/tools/zones"
 import { VSpans } from "@/app/components/tools/vspans"
-
 import * as d3 from "d3"
+import { createAnnotationToDisplayAnnotationFunc, updateAnnotations } from "@/app/utils"
 
 const linspace = (start: number, end: number, num: number) => {
```
jblake42: Could certainly be a util - I assume we'd use this else where

services/ui/src/app/spectrogram/components/spectrogram.tsx
```diff
@@ -20,32 +20,58 @@ const linspace = (start: number, end: number, num: number) => {
...
+const lockedModeCategories: Category[] = [
+    { name: "Locked Mode", color: "rgb(255, 0, 0)" },
+]
+
+const zoneCategories: Category[] = [
+    { name: "NTM", color: 'rgb(0, 255, 255)' },
+    { name: "LLM", color: 'rgb(200, 100, 100)' },
+]
+
+const zoneCategoryColors = zoneCategories.reduce<Record<string, string>>((acc, curr) => {
```
jblake42: We have done this sort of thing in other places - maybe we abstract

services/ui/src/app/spectrogram/components/spectrogram.tsx
```diff
@@ -20,32 +20,58 @@ const linspace = (start: number, end: number, num: number) => {
...
+type SpectrogramViewInfo = {
+    data: SpectrogramData, 
+    annotations: Annotations,
+    setAnnotations: (annotations: Annotations) => void
+};
+
+export const SpectrogramView = ({data, annotations, setAnnotations}: SpectrogramViewInfo) => {
+    const convertAnnotationToDisplayAnnotation = createAnnotationToDisplayAnnotationFunc(colorMapping);
+    const displayAnnotations: DisplayAnnotation[] = annotations.map(convertAnnotationToDisplayAnnotation);
+    const zones: Zone[] = displayAnnotations.filter((x: DisplayAnnotation) => ZoneSchema.safeParse(x).success);
```
jblake42: I have commented a similar thing in the other place this is done - type error

services/ui/src/app/spectrogram/components/spectrogram.tsx
```diff
@@ -109,26 +135,17 @@ export const LockedMode = ({ data, annotations, setAnnotations }: {data: LockedM
         modeBarButtonsToRemove: ['pan'],
```
jblake42: `pan` is not valid - `pan2d` is

services/ui/src/app/spectrogram/components/spectrogram.tsx
```diff
@@ -109,26 +135,17 @@ export const LockedMode = ({ data, annotations, setAnnotations }: {data: LockedM
...
 
     return (
         <div className="flex flex-col items-center space-y-3">
             <ContextMenuProvider menuId="locked-mode-menu">
-                <VSpanProvider categories={lockedModeCategories} initialData={initialLockedMode}>
-                    <ZoneProvider categories={zoneCategories} initialData={zones} onModifyZone={updateAnnotations}>
+                <VSpanProvider categories={lockedModeCategories} initialData={vspans} onModifyVSpan={updateVSpans}>
+                    <ZoneProvider categories={zoneCategories} initialData={zones} onModifyZone={updateZones}>
                         <TimeSeries plotId="LockedMode" plotConfig={{ data: plotData, config: plotConfig, layout: plotLayout }} >
                             <Zones />
```
jblake42: `onZoneUpdate`

services/ui/src/app/utils.tsx
```diff
@@ -0,0 +1,56 @@
+import { ZodSchema } from "zod/v4";
+import { DisplayAnnotation, Annotation, Annotations, TimeRegionSchema, ZoneSchema, VSpanSchema, TimePointSchema, VSpan, Zone, TimePoint, TimeRegion} from "@/types"
+
+export const convertDisplayAnnotationToAnnotation = (annotation: DisplayAnnotation): Annotation => {
```
jblake42: It would be beneficial to provide a brief comment before the function to give some tooltip context - the name is fairly explanatory but I still had to dive into the implementation. This is the case for the other util functions, we should be adding some usage instructions imo

services/ui/src/app/utils.tsx
```diff
@@ -0,0 +1,56 @@
...
+            };
+            return vspan;
+        } else {
+            throw new Error("Unsupported annotation type");
+        }
+    };
+    return convertAnnotationToDisplayAnnotation;
+}
+export function updateAnnotations<T> (setAnnotations: (annotations: Annotations) => void, newDisplayAnnotations: DisplayAnnotation[], schema: ZodSchema<T>): void {
+    setAnnotations((prevAnnotations: Annotations) => {
```
jblake42: I assume this is another `Annotation` type issue

services/ui/src/types.ts
```diff
@@ -1,30 +1,113 @@
...
+  query_strategy: z.string(),
+  data_loader: z.string()
+}); 
+export type Project = z.infer<typeof ProjectSchema>;
+
+export const SampleSchema = z.object({
+  _id: z.string(),
+  project_id: z.string(),
+  shot_id: z.number(),
+  data: z.record(DataSchema)
```
jblake42: This is not correct usage of `record`

services/ui/src/types.ts
jblake42, line 1: Whilst there are no specific comments in this file really, I think quite a few comments I have left may be caused by changes needed in the `zod` interface definitions

jblake42 (changes requested):
This still needs work before merging in my opinion

jblake42:
I have push a small commit with some of the small refactoring changes I made as I went. Feel free to review and revert if you are not happy

services/ui/src/app/components/tools/toolbar.tsx
samueljackson92: Because in future there will be many tools. This is the case in downstream PRs.

services/ui/src/app/components/tools/toolbar.tsx
samueljackson92: Not sure what you mean here. Why is it invalid?

services/ui/src/app/disruption/components/disruption.tsx
samueljackson92: Yes. But leave it for now. I think the repetition is not too bad, and this can be cleaned up later as we slowly make things more abstract.

services/ui/src/app/disruption/components/disruption.tsx
samueljackson92: Could be. For now I've move them to a utility place, as it is likely they are modified a bit in upcoming work.

services/ui/src/app/elms/components/elms.tsx
samueljackson92: Response as above.

services/ui/src/app/elms/components/elms.tsx
samueljackson92: For now I want to keep it separate, until we are a bit more stable.

samueljackson92:
@jblake42 Ok, I've updated with fixes for a lot of what you've suggested above, and managed to catch a lot of small issues this time. Can you please re-review?

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -130,7 +130,7 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
 
             registerMenuItem("zone", menuElement)
 
-        }, [categories, registerMenuItem])
+        }, [addZone, categories, registerMenuItem])
```
jblake42: Adding this as a dependency may cause a re-render loop - `addZone` should be defined inside the useEffect which shouldn't be an issue as this is the only place it is used

services/ui/src/app/components/tools/dataRangeSlider.tsx
```diff
@@ -1,35 +1,50 @@
 "use client";
-import { ViewParams } from '@/types';
-import { RangeSlider, RangeValue } from '@adobe/react-spectrum'
-import { useEffect, useState } from 'react';
+import { ViewParams } from "@/types";
```
jblake42: Unused - added to commit

services/ui/src/app/components/tools/dataRangeSlider.tsx
```diff
@@ -1,35 +1,50 @@
...
-            let tmax = Math.max(...time); 
-            setTimeMinDefault(tmin);
-            setTimeMaxDefault(tmax);
-            setTimeRange({start: tmin, end: tmax});
-        }
-    }, [data]);
+  useEffect(() => {
+    if (data && (timeMinDefault === null || timeMinDefault === null)) {
+      const time = data;
+      let tmin = Math.min(...time);
```
jblake42: Can be `const` - added to commit

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -1,176 +1,188 @@
...
+    });
+  }, [
+    handleZoneUpdate,
+    plotId,
+    plotReady,
+    showZoneMenu,
+    zones,
+    triggerUpdate,
+    forceUpdate,
+  ]);
```
jblake42: Missing dependnecy - added in commit

services/ui/src/app/components/peaks.tsx
```diff
@@ -1,84 +1,137 @@
...
-                    signal_name: signalName,
-                    prominence: prominence,
-                    distance: distance,
-                    time_min: timeRange.start,
-                    time_max: timeRange.end,
-                }),
-            });
+  useEffect(() => {
+    const fetchData = async () => {
+      if (signalName === null || !(signalName in data.values)) {
```
jblake42: Using `data.values` here is not ideal as we then need to include it as a dependency which causes other issues. In general we should be limiting what happens in the `useEffect` - in this case `const validSignal = signalName !== null && (signalName in data.values)` can be written outside the effect and then the `validSignal` status can be used inside the effect (with the proper dependencies). This is also the case for the effect above this. Using this approach both ensures we fully form out dependencies and prvents additional rerenders which cause us issues.

jblake42: I have included my proposal in the commit for reference

services/ui/src/app/core.tsx
jblake42, line 1: This file was missing a lot of `useEffect` deps - added in commit

jblake42, line 1: All the functions in here have hooks in. We can only use hooks inside components or custom hooks - these functions are currently neither. I think the best approach is just to change the functions to start with `use` which then means their correct usage is enforced in other files e.g if you try to use one of these funcitons outside a component it will complain.

services/ui/src/app/disruption/components/disruption.tsx
jblake42: Ok, that's fine

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -86,46 +168,68 @@ export default function SamplePage({ params }: SamplePageInfo) {
...
   };
 
   useEffect(() => {
     const run = async () => {
       await refreshData(viewParams);
-    }
+    };
     run();
-  }, [viewParams]);
+  }, []);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: There are missing params here - it looks like you removed at least one as it was causing issues. This is a code smell and indicates we should be doing something different here. The easy way to is to have the initialisation `useEffect` to directly include all the initialisation code and use a seperate callback for other stuff

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
Comment (jblake42) [reply services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx:55]: Based on previous comments abstraction is being pushed back

jblake42 (changes requested):
This looks so much better than before. I have left some comments in places I have directly tweaked bits as they were minor changes - please review these and check you are happy. I have a couple more minor points which I would like to be cleared up before merging if possible. It looks really good though.

samueljackson92:
@jblake42 thanks for updating on the small things. All looks good to me. I have updated with changes for the last few bits that you did not solve.

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -130,7 +131,7 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
 
             registerMenuItem("zone", menuElement)
 
-        }, [addZone, categories, registerMenuItem])
+        }, [categories, registerMenuItem])
```
jblake42: This introduced a missing dependency - I have added it and checked it functions ok but please also review

services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx
```diff
@@ -157,46 +157,49 @@ export default function SamplePage({
...
+
+    const run = async (viewParams: ViewParams) => {
       await refreshData(viewParams);
     };
-    run();
-  }, []);
+
+    run(viewParams);
+
+  }, [viewParams]);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/samples/[sample_id]/page.tsx]: Same as previous comment

jblake42 (approved):
Once you review the small changes I have made I am happy for this to be merged

PR #49, opened 2025-06-23
Refactor Annotators into their own endpoint

wk9874:
* Moves code to run FindPeaks annotator on a sample into `/projects/{project_id}/samples/{sample_id}/annotators/find_peaks`
* Generalises code so that it is not specific to Find Peaks
* Create mapping between Task type and Annotator. This may not be the best solution, and will need changing when we move past MVP and into the ability to setup their projects with more granular control. But for now I think it is tied to task type and not to eg data type (since not all Time Series will need a find peaks method)
* Removed `services/core/api/elms` - I copied this over when initially drafting the endpoints, but no longer needed as stuff has been split out
* UI homepage no longer points to ELMs, Disruptions etc since these dont exist - goes to Projects instead
* Commented out `torch` as a requirement - just to speed up installation while we dont have models implemented

Ready for review, set to Draft as this should not be merged until https://github.com/ukaea/viz-annotation/pull/45 is merged


samueljackson92 (approved):
Code looks good. Tested changes local and all still works 👍

samueljackson92:
@wk9874 fix the merge conflicts here and this is ready to go.

PR #50, opened 2025-06-23
Add pagination and sorting to Projects and Samples tables

### LINKED ISSUE #40: Add paginated project/sample views
Matt has already created endpoints which can return a limited number items. (see http://localhost:8002/docs#/Projects/get_projects_projects_get)

Instead of having one giant list of samples/projects it would be better if they could be paginated showing a limited number per page

wk9874:
* Adds 'Previous' and 'Next' buttons to the project and samples table, which returns the next subset of projects to the user
* Adds 'projects per page' dropdown so user can choose how many items to view at a time
* Change endpoints which use pagination from `end` to `count` - makes it easier for the front end to use
* Added 'sort_by' and 'sort_direction' to API get all endpoints
* Added sorting to the table columns, which then triggers a callback to make the relevant request to the server

also changed eg `getProjects` to `useGetProjects` to solve an issue I was seeing:
```
  20 |
  21 | export const getSamples = (project_id: string, page: number, projectsPerPage: number) => {
> 22 |   const [samples, setSamples] = useState<any>(null);
     |                                               ^
  23 |
  24 |   useEffect(() => {
  25 |     const fetchData = async () => {

React has detected a change in the order of Hooks called by ProjectView. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://react.dev/link/rules-of-hooks
```

But I'm not sure why this is necessary or whether this is the appropriate fix (it's on the advice of Mr GPT) so please review and let me know if theres a better way to fix that...

Ready for review, set to Draft as https://github.com/ukaea/viz-annotation/pull/45 must be merged first

Closes #40


wk9874:
Whoops forgot about the samples page - will add that too as part of this PR

samueljackson92 (changes requested):
Couple of comments on UI side code.

Additionally, I think you should add some margin/padding to the new controls so they are not flush against the edge of the window table:

image

samueljackson92:
Adding @jblake42  and @abdullah-ukaea  as reviewers for the front end as they may have more comments.

services/ui/src/app/core.tsx
```diff
@@ -18,23 +18,23 @@ export const getURL = (url: string) => {
   return data;
 }
 
-export const getSamples = (project_id: string) => {
+export const useGetSamples = (sortDescriptor, project_id: string, page: number, samplesPerPage: number) => {
```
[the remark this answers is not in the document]
wk9874: Hopefully all fixed, please take a look @samueljackson92 :)

services/ui/src/app/projects/page.tsx
```diff
@@ -51,6 +57,18 @@ export const ProjectsTable = () => {
 }
 
 export default function Projects() {
+  const [projectsPerPage, setProjectsPerPage] = useState<number>(5);
+  const [currentPage, setCurrentPage] = useState<number>(1);
+  const [sortDescriptor, setSortDescriptor] = useState<SortDescriptor<Project>>({ column: '_id', direction: 'descending' });
```
samueljackson92, line 62: 👍

samueljackson92 (approved):
Looks good to me!

services/ui/src/app/core.tsx
```diff
@@ -18,18 +19,18 @@ export const getURL = (url: string) => {
   return data;
 }
 
-export const getSamples = (project_id: string) => {
+export const getSamples = (sortDescriptor: SortDescriptor<Sample>, project_id: string, page: number, samplesPerPage: number): Sample[] | null => {
```
jblake42: `SortDescriptor` is not defined as a generic yet it is used as one here. As far as I can see the interface just exposes a direction and column property - what is the reason for the `Sample` generic type?

services/ui/src/app/core.tsx
```diff
@@ -50,23 +51,23 @@ export const getSample = (project_id: string, sample_id: string) => {
   return sample;
 } 
 
-export const getProjects = (project_id: string) => {
+export const getProjects = (sortDescriptor: SortDescriptor<Project>, page: number, projectsPerPage: number): Project[] | null => {
```
jblake42: As above

services/ui/src/app/projects/[project_id]/page.tsx
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx:1]: Just so I don't keep leaving the same comments - as far as I can see there is no reason to treat `SortDescriptor` as a generic

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -73,7 +90,26 @@ export default function ProjectView({params} : ProjectViewInfo) {
...
-        <SamplesTable project_id={project_id}></SamplesTable>
+        <SamplesTable project_id={project_id} samples={samples} sortDescriptor={sortDescriptor} onSortChange={onSortChange}></SamplesTable>
+        <Provider theme={defaultTheme}>
+          <div className="flex items-center justify-between pl-4 pr-4">
+            <Button variant="primary" onPress={() => setCurrentPage((p) => p - 1)} isDisabled={currentPage === 1}>
+              Previous
+            </Button>
+            <div className="flex items-center justify-center gap-8 pb-2">
+              <p> Page: {currentPage} </p>
+            <Picker label="Samples per Page:" onSelectionChange={(selected) => {setSamplesPerPage(selected); setCurrentPage(1)}} defaultSelectedKey="5">
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx]: `selected` needs to be checked for a `null` value

services/ui/src/app/projects/page.tsx
jblake42, line 1: Same comment regarding generic

services/ui/src/app/projects/page.tsx
```diff
@@ -1,6 +1,15 @@
 "use client";
 import { getProjects } from '@/app/core';
-import {Provider, defaultTheme, Cell, Column, Row, TableView, TableBody, TableHeader, Breadcrumbs, Item} from '@adobe/react-spectrum'
+import { use, useState, useEffect } from 'react';
```
jblake42: `use` and `useEffect` have been added but not used

services/ui/src/app/projects/page.tsx
```diff
@@ -60,7 +78,26 @@ export default function Projects() {
...
-          <ProjectsTable></ProjectsTable>
+          <ProjectsTable projects={projects} sortDescriptor={sortDescriptor} onSortChange={onSortChange}></ProjectsTable>
+          <Provider theme={defaultTheme}>
+            <div className="flex items-center justify-between pl-4 pr-4">
+              <Button variant="primary" onPress={() => setCurrentPage((p) => p - 1)} isDisabled={currentPage === 1}>
+                Previous
+              </Button>
+              <div className="flex items-center justify-center gap-8 pb-2">
+                <p> Page: {currentPage} </p>
+              <Picker label="Projects per Page:" onSelectionChange={(selected) => {setProjectsPerPage(selected); setCurrentPage(1)}} defaultSelectedKey="5">
```
jblake42: Please perform `null` validation

jblake42 (changes requested):
Looks really good and works well - just a few tweaks suggested

wk9874:
Have also found that the table massively overfills its box when you have 25 projects/samples per page:

image


Could either have the upper limit be 10 projects/samples per page (but that seems very low), or can find a way to make the table scroll

services/ui/src/app/core.tsx
wk9874: Agreed, not sure where I got this from... fixed by removing <Sample> or <Project> from the type definition

services/ui/src/app/projects/[project_id]/page.tsx
Comment (wk9874) [reply services/ui/src/app/projects/[project_id]/page.tsx]: Done

wk9874:
Re the above bug, React Spectrum claims that fixing the headers inplace and scrolling the entries should be default behaviour. Have tried adding a Flex container as described in https://react-spectrum.adobe.com/react-spectrum/TableView.html#layout and adding tailwind CSS to a div to fix the height as described in https://stackoverflow.com/questions/69897289/how-to-make-react-table-to-scrollable but no luck...

jblake42:
I didn't have any problems when using this code:
```
      <Flex height="size-5000" width="100%" direction="column">
        <TableView
        flex
        selectionMode="none"
        selectionStyle="highlight"
        sortDescriptor={sortDescriptor}
        onSortChange={onSortChange}
        >
          <TableHeader>
            <Column key="name" allowsSorting>Name</Column>
            <Column key="task" allowsSorting>Task</Column>
            <Column key="_id" allowsSorting>Date Created</Column>
            <Column key="data_loader" allowsSorting>Loader</Column>
          </TableHeader>
          <TableBody items={rows}>
            {item => (
              <Row href={`${process.env.NEXT_PUBLIC_API_URL}/projects/${item['id']}`}>
                <Cell>{item['name']}</Cell>
                <Cell>{item['task']}</Cell>
                <Cell>{item['timestamp']}</Cell>
                <Cell>{item['data_loader']}</Cell>
              </Row>
            )}
          </TableBody>
        </TableView>
      </Flex>
    </Provider>
```
Although I did notice it doesn't appear to work if you put it at a higher level in the component tree


https://github.com/user-attachments/assets/2f7cead6-380b-4616-9ba4-b7122de3c58f

(Apologies for the recording I didn't want to redo it when I realised OBS was set up wrong)

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -98,7 +98,15 @@ export default function ProjectView({params} : ProjectViewInfo) {
             </Button>
             <div className="flex items-center justify-center gap-8 pb-2">
               <p> Page: {currentPage} </p>
-            <Picker label="Samples per Page:" onSelectionChange={(selected) => {setSamplesPerPage(selected); setCurrentPage(1)}} defaultSelectedKey="5">
+            <Picker 
+              label="Samples per Page:" 
+              onSelectionChange={(selected) => {
+                if (selected != null) {
+                  setSamplesPerPage(selected);
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx]: The null check clears this up - I didn't go into depth enough on my intial check though. `Key` is a `string | number` type whereas `setSamplesPerPage` expects `number` only

services/ui/src/app/projects/page.tsx
```diff
@@ -86,7 +86,15 @@ export default function Projects() {
               </Button>
               <div className="flex items-center justify-center gap-8 pb-2">
                 <p> Page: {currentPage} </p>
-              <Picker label="Projects per Page:" onSelectionChange={(selected) => {setProjectsPerPage(selected); setCurrentPage(1)}} defaultSelectedKey="5">
+              <Picker 
+              label="Projects per Page:" 
+              onSelectionChange={(selected) => {
+                if (selected != null) {
+                  setProjectsPerPage(selected);
```
jblake42: As above

jblake42 (changes requested):
Last comments ... I promise ... maybe

services/ui/src/app/projects/[project_id]/page.tsx
Comment (wk9874) [reply services/ui/src/app/projects/[project_id]/page.tsx]: oh gotcha, didn't realise 'selected' was the key and not the item. So should I do: `setSamplesPerPage(Number(selected));`? Or is there a way for the key to just be a number, same as the item?

Comment (wk9874) [reply services/ui/src/app/projects/[project_id]/page.tsx]: (will also change `selected` to `selectedKey` to avoid getting confusd in the future)

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -114,9 +117,9 @@ export default function ProjectView({
               <p> Page: {currentPage} </p>
             <Picker 
               label="Samples per Page:" 
-              onSelectionChange={(selected) => {
-                if (selected != null) {
-                  setSamplesPerPage(selected); 
+              onSelectionChange={(selectedKey) => {
+                if (selectedKey != null) {
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx:142]: I would perhaps put a check like: `typeof(selectedKey) === "number"` in the if statement rather than just blindly casting as this could lead to issues if we aren't careful

services/ui/src/app/projects/page.tsx
```diff
@@ -89,9 +92,9 @@ export default function Projects() {
                 <p> Page: {currentPage} </p>
               <Picker 
               label="Projects per Page:" 
-              onSelectionChange={(selected) => {
-                if (selected != null) {
-                  setProjectsPerPage(selected); 
+              onSelectionChange={(selectedKey) => {
+                if (selectedKey != null) {
+                  setProjectsPerPage(Number(selectedKey));
```
jblake42: As above

services/ui/src/app/projects/[project_id]/page.tsx
Comment (jblake42) [reply services/ui/src/app/projects/[project_id]/page.tsx:142]: With this check carried out, type narrowing will automatically happen within the if statement body

jblake42 (approved):
Once the last comment has been reviewed I am happy for this to be merged

services/ui/src/app/projects/[project_id]/page.tsx
Comment (wk9874) [reply services/ui/src/app/projects/[project_id]/page.tsx:142]: But isn't the issue that the selectedKey is a string, not a number?

Or do you mean check `typeof(selectedKey) === "string"`?

Comment (jblake42) [reply services/ui/src/app/projects/[project_id]/page.tsx:142]: I see, apologies. From looking at the code I assumed we were always getting an actual number back. We can go with this approach then, however we still need to be safe about it. Either the result of the cast needs to be checked for `NaN` or a default value should be given using `||`

Comment (wk9874) [reply services/ui/src/app/projects/[project_id]/page.tsx:142]: Done using ||

PR #52, opened 2025-06-26
Collapse single-category Add VSpan submenu 

abdullah-ukaea:
This change updates VSpanProvider to address issue https://github.com/ukaea/viz-annotation/issues/19 so that when only one V-Span category exists (e.g. “Disruption”) the context-menu registers a single top-level Item (“Add Disruption”) instead of a Submenu with a lone child. The provider now checks categories.length; if it equals 1 it renders the direct item, otherwise it falls back to the original submenu logic for multi-category tools. This removes an unnecessary click for users without altering behaviour where multiple V-Span types are present. No other components are touched and the API for registering V-Spans remains unchanged.

### DIFF
--- a/services/ui/src/app/components/providers/vpsan-provider.tsx
+++ b/services/ui/src/app/components/providers/vpsan-provider.tsx
@@ -93,11 +93,29 @@ export const VSpanProvider = ({categories, initialData, children} : {
             )
         })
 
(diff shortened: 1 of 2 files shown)


jblake42 (approved):
Looks great and tested well

PR #53, opened 2025-06-27
Add search functionality

### LINKED ISSUE #43: Add ability to jump to a shot number
Previously we could jump to a particular shot number. It would be nice if we could bring this back in to the tool bar on each sample view.
It would also be useful if we could add the same input field to the page listing samples.

wk9874:
Adds search bars to Project, Sample pages and Toolbar

* Adds new `name` query parameter to `/projects` endpoint which allows you to filter projects returned by their name. This is done with regex and is case insensitive, so that eg 'local' would return 'Local ELM Project' and 'Local Disruption Project'
* Adds new `shot_id` query parameter to `/samples` endpoint, which allows you to filter returned samples by the shot ID they relate to. This is exact matches only since shot_id is stored as an integer in mongodb, so regex not possible (and probably not desired here)
* Adds search bar to Projects table which allows you to search by project name, displaying the resulting projects found which start with the string entered in the table
* Adds search bar to Samples table which allows you to search by shot ID, returning you the matching shot (if found) in the table. Includes validation so that only valid numbers can be entered. Had to do this manually because could not get the validation built into the React Spectrum widget to work properly :(
* Adds search bar to toolbar allowing you to jump straight to a given shot ID. Includes the same validation as above to check it is a number. If a sample with that shot ID is not found, prints an error and doesnt change page. If it is found, saves current annotations and then changes to new sample page

Closes #43 

Ready for review but set to WIP as this should not be merged before #50


services/api/routers/projects.py
```diff
@@ -33,18 +33,31 @@ async def get_projects(
...
 ) -> list[Project]:
     """
     Get a list of all available projects.
     -------------------------------------
     """
+    filters = {}
+    if name:
+        # Search with regex, return any projects which start with the searched for string, case insensitive
+        filters['name'] = {
+            "$regex": f"^{name}",
```
samueljackson92, line 53: Maybe this should not require the "^" in the regex?

samueljackson92 (changes requested):
1. Searching by project name only matches if the string begins with the search. E.g. Searching for `UDA` with return `UDA Project` but searching for `Project` will not.

2. I would add a little extra padding between the new component and the buttons here.
image

services/api/routers/projects.py
wk9874: I did this intentionally because I personally thought it would make sense for the search to only give you projects which start with whatever you typed in, but happy to be talked around the other way! Will probably mention at the meeting on Monday and see what the UI guys think would feel the most natural for users

samueljackson92: No, usually search is a "contains" rather than "starts with". Imagine if google only returned "starts with" by default.

jblake42: I agree it seems reasonable to be a contains - for example if people wanted to view all projects relating to JET or disruptions

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -65,6 +72,41 @@ export function SaveButton({project_id, sample_id, annotations}: SaveButtonInfo)
...
+      }
+    } else {
+      setErrorMessage("Please enter a number.");
+    }
+    }
+
+  return  <SearchField 
+            label="Jump to Shot" 
+            onSubmit={onSearchSubmit}
+            validationState={errorMessage ? 'invalid' : ''}
```
jblake42: This should not be an empty string - it should be `undefined` I beleive

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -1,8 +1,8 @@
 "use client";
 import { use, useState } from 'react';
 import { getSamples, getProject } from '@/app/core';
-import {Provider, defaultTheme, Cell, Column, Row, TableView, TableBody, TableHeader, Breadcrumbs, Button, Picker, Item} from '@adobe/react-spectrum'
-import { SortDescriptor } from '@react-types/shared';
+import {Provider, defaultTheme, Cell, Column, Row, TableView, TableBody, TableHeader, Breadcrumbs, Button, Picker, Item, SearchField, Form} from '@adobe/react-spectrum'
+import { SortDescriptor, ValidationDetails, ValidationResult } from '@react-types/shared';
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx]: The types you have added aren't being used?

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -90,8 +102,18 @@ export default function ProjectView({params} : ProjectViewInfo) {
...
         </h1>
-        <SamplesTable project_id={project_id} samples={samples} sortDescriptor={sortDescriptor} onSortChange={onSortChange}></SamplesTable>
         <Provider theme={defaultTheme}>
+        <div className="pl-4">
+          <SearchField 
+          label="Search By Shot ID" 
+          // SearchField should be able to do validation when provided a 'pattern' inside a Form element
+          // But I could not for the life of me get that to work, so will do it manually...
+          onSubmit={onSearchSubmit}
+          validationState={errorMessage ? 'invalid' : ''}
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx]: Same comment

services/ui/src/app/projects/[project_id]/page.tsx
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx:1]: Generics comment again

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -90,8 +102,18 @@ export default function ProjectView({params} : ProjectViewInfo) {
         <h1 className="text-2xl font-bold mb-4">
           Samples
         </h1>
-        <SamplesTable project_id={project_id} samples={samples} sortDescriptor={sortDescriptor} onSortChange={onSortChange}></SamplesTable>
         <Provider theme={defaultTheme}>
+        <div className="pl-4">
+          <SearchField 
+          label="Search By Shot ID"
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx:125]: While your making changes could this be changed to a self closed tag as there is nothing between the tags

services/ui/src/app/projects/[project_id]/page.tsx
```diff
@@ -1,8 +1,8 @@
 "use client";
 import { use, useState } from 'react';
 import { getSamples, getProject } from '@/app/core';
-import {Provider, defaultTheme, Cell, Column, Row, TableView, TableBody, TableHeader, Breadcrumbs, Button, Picker, Item} from '@adobe/react-spectrum'
-import { SortDescriptor } from '@react-types/shared';
+import {Provider, defaultTheme, Cell, Column, Row, TableView, TableBody, TableHeader, Breadcrumbs, Button, Picker, Item, SearchField, Form} from '@adobe/react-spectrum'
```
Comment (jblake42) [inline services/ui/src/app/projects/[project_id]/page.tsx]: Form is not being used

services/ui/src/app/projects/page.tsx
jblake42, line 1: Last generics comment for good luck

services/ui/src/app/projects/page.tsx
```diff
@@ -78,8 +79,11 @@ export default function Projects() {
           <h1 className="text-2xl font-bold mb-4">
             Projects
           </h1>
-          <ProjectsTable projects={projects} sortDescriptor={sortDescriptor} onSortChange={onSortChange}></ProjectsTable>
           <Provider theme={defaultTheme}>
+          <div className="pl-4">
+            <SearchField label="Search By Name" onSubmit={(name) => {setProjectName(name); setCurrentPage(1)}}></SearchField>
```
jblake42: `null` check

jblake42 (changes requested):
Looks good, only some small comments

services/ui/src/app/projects/[project_id]/page.tsx
Comment (wk9874) [reply services/ui/src/app/projects/[project_id]/page.tsx]: Believe this is left over from my failed attempts to use React Spectrum's built in search field validation - have removed

jblake42:
The only issues I see in this relate to the null checks for the previous PR - LGTM

samueljackson92:
@wk9874 again, fix the merge conflicts and this seems ready to go.

PR #56, opened 2025-06-30
Fix missing context menu handlers

jblake42:
It was noticed that there appears to have been a regression in the context menu handling across subplotting.

The issue was that the context menu handler was not being assigned to all subplots - I am unsure why this was working before to be honest.

The solution was to query all selectors with the required class rather than just the first one.


samueljackson92 (approved):
Changes work as expected. Code looks good to me.

PR #57, opened 2025-06-30
Auto scaling

### LINKED ISSUE #29: Auto scale y-range for plots on zoom
I think we should add the ability for plots to be autoscaled on the y axis when the user zooms in. So, when the range of y values is very small, the y axis auto scales to match the range of the data being viewed.

This is apparently not currently directly possible in plotly:
https://community.plotly.com/t/autorescaling-y-axis-range-when-range-slider-used/12214

But can possibly be achieved by using `relayout`:
https://stackoverflow.com/questions/75876460/plotly-autorescaling-y-axis-range-when-range-slider-used

jblake42:
Closes #29 

When zooming in on a plot, it has been identified that it is beneficial for the Y-Scale of the graph to automatically zoom to fit the data shown in the currently selected x-range as demonstrated in the clip below:

https://github.com/user-attachments/assets/4ffb5353-0eca-4b01-b43f-6468fcf38936

This works for subplotting with shared x-axes, a potential issue is that this is designed only for shared x-axes.

Additionally it should be noted that auto-scaling will not occur if it would lead to the y-scale being tighter than the user selection. This is to allow users to accurately target data.

### DIFF
--- a/services/ui/src/app/components/plots/time-series.tsx
+++ b/services/ui/src/app/components/plots/time-series.tsx
@@ -108,13 +108,17 @@ export const TimeSeries = ({
                     }
                 }
 
+                const xArray = (dataSet as PlotData).x as number[];
+                const yArray = (dataSet as PlotData).y as number[];
+
                 // Find min and max y data values
                 const yValues: number[] = [];
-                ((dataSet as PlotData).x as number[]).forEach((xVal, i) => {
+                for (let i = 0; i < xArray.length; i++) {
+                    const xVal = xArray[i];
                     if (xVal >= x0 && xVal <= x1) {
-                        yValues.push(((data[index] as PlotData).y as number[])[i])
+                        yValues.push(yArray[i]);
                     }
-                })
+                }
 
                 if (yValues.length > 0) {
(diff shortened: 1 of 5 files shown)


samueljackson92 (changes requested):
Generally, this look good to me and the functionality matches what I would expect.

One thing I've noticed though is that it seems to have a slight lag when scaling. I am not sure if it is possible to do any further optimisation to avoid the slight but noticeable lag?

jblake42:
> Generally, this look good to me and the functionality matches what I would expect.
> 
> One thing I've noticed though is that it seems to have a slight lag when scaling. I am not sure if it is possible to do any further optimisation to avoid the slight but noticeable lag?

I have investigated this and the lag appears to come from the plotly relayout call when adjusting the y-axis ranges. I have included some execution time analysis which shows ~98% of the lag time comes from plotly specific code, my min/max calcs only account for roughly 7-10ms when there are 5 subplots.


image

I personally do not know how I can reduce this as we need these seperate calls to resize the axes

samueljackson92:
As discussed, this is as far as we can go with plotly.

PR #58, opened 2025-07-03
Updating existing tooling to make use of generic context menu props

abdullah-ukaea:
This PR removes the last dependency on the legacy x0/x1 context-menu props and makes all tooling rely exclusively on the new generic payload. In time-series.tsx the menu callback now sends only generic data, adding xScale while dropping x0/x1. In vspan-provider.tsx the two call-sites that previously read props.x0 were switched to props.x, so v-span creation still lands at the click position. The main re-work is in zone-provider.tsx: a lightweight addFromClick helper derives x0 and a 100-px-wide x1 from props.x and props.xScale, all menu items now invoke this helper, and add(props.x0, props.x1, …) has been removed. Existing drag, delete and type-change logic is untouched, so behaviour is unchanged. This should address issue https://github.com/ukaea/viz-annotation/issues/26

### DIFF
--- a/services/ui/src/app/components/plots/time-series.tsx
+++ b/services/ui/src/app/components/plots/time-series.tsx
@@ -198,7 +198,7 @@ export const TimeSeries = ({
                 event,
                 props: {
                     // new generic props 
-                    xScale, yScale, 
+                    xScale
                     x, y,   // generic data-space click position
                     xRange, yRange,  // current axis spans
                     xLimits: [xMin, xMax], yLimits: [yMin, yMax]  // explicit axis limits
--- a/services/ui/src/app/components/plots/time-series.tsx
+++ b/services/ui/src/app/components/plots/time-series.tsx
@@ -162,15 +162,11 @@ export const TimeSeries = ({
         /* 
         Context-menu dispatcher
 
-            1. Converts the mouse click (pixel-space) to data-space coordinates (x, y) using Plotly’s axis converters.
+            Converts the mouse click (pixel-space) to data-space coordinates (x, y) using Plotly’s axis converters.
 
-            2. Derives the current axis ranges (xRange, yRange) so tools can size new elements as a fraction of the view, independent of zoom level.
+            Derives the current axis ranges (xRange, yRange) so tools can size new elements as a fraction of the view, independent of zoom level.
     
-            3. Computes a 100-pixel-wide default window (x0, x1) to keep legacy menu actions working 
-
-            information delivered to the menu is  { x, y, x0, x1, xRange, yRange, xLimits: [xMin, xMax], yLimits: [yMin, yMax] }
-
-            xScale, yScale, relX, relY are not exposed because current tooling must remain plot-agnostic and operate purely in data coordinates.
+            information delivered to the menu is  { x, y, xScale, yScale, xRange, yRange, xLimits: [xMin, xMax], yLimits: [yMin, yMax] }
 
         */
         function handleContextMenu(event: MouseEvent, plot) {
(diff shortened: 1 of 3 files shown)


services/ui/src/app/components/plots/time-series.tsx
```diff
@@ -197,15 +193,12 @@ export const TimeSeries = ({
...
-            const x0 = x   // left edge of default window
-            const x1 = x + unitWidth    // right edge of default window
 
             showContextMenuRef.current({
                 event,
                 props: {
-                    // backwards compatible props
-                    x0, x1,   // legacy 100 px helpers
                     // new generic props 
+                    xScale,
```
jblake42, line 201: We decided against using this due to the issue with large axis ranges causing issues - the axis range / min and max should be used instead. Please remove this prop and also the scale calculations in the handler further up

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -88,21 +88,22 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
...
-        }
+        /**
+         * Converts generic context-menu props into a new zone and
+         * re-creates the legacy “100 px default width” behaviour.
+         */
+        const addFromClick = (
+            props: { x: number; xScale: number },
+            category: Category,
+            ) => {
+            const width = 100 / props.xScale; // 100 px → data-units
```
jblake42, line 99: As stated above, we decided not to use the scaling, please make use of the range data provided - e.g. make the zone create with a width of 5% or 10% of the x-range - whatever looks right

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -88,21 +88,22 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
...
+            addZone(props.x, props.x + width, category)
+            }
     
             const addZoneItems = categories.map((category, index) => {
                 return (
                     <Item key={`add${index}`} id={`add${index}`} onClick={({props}) => {
-                        add(props.x0, props.x1, category)
+                        addFromClick(props as any, category)
```
jblake42, line 106: I think we need to try and be a bit more type safe here. The props supplied by the handler are not the same as that required by `addFromClick` I personally think you need to pull out the required values from the props that are passed in (validating they are there) and then pass these into the function you have made. Whenever we force things through with an `any` type assertion it is a bit of a code smell

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -117,7 +118,7 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
                 categories.length === 1
                     ? (
                         <Item key="add-zone-single" id="add-zone-single" onClick={({props}) => {
-                            add(props.x0, props.x1, categories[0])
+                            addFromClick(props as any, categories[0])
```
jblake42, line 121: As above

jblake42 (changes requested):

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -88,21 +88,23 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
...
+            const x0 = menu.x
+            const x1 = Math.min(x0 + width, menu.xLimits[1]) // clamp to upper limit
+            addZone(x0, x1, category)
+            }
     
             const addZoneItems = categories.map((category, index) => {
                 return (
                     <Item key={`add${index}`} id={`add${index}`} onClick={({props}) => {
-                        add(props.x0, props.x1, category)
+                        addFromClick(props as MenuProps, category)
```
jblake42, line 107: I would have liked to see a it more validation as we are getting an input of type `any` and type casting that. However I am happy that at the moment it is guarenteed that these properties will be available. The issue would come if someone changed what props an event generated - this would just lead to an undefined crash which are hard to trace

jblake42 (approved):
This is good and seems to work well when tested. I am not 100% on the typing of the context props but I am happy for it to merge - we just need to be strict on the props that the generator provides

PR #60, opened 2025-07-07
Fix context-menu: correctly use subplot-specific axes for pixel-to-data conversion

abdullah-ukaea:
This PR address issue (https://github.com/ukaea/viz-annotation/issues/27) by first confirming and then fixing the subplot-axis bug in handleContextMenu.

During the investigation I reproduced the issue: right-clicks on the lower subplot were always converted with the primary yaxis, giving values in the top panel’s range (see Before screenshot). The culprit was that we always referenced plot._fullLayout.xaxis / yaxis.

The dispatcher now inspects the data-subplot attribute present on every Plotly drag rectangle to identify which axis pair (xaxisN / yaxisN) should handle the click. If that attribute is absent it falls back to the nearest .subplot group and parses its xNyN class. Once the suffixes are extracted, the code swaps the default primary axis objects for the subplot-specific ones before calling p2d, ensuring that pixel-to-data conversion always reflects the panel under the mouse. A couple of short debug logs were added (subplot id and axes in use). no other code was changed outside of handleContextMenu.

**Before**  (baseline with wrong y on bottom subplot)

test1: right click on top subplot
screenshot3

test2: right click on bottom subplot
screenshot4

**After** (axis pair correctly switches to x2/y2)

test 1: right click on top subplot
Screenshot1

test2: right click on bottom subplot
screenshot2

With the correct axes in use, pixel clicks on any subplot now yield the expected data-space coordinates and the surrounding tools (zone creation, v-spans, etc.) behave consistently across panels.

### DIFF
--- a/services/ui/src/app/components/plots/time-series.tsx
+++ b/services/ui/src/app/components/plots/time-series.tsx
@@ -182,17 +182,8 @@ export const TimeSeries = ({
             /* 
             determine local axes for the subplot clicked
             Prefer the data-subplot attribute available on drag layers;
-            fall back to searching for the closest subplot element
             */
-            let subplotId = (event.target as HTMLElement).dataset.subplot // e.g. "x2y2"
-            //  Fast path: drag rectangle always has data-subplot
-            if (!subplotId) {
-                const subplotEl = (event.target as HTMLElement).closest(".subplot") as HTMLElement | null
-                //  Fallback: walk up DOM if attribute missing
-                if (subplotEl) {
-                    subplotId = [...subplotEl.classList].find(c => c !== "subplot") // extract "xNyN"
-                }
-            }
+            let subplotId = (event.target as HTMLElement).dataset.subplot // e.g. "x2y2"         
             if (subplotId) {
                 const m = subplotId.match(/^x(\d*)y(\d*)$/)               // ['', '2', '2']
                 // m[1]/m[2] hold numeric suffixes empty string -> primary axis
--- a/services/ui/src/app/components/plots/time-series.tsx
+++ b/services/ui/src/app/components/plots/time-series.tsx
@@ -172,8 +172,8 @@ export const TimeSeries = ({
             picks the matching xaxisN / yaxisN, so the props are correct for any subplot.
         */
         function handleContextMenu(event: MouseEvent, plot) {
-            let xaxis = plot._fullLayout.xaxis  // x-axis descriptor
-            let yaxis = plot._fullLayout.yaxis  // y-axis descriptor
+            let xaxis: any // will be assigned to the subplot-specific or primary x-axis below
+            let yaxis: any  // will be assigned to the subplot-specific or primary y-axis below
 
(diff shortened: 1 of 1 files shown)


services/ui/src/app/components/plots/time-series.tsx
```diff
@@ -168,19 +168,58 @@ export const TimeSeries = ({
...
             const relY = event.clientY - bb.top       // click Y in pixels, relative to plot
 
+            /* 
+            determine local axes for the subplot clicked
+            Prefer the data-subplot attribute available on drag layers;
+            fall back to searching for the closest subplot element
+            */
+            let subplotId = (event.target as HTMLElement).dataset.subplot // e.g. "x2y2"
+            //  Fast path: drag rectangle always has data-subplot
+            if (!subplotId) {
```
jblake42, line 189: Could you explain this alternative method please. From inspection I can't see that the event.target ever has an ancestor with the class `subplot` so I can't see that this adds anything as a fallback?

services/ui/src/app/components/plots/time-series.tsx
```diff
@@ -168,19 +168,58 @@ export const TimeSeries = ({
...
+                }
+            }
+            if (subplotId) {
+                const m = subplotId.match(/^x(\d*)y(\d*)$/)               // ['', '2', '2']
+                // m[1]/m[2] hold numeric suffixes empty string -> primary axis
+                if (m) {
+                    const suffixX = m[1] ?? ""                            // '' -> xaxis
+                    const suffixY = m[2] ?? ""                            // '' -> yaxis
+                    // Swap to subplot-specific axes if they exist
+                    xaxis = plot._fullLayout[`xaxis${suffixX}`] ?? xaxis
```
jblake42, line 203: Nice use of the `??` operator

services/ui/src/app/components/plots/time-series.tsx
```diff
@@ -168,19 +168,58 @@ export const TimeSeries = ({
...
+                    yaxis = plot._fullLayout[`yaxis${suffixY}`] ?? yaxis
+                }
+            }
+
             // Coordinates in data space
             const x      = xaxis.p2d(relX)   // data-space X at click
             const y      = yaxis.p2d(relY)     // data-space Y at click
-           
+            
+            // DEBUG statemtents to verify the axis data is correct when testing
```
jblake42, line 212: Good debugging - please remove before we merge

jblake42 (changes requested):
This is great Abdullah, really nice investigation into what data is available - please could you just have a look at the comments (specifically my question regarding the fallback)

services/ui/src/app/components/plots/time-series.tsx
abdullah-ukaea: I have now tried to address this comment in https://github.com/ukaea/viz-annotation/pull/60/commits/62db059abd2307c7b023255b0970046b6ad60aa5

jblake42 (approved):
LGTM

PR #62, opened 2025-07-15
Create LICENSE for the project

samueljackson92:
Using MIT as this is open source and follows what we used for [FAIR MAST](https://github.com/ukaea/fair-mast)


NathanCummings (approved):
As long as Dan is happy with use of MIT, I am as well. There is a licensing committee that I learned about today, but apart from Aby, I don't know who is on it.

PR #66, opened 2025-07-16
Project configuration UI

### LINKED ISSUE #47: Add project configuration pages
We need to add the ability for the user to create a project with a collection of samples. I think we should

- Add a button to the project list page to create a project
- Create a new page which allows the user to enter the information for setting up a project, including
    -  name
    -  query strategy (probably just a dropdown).
    - data loader (UDA/Parquet/ etc.), perhaps just a drop down for now?
    - additional info for the data loader
        -  e.g. column names, signals names, folder etc...

samueljackson92:
Add the ability to create a project through the UI.
Closes #47


services/api/crud/db.py
```diff
@@ -9,79 +8,98 @@
...
-        
+
+    async def update(
+        self,
+        collection: typing.Literal["projects", "annotations", "models", "samples"],
+        object_id: ObjectId,
+        model: pydantic.BaseModel,
+    ):
+        document = model.model_dump(mode="python")
+        result = await self.db[collection].update_one(
```
wk9874: Here you will require the full Project definition to be defined again, even if you are only updating one aspect of the Project. This means there is no validation check to see what you are / are not allowed to update about a project from the backend side.

I would recommend we create new schemas called `ProjectUpdate` which contains only the fields which you want to be updatable, and have them all be optional. Then in this update method, get the existing document, apply the changes by unpacking the existing and new models into a single dict, and then update the entry with that.

See my models branch for how I've done it: 

New schema: https://github.com/ukaea/viz-annotation/blob/wk9874/model_backend/services/api/schemas/models.py#L17

DB update method: https://github.com/ukaea/viz-annotation/blob/wk9874/model_backend/services/api/crud/db.py#L47C15-L47C21

(or happy to be corrected if you think your way is better!)

services/api/crud/utils.py
```diff
@@ -76,4 +76,5 @@ async def get_samples(
         start=start,
         limit=end - start + 1 if end is not None else 0,
     )
+    samples = [Sample(**sample) for sample in samples]
```
wk9874: Agree we should do this, just need to make sure it has been changed in all places which use this method below.

Also should change `get_annotations` to return a list of Annotations to be consistent :)

services/api/routers/files.py
```diff
@@ -0,0 +1,12 @@
+from pathlib import Path
+from fastapi import APIRouter, Request
+
+router = APIRouter(prefix="/files", tags=["Files"])
+
+
+@router.get("")
+async def get_files(request: Request, file_path: str, file_type: str) -> list[str]:
```
wk9874: What's this used for?

services/api/routers/files.py
```diff
@@ -0,0 +1,12 @@
...
+    file_names = Path(file_path).glob(f"*.{file_type}")
```
wk9874: Possibly `file_path` should be changed to `dir_path`?

services/api/routers/projects.py
```diff
@@ -101,79 +97,15 @@ async def get_project(
 )
 async def set_project(
```
wk9874: Should change the name of this function to `update_project` - also need to update the description of the 200 response above

services/api/routers/projects.py
```diff
@@ -101,79 +97,15 @@ async def get_project(
...
     request: Request,
+    project: ProjectIn,
```
wk9874: Yeah so here I would use a new `ProjectUpdate` schema instead

services/api/routers/projects.py
```diff
@@ -101,79 +97,15 @@ async def get_project(
...
-    obj_id = convert_to_objectid(project_id, "projects")
-
-    projects = await request.app.state.db_client.get_filtered_documents(
-        collection="projects", filters={"_id": obj_id}
-    )
-
-    if len(projects) == 0:
+    project_id = convert_to_objectid(project_id, "projects")
+    db_client: MongoDBClient = request.app.state.db_client
+    result = await db_client.update("projects", project_id, project)
```
wk9874: Should this stuff be abstracted out into its own `update_project` function in `utils`? ;)

services/api/routers/projects.py
```diff
@@ -101,79 +97,15 @@ async def get_project(
...
+    if result.modified_count == 0:
```
wk9874: Something I fell headfirst into in the models work - if you try to update a project with the state it is already in, this will fail as the document was not modified. That might be what you want in the case of this endpoint? But just be aware that if someone goes to update a project and presses save twice (or unknowingly tries to save the project with the same settings it already has) then the UI will get a 404 back.

If you do want to keep that behaviour, it shouldn't be a 404 since the project might have been successfully found, but just not updated...

services/api/routers/projects.py
```diff
@@ -190,10 +122,21 @@ async def delete_project(
...
 
-    result = await request.app.state.db_client.delete_filtered_documents(
+    # Clean up all associated samples
+    await db_client.delete_filtered_documents(
+        collection="samples", filters={"project_id": project_id}
+    )
+
+    # Clean up all associated annotations
+    await db_client.delete_filtered_documents(
+        collection="annotations", filters={"project_id": project_id}
```
wk9874: Why do the annotations and samples deletes use project_id as a string, but the projects delete uses it as an object ID? Suspect this will not work for annotations and samples delete?

wk9874 (changes requested):
Have reviewed the backend changes only

services/api/crud/utils.py
wk9874: Oh wait maybe you cant do it for annotations since you dont know which Annotation type to unpack to...

services/api/routers/files.py
samueljackson92: This is used by the front end UI to query the list of files in the path that you've specified. This can then be used to populate the list of samples you want to create

services/api/crud/db.py
samueljackson92: I am not sure I understand the reasoning for this? Doesn't this produce a lot of code duplication? I am not sure of the motivation for not just updating the whole record? Performance is likely not an issue?

services/api/crud/utils.py
samueljackson92: Should be able to do
```python
    annotations = [
        AnnotationTypes.model_validate(annotation) for annotation in annotations
    ]
```

services/api/crud/db.py
wk9874: I don't think it creates much code duplication, only requires a new Schema to be defined for each type of thing. What it allows for is

* Better validation - if there are things about a schema which should not be allowed to change, eg say a Project is not allowed to change Task, then the API will throw back a validation error instead of allowing it to be overwritten in the database. Obviously this should be constrained within the UI but we should also check for it on the backend side

* If doing updates on the backend, you would either need to create a new model by unpacking the values from the existing model and new model together (annoying, prone to go wrong) or override the attribute in the model instance. However doing this will not revalidate the model. Eg:
```
# Say I needed to update a Project instance on the backend after instantiation
project = Project(name="test", _id = "skskssksksksksk", task="disruption", query_strategy="random", data_loader="uda")
# With your way, I need the full Project definition again, even if I just want to update eg query_strategy
# So I would probably do
project.query_strategy = "sequential"
db_client.update_one(model=project)

# But this doesn't revalidate - it's valid to do
project.query_strategy = "garbage"
db_client.update_one(model=project)

# If we have a specific ProjectUpdate class though, we could do something like
updates = ProjectUpdate(query_strategy="sequential)
db_client.update_one(id=project._id, update_model=updates)
```

* Reduces risk of overwriting the project with stale data. Isn't too much of a concern atm, but imagine a UI open with a cached version of a project. Say the project name changes elsewhere (another UI open in a different tab, someone else changes it, the backend needs to change it for some reason, etc). If the UI sends the full cached model back on update, it will overwrite the new name with the old one unintentionally. If it only sends the fields which the user changed, it wont update any other fields

services/api/crud/utils.py
wk9874: Does that work? The thing I use for models (added in Pydantic v2) is:

```
from pydantic import TypeAdapter

annotator_model = TypeAdapter(AnnotationTypes)
annotation_models = [annotator_model.validate_python(annotation) for annotation in annotations]
```

You may also need to do somehting similar for Samples since they can have a range of different `data` models

samueljackson92:
Ok. I've made changes based on the comments above. 
I've also merged the base branch into this one and fixed any front end UI type issues.

services/api/crud/utils.py
```diff
@@ -78,3 +82,35 @@ async def get_samples(
     )
     samples = [Sample(**sample) for sample in samples]
```
wk9874, line 83: Do we need a typeadapter here too? Thinking for the `data` attribute which has multiple different possible Models. (You'll need to check how deep typeadapter checks...)

services/api/crud/utils.py
```diff
@@ -78,3 +82,35 @@ async def get_samples(
...
     return samples
+
+
+async def update_project(
+    db_client: MongoDBClient, project_id: str, project: ProjectUpdate
+) -> None:
+    project_id = convert_to_objectid(project_id, "projects")
+
+    result = await db_client.update("projects", project, project_id)
+    if result.modified_count == 0:
```
wk9874: Again this will return a 404 if you try to update a project with its current settings. I suspect you dont want it to do this, since if the user presses save twice in a row, theyll get a 404 error back the second time. Either change the 404 to something else (maybe a 204?) or just return a 200.

Alternatively (and this is probably preferred), change `.modified_count` to `.matched_count`

services/ui/src/app/projects/page.tsx
```diff
@@ -63,33 +77,50 @@ export const ProjectsTable = () => {
...
+          }
+        }}
       >
         <TableHeader>
           <Column>Name</Column>
           <Column>Task</Column>
           <Column>Date Created</Column>
           <Column>Loader</Column>
-          <Column>Controls</Column>
+          <Column>Edit</Column>
```
wk9874: I think this would look better if there was just a 'Edit' button in the same style as the 'Delete' button once you've selected a row in the project table. Doesn't need to be another column

services/ui/src/app/projects/create/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <div>
+      <ProjectCreateBreadCrumbs />
+      <div className="w-screen h-screen flex items-center justify-center bg-gradient-to-br from-gray-200 via-gray-300 to-gray-400">
+        <div className="w-full md:w-4/5 p-6 bg-white/60 text-gray-800 rounded-lg shadow-lg backdrop-blur-sm">
+          <h1 className="text-2xl font-bold mb-4">
+            Create Project
+          </h1>
+            <Provider theme={defaultTheme}>
+            <ToastContainer placement="top" />
```
wk9874: If I press enter before the form is complete, eg on File Path, I get this broken error message at the top:

image

services/ui/src/app/projects/create/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Provider theme={defaultTheme}>
+        <Breadcrumbs>
+          <Item key="projects" href={`${process.env.NEXT_PUBLIC_API_URL}/projects/`}>Projects</Item>
+          <Item key="create">Create</Item>
+        </Breadcrumbs>
+      </Provider>
+  );
+};
+
+export default function ProjectCreateForm() {
```
wk9874: In this we need to be careful how the project and task interact. Eg I tried to create an ELM project from UDA with shot IDs 10000 - 11000, and with column names Ip. When I try to load a sample from the project I get:

image


Suspect there needs to be some more validation on what columns are required for specific tasks etc?

services/ui/src/app/projects/page.tsx
```diff
@@ -12,47 +17,121 @@ export const ProjectsBreadCrumbs = () => {
...
-    ...rest,
-    id: _id
-  }));
 
   return (
-    <Provider theme={defaultTheme}>
+    <>
+      <Flex direction='row' margin='size-100' gap="size-100"  alignItems="end">
+        <ButtonGroup UNSAFE_className="py-2">
+          <Button elementType="a" variant='primary' href={`${process.env.NEXT_PUBLIC_API_URL}/projects/create`}><AddCircle/><Text>Create</Text></Button>
```
wk9874: Seems like the create button doesn't appear if you dont have any projects already:

image

services/ui/src/app/projects/page.tsx
wk9874: Might also look nicer to have the edit project stuff in a modal rather than a new page?

wk9874 (changes requested):
Re-reviewed backend changes. Haven't reviewed UI code but did play around with the interface, and left comments based on that :)

services/ui/src/app/projects/create/page.tsx
samueljackson92: This is going away... In later PRs `elms.tsx` moves to -> `time-series.tsx` and it is agnostic to the data used. In this PR it still is very specific about what signals are allowed.

services/ui/src/app/projects/page.tsx
samueljackson92: > Might also look nicer to have the edit project stuff in a modal rather than a new page?

Quite possibly. I will experiment when I work on this next.

> I think this would look better if there was just a 'Edit' button in the same style as the 'Delete' button once you've selected a row in the project table. Doesn't need to be another column

Then what happens if I select multiple items?

services/api/crud/utils.py
samueljackson92: I don't think we need it here. We only needed it for the top level `Union` type. I think pydantic can figure out the data type given it knows it is a sample.

services/ui/src/app/projects/page.tsx
samueljackson92: I've removed the ability to do multi-select deletion. I think you're unlikely to have loads of projects. Also with the pagination PR now in, it is a more complicated thing to do. For for now I'll leave it as a button per row.

samueljackson92:
@wk9874 ok, I think most of the issues above should be address. Please give it another go when you have the time.

services/api/crud/utils.py
```diff
@@ -58,10 +58,6 @@ async def get_annotations(
         start=start,
         limit=count if count is not None else 0,
     )
-
-    annotations = [
-        AnnotationTypeAdapter.validate_python(annotation) for annotation in annotations
```
wk9874: Why has this been removed in this version?

services/ui/src/app/projects/page.tsx
wk9874: Can't we do single row selection, and keep the edit / delete buttons in the header of the table instead of needing a new column?

https://react-spectrum.adobe.com/react-spectrum/TableView.html#single-selection

services/ui/src/app/projects/components/project_config.tsx
```diff
@@ -366,36 +366,88 @@ const makeSamples = async (projectId: string, samples: Sample[]) => {
 
   if (!response.ok) {
     const error = await response.json();
-    ToastQueue.negative(`Error creating samples: ${error}`, {timeout: 3000})
+    throw new Error(`Error creating samples: ${error.message}`);
   }
 }
 
-const createProject = (projectName: string, dataLoaderName: string, task: string, queryStrategy: string): Project => {
+const buildProject = (projectName: string, dataLoaderOptions: DataLoaderOptions, task: string, queryStrategy: string): Project => {
```
wk9874: I now get this error when I try to setup a project as shown:

image

wk9874 (changes requested):
The modals look much better than the pages! A couple more things for you to look into though...

(Again, haven't reviewed any frontend code)

services/api/crud/utils.py
wk9874: Have you also made sure that you have changed wherever this function is used to expect an Annotation class instead of a dict? (Assuming the above removal is a mistake and you do in fact want this to return Annotations)

samueljackson92: Accidentally removed during merge.

samueljackson92:
Tests checked locally, but are failing due to CI build credits.

services/api/crud/db.py
```diff
@@ -47,6 +47,25 @@ async def insert_many(
         result = await self.db[collection].insert_many(documents)
         return [str(object_id) for object_id in result.inserted_ids]
 
+    async def update(
```
wk9874, line 60: Need to add a test into `/tests/api/crud/test_db.py` for this

services/api/crud/utils.py
```diff
@@ -85,9 +78,43 @@ async def get_samples(
         start=start,
         limit=count if count is not None else 0,
     )
+
+    samples = [Sample(**sample) for sample in samples]
     return samples
 
 
+async def update_project(
```
wk9874, line 87: Need test for this in `tests/api/crud/test_utils.py`

services/api/routers/files.py
```diff
@@ -0,0 +1,12 @@
+from pathlib import Path
+from fastapi import APIRouter, Request
+
+router = APIRouter(prefix="/files", tags=["Files"])
+
+
+@router.get("")
+async def get_files(request: Request, dir_path: str, file_type: str) -> list[str]:
```
wk9874, line 8: Need to add tests for this in `tests/api/routers/test_files.py

services/api/routers/projects.py
```diff
@@ -110,80 +105,13 @@ async def get_project(
         404: {"description": "Project not found with that ID."},
     },
 )
-async def set_project(
+async def update_project(
     request: Request,
+    project: ProjectUpdate,
     project_id: str = Path(description="The ID of the project to activate"),
 ):
```
wk9874, line 112: Missing docstring

services/api/routers/samples.py
```diff
@@ -201,6 +202,29 @@ async def get_next_sample(
     return sample
 
 
+@router.get("/summary")
+async def get_sample_summary(
```
wk9874, line 205: Requires tests in `tests/api/routers/test_samples.py`

services/api/routers/samples.py
```diff
@@ -201,6 +202,29 @@ async def get_next_sample(
...
+    request: Request,
+    project_id: str = Path(
+        description="The ID of the project to get a summary of samples from."
+    ),
+) -> SampleSummary:
```
wk9874, line 210: Missing dostring

services/api/routers/samples.py
wk9874: Also please insert relevant test into `tests/api/routers/test_common.py` to check that it throws the correct error if non-existent or invalid object ID is entered - essentially just add this line into the parameterization:

` ("/projects/id/samples/summary", "project", "get", {}),`
And a relevant ID, eg:
`"get_sample_summary-wrong_project_id"`

services/api/schemas/projects.py
```diff
@@ -37,3 +38,9 @@ class ProjectIn(ConfiguredModel):
 
 class Project(ProjectIn):
     id: str = Field(..., alias="_id", description="The ID of this project.")
+
+
+class ProjectUpdate(ConfiguredModel):
+    name: Optional[str]
```
wk9874: I think you will need to set all of these to have `None` as the default value, otherwise if you try to instantiate one without any of the 'optional' fields provided in Pydantic v2, you will still get a validation error. Eg:

```
>>> class MyModel(pydantic.BaseModel):
...     a: typing.Optional[int]
... 
>>> mymodel = MyModel()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/wk9874/Documents/viz-annotation/.venv/lib/python3.12/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyModel
a
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```

See https://docs.pydantic.dev/dev/migration/#required-optional-and-nullable-fields

services/api/schemas/samples.py
```diff
@@ -45,7 +45,7 @@ class ShotData(BaseModel):
 
 class SampleBase(ConfiguredModel):
     shot_id: int
-    data: FileData | ShotData | TimeSeriesFileData
+    data: TimeSeriesFileData | FileData | ShotData
```
wk9874: Have seen this repeated in a couple of places, should we define a Union somewhere, similar to what we do with AnnotationTypes, and reference that everywhere?

services/api/schemas/samples.py
```diff
@@ -55,3 +55,10 @@ class SampleIn(SampleBase):
 class Sample(SampleBase):
     id: str = Field(..., alias="_id")
     project_id: str
+
+
+class SampleSummary(BaseModel):
+    total: int
+    shot_min: Optional[int]
+    shot_max: Optional[int]
+    data: Optional[TimeSeriesFileData | FileData | ShotData]
```
wk9874: ie here should be the same union as above

services/ui/src/app/projects/components/project_config.tsx
```diff
@@ -0,0 +1,735 @@
...
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(samples),
+    }
+  );
+
+  if (!response.ok) {
+    const error = await response.json();
+    throw new Error(`Error creating samples: ${error.message}`);
```
wk9874: This gives me 'Error creating samples: Undefined' in the case of a validation error. The structure of the response for a 422 is:

```
{"detail": [
    "input": 1000.1
    "loc": ['body', 0, 'shot_id']
    "msg":  "Input should be a valid integer, got a number with a fractional part"
    "type": "int_from_float"
    ], ...
}
```
Need to make sure you access the correct thing as 'message' is undefined

wk9874: May need some kind of error parser function in the UI if there are multiple different possible formats of error being returned?

services/ui/src/app/projects/components/project_config.tsx
```diff
@@ -0,0 +1,735 @@
...
+      borderRadius="medium"
+      padding="size-250"
+    >
+      <Flex direction="column">
+        <Flex direction="row" gap="size-200" alignItems="center">
+          <NumberField
+            label="Shot Min"
+            isRequired
+            value={shotMin ?? undefined}
+            onChange={setShotMin}
```
wk9874, line 200: This, and the Shot Max field below, currently allow decimals, which will cause a validation error in the server if entered. To fix this, you could add:

```
            minValue={1}
            step={1}
```
to each NumberField - that way if the user enters a decimal, itll snap to the next nearest integer

services/ui/src/app/projects/components/project_config.tsx
```diff
@@ -0,0 +1,735 @@
...
+
+const DataLoaderOptionsSchema = z.object({
+  name: z.string(),
+  signal_names: z.array(z.string()),
+});
+type DataLoaderOptions = z.infer<typeof DataLoaderOptionsSchema>;
+
+const UDADataLoaderOptionsSchema = DataLoaderOptionsSchema.extend({
+  shot_min: z.number(),
+  shot_max: z.number(),
```
wk9874, line 63: This contains no validation that max >= min. We should validate this here, as otherwise you'll pass in an empty list of samples into the server which causes an internal server error - I think zod can do this type of validation?

wk9874 (changes requested):
Mostly comments on new tests needed :)

wk9874:
@samueljackson92 just checking on the status of this - have you had a chance to address my latest comments? Is it ready for another review? :)

samueljackson92:
@wk9874 I believe i've fixed the above points and added the unit tests (which are passing! :) ). Can we merge?

services/ui/src/app/projects/components/project_config.tsx
wk9874: Dont think this has been fixed, I get this error if I try to set shot min < shot max:

```
api_app        | INFO:     172.18.0.2:52772 - "POST /projects/6911b8f7486affb919bd1ea2/samples HTTP/1.1" 500 Internal Server Error
api_app        | ERROR:    Exception in ASGI application
api_app        | Traceback (most recent call last):
api_app        |   File "/usr/local/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
api_app        |     result = await app(  # type: ignore[func-returns-value]
api_app        |   File "/usr/local/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
api_app        |     return await self.app(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/fastapi/applications.py", line 1134, in __call__
api_app        |     await super().__call__(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/applications.py", line 113, in __call__
api_app        |     await self.middleware_stack(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 186, in __call__
api_app        |     raise exc
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 164, in __call__
api_app        |     await self.app(scope, receive, _send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 63, in __call__
api_app        |     await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
api_app        |     raise exc
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
api_app        |     await app(scope, receive, sender)
api_app        |   File "/usr/local/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
api_app        |     await self.app(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 716, in __call__
api_app        |     await self.middleware_stack(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 736, in app
api_app        |     await route.handle(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/routing.py", line 290, in handle
api_app        |     await self.app(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/fastapi/routing.py", line 125, in app
api_app        |     await wrap_app_handling_exceptions(app, request)(scope, receive, send)
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
api_app        |     raise exc
api_app        |   File "/usr/local/lib/python3.10/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
api_app        |     await app(scope, receive, sender)
api_app        |   File "/usr/local/lib/python3.10/site-packages/fastapi/routing.py", line 111, in app
api_app        |     response = await f(request)
api_app        |   File "/usr/local/lib/python3.10/site-packages/fastapi/routing.py", line 391, in app
api_app        |     raw_response = await run_endpoint_function(
api_app        |   File "/usr/local/lib/python3.10/site-packages/fastapi/routing.py", line 290, in run_endpoint_function
api_app        |     return await dependant.call(**values)
api_app        |   File "/app/services/api/routers/samples.py", line 90, in add_samples
api_app        |     ids = await request.app.state.db_client.insert_many(
api_app        |   File "/app/services/api/crud/db.py", line 47, in insert_many
api_app        |     result = await self.db[collection].insert_many(documents)
api_app        |   File "/usr/local/lib/python3.10/site-packages/pymongo/_csot.py", line 115, in csot_wrapper
api_app        |     return await func(self, *args, **kwargs)
api_app        |   File "/usr/local/lib/python3.10/site-packages/pymongo/asynchronous/collection.py", line 960, in insert_many
api_app        |     raise TypeError("documents must be a non-empty list")
api_app        | TypeError: documents must be a non-empty list
```

More worryingly, thats not even a valdation error from the schema, its a database crash..... Need to fix that in the schema (can either do that here or a new hotfix branch)

wk9874: This is also the error that pops up at the top of the UI page :/

```
SyntaxError: Unexpected token 'I', "Internal S"... is not valid JSON
```

wk9874:
@samueljackson92 only change is the min and max shot numbers validation (both client and server side), otherwise looks good to me!

samueljackson92:
@wk9874 ok, validation should be fixed now. And I've fixed the merge issues. Ready for retest...

wk9874 (approved):
Looks good to me!

PR #74, opened 2025-07-28
Annotation upgrades

### LINKED ISSUE #34: Better time series annotations
As discussed a little bit on Monday. I think we can improve the usability of the region annotations a little bit.

1. As a user, I would like to be able to create a region by toggling, say, a button and then use left-click and dragging to create the region selection. Similarly, I would like to create a time point by toggling a button, then using left-click to create a time point.

2. As a user I would like to be able to "select" a region and delete it using the backspace key.

jblake42:
This PR addresses #34 which aims to improve the time series annotation process.

The overall goal of the PR is to allow users to select an annotation mode use a button and then be able to draw annotations without needing the context menu - though the context menu should still be a valid annotation option

Some limitations were identified that steered the implementation:
- Standard left click and drag to draw is not possible as plotly blocks many of the required event handlers such as mouse up even when drag mode is disabled
- Shift-click is used for zooming - we want to keep this functionality
- Ctr-click is a good option for draw mode but requires all panning controls to be disabled temporarily to prevent a plotly bug

The implemenatation is as follows:
- The context provider has been expanded to provide a range of generic functionality
- The context provider allows for a tooling callback to be set which can be used by the time series when user interactions occur
- When control is held all tooling interaction is disabled to allow for drawing of annotations over the top of exisiting tooling
- When control is held, all plotly interaction modes are disabled to prevent weird plotly behaviour and to set the cursor correctly
- Example of tooling buttons added which use context functions provided by the tooling providers to set the required callbacks. This can be done from any component within the tooling providers so is very flexible
- Zone creation logic has also be tweaked to allow for zones to be dragged past their start point to improve usability
   -  Part of this included adding a batched update handler which is useful for updating the backend too


samueljackson92 (changes requested):
This is looking good! A couple of comments:

1. When add or delete zones I am getting a `TypeError: onModifyZone is not a function` on the UI. I think this is because that callback is currently not optional and not defined.
2. On my machine, when I ctrl+left click drag to create an annotation the context menu always seems to open? I don't think this should be the case
image

samueljackson92:
Another thing to note here, I think as discussed with Rui this morning, ideally we should have:

1) By default the user can pan & zoom. This should include scroll zooming as in #80, without the user getting "caught" but the annotations. 
2) The user can has to hold a key modifier to enter drag create mode

samueljackson92:
Or another idea for 2): the user enters into creation mode when they toggle the annotation button..

So the flow is:
1) Click a button for the type of zone
2) Left click + drag now creates a new zone.
3) On release the button is now toggled back to deactivated

jblake42:
> This is looking good! A couple of comments:
> 
>     1. When add or delete zones I am getting a `TypeError: onModifyZone is not a function` on the UI. I think this is because that callback is currently not optional and not defined.
> 
>     2. On my machine, when I ctrl+left click drag to create an annotation the context menu always seems to open? I don't think this should be the case
> 
> 
> image

1. Yes this is due to the previous merge - I am hoping it is cleared up with proper typing in the ongoing merge request but if not we will have to fix it in a new PR. I do not think it should go in here as it is not related to this work
2. I'll look into this - I haven't noticed this

jblake42:
1. The onModify error was cleaned up in the other PR
2. I cannot reproduce the context menu bug
3. I propose that the handling of pan/zoom is done in a new PR to limit the scope of this PR - I have created an issue and will work on that next

samueljackson92 (changes requested):
Hi @jblake42, this is looking good. However, I am still getting the issue with context menu, but I believe I understand the problem.

On OSX you can do a right click event by right clicking the mouse, but you can also trigger a right click event by doing "Ctrl + Left Click" which is what we're using to trigger draw mode.

I've pushed a fix for this which works for OSX to your branch. Can you check it also works the machine you're using?

jblake42:
> Hi @jblake42, this is looking good. However, I am still getting the issue with context menu, but I believe I understand the problem.
> 
> On OSX you can do a right click event by right clicking the mouse, but you can also trigger a right click event by doing "Ctrl + Left Click" which is what we're using to trigger draw mode.
> 
> I've pushed a fix for this which works for OSX to your branch. Can you check it also works the machine you're using?

Thanks for looking into this - it still works my end. Are you happy for this to merge?

PR #76, opened 2025-07-28
Add support for Time Series Annotators

### LINKED ISSUE #70: Add change point detection annotator
We can use something like `ruptures` to add simple changepoint detection to timeseries data.

### LINKED ISSUE #71: Add simple time series outlier detection.
Add basic time series outlier detection tools such as:
- Deviation from the mean
- IsolationForest
- GP outliers.

### LINKED ISSUE #72: Add support for shot level annotation
As a user, I want to be able to add annotations for a whole shot, not just for time regions or time points. For example I many want to classify the this particular shot as "good" or "bad"

### LINKED ISSUE #77: Add documentation for different annotators
Add some documentation for the different annotators added as part of #76

samueljackson92:
This PR adds support for different types of annotators for time series data. This includes:

- **Shot labels**: class labels for the entire shot currently being viewed. This is currently a hard coded fixed list, but can be expanded to be dynamically set in the project settings later.
- **Peak detection**: detect peaks in data using scipy find peaks
- **Outlier detection**: detect outliers in data using either `Isoforest` or mean absolute deviation (`MAD`).
- **Changepoint detection**: detect changepoints in a time series trace using either the `Pelt` method of peanlised optimization or a hidden markov model `HMM`.
- **Jump detection**: detect sharp drops in a time series signal, such as see in soft x-rays when sawteething occurs.

To test:

 - Try out the different tools in the ELMs interface.
 - Code review of front end & back end code.

Closes #70
Closes #71 
Closes #72 
Closes #77


services/api/schemas/annotations.py
```diff
@@ -1,9 +1,15 @@
-from typing import Tuple, Optional, Union
+from typing import Literal, Tuple, Optional, Union
 from services.api.schemas import ConfiguredModel
+from services.api.schemas.annotators import AnnotatorIds
 from pydantic import Field, model_validator
 
 
 class AnnotationIn(ConfiguredModel):
+    label: str
+    created_by: AnnotatorIds
```
wk9874: Note to self for models work: This should also include the list of model types as valid entries

services/api/routers/annotations.py
```diff
@@ -45,7 +43,9 @@ async def get_all_annotations(
     annotations = await utils.get_annotations(
         db_client, project_id, validated, start, end
     )
-    print(annotations)
+    for annotation in annotations:
+        print(annotation)
```
wk9874: Can we get rid of this? Assume it was leftover from some debugging at some point

services/api/core/data_loaders.py
```diff
@@ -32,6 +32,19 @@ def get_sample(self, sample: Sample) -> ImageData:
         return ImageData(data=arr.tolist())
 
 
+def downsample_time_series(time, signal, num_points=500):
```
wk9874: This is duplicated - here and in `services/api/core/annotators.py`. Need to decide where it should live

Don't think it should be here as it doesn't do any data loading

services/api/routers/annotations.py
```diff
@@ -178,6 +172,18 @@ async def add_annotations(
     await request.app.state.db_client.delete_filtered_documents(
         collection="annotations", filters=ids
     )
+
+    if len(annotations) == 0:
+        return []
+
+    for annotation in annotations:
```
wk9874: Why do we need this? So long as the IDs are passed into `db_client.insert_many`, that should automatically add the IDs into the document before insertion

services/api/routers/annotations.py
```diff
@@ -178,6 +172,18 @@ async def add_annotations(
...
+        if annotation.project_id is None:
+            annotation.project_id = project_id
+        if annotation.sample_id is None:
+            annotation.sample_id = sample_id
+
+    print(annotations)
```
wk9874: Can also delete this I imagine

services/api/routers/annotators.py
```diff
@@ -1,23 +1,31 @@
...
     pass
 
+
 @router.post("/samples/{sample_id}/annotator/{annotator_id}")
-async def create_annotations(request: Request, project_id: str, sample_id: str, annotator_id: AnnotatorIds, params: AnnotatorTypes):
+async def create_annotations(
+    request: Request,
+    project_id: str,
+    sample_id: str,
+    annotator_id: AnnotatorIds,
```
wk9874: Wonder if we should change `annotator_id` to `annotator_type`? Since other `..._id` parameters refer to an ObjectID from a DB document, whereas these refer to the type of annotator to use

services/api/schemas/annotations.py
```diff
@@ -13,38 +21,42 @@ def set_uncertainty(cls, values):
 
         return values
 
-    validated: bool = False
-    uncertainty: Optional[float] = None
-    label: str
+
+class Annotation(AnnotationIn):
```
wk9874, line 25: Why does this no longer have project_id and sample_id?

services/api/core/annotators.py
```diff
@@ -1,30 +1,135 @@
...
+    OutlierDetectionParams,
+)
 from services.api.schemas.annotations import TimeRegion
 from services.api.schemas.projects import Task
 from services.api.schemas.annotators import AnnotatorIds
 
+
+def binary_runs_to_tuples(arr: np.ndarray) -> list[tuple[int, int]]:
+    """
+    Convert a 1D binary array into a list of (start, end) index tuples for each contiguous run of 1s.
```
wk9874, line 32: We should decide on a docstring format and stick with it, I think I normally use numpy style, but happy to use whatever!

services/api/core/annotators.py
```diff
@@ -1,30 +1,135 @@
...
+    representing contiguous segments of the same labels.
+
+    Parameters:
+        arr (np.ndarray or list): Input 1D array or list of labels.
+
+    Returns:
+        List[Tuple[int, int, int]]: List of (start, end, label) for each segment.
+    """
+    arr = np.asarray(arr)
+    if arr.ndim != 1:
```
wk9874, line 92: You do this in this function, but not in the `binary_runs_to_tuples` or `downsample_time_series` functions. Should you add it to those too?

services/api/core/annotators.py
```diff
@@ -1,30 +1,135 @@
...
+    Returns:
+        tuple[np.ndarray, np.ndarray]: Tuple containing the downsampled time and signal arrays.
+    """
+    if len(time) <= num_points:
+        return time, signal
+
+    time_coarse = np.linspace(time.min(), time.max(), num_points)
+    interpolator = interp1d(time, signal, kind="linear")
+    signal = interpolator(time_coarse)
+    time = time_coarse
```
wk9874: why not just return time_coarse?

services/api/core/annotators.py
```diff
@@ -46,22 +151,267 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
+
+    def mad_outliers(self, time: np.ndarray, data: np.ndarray) -> list[TimeRegion]:
+        median = np.median(data)
+        abs_deviation = np.abs(data - median)
+        mad = np.median(abs_deviation)
+
+        if mad == 0:
+            return []
+
+        modified_z_scores = 0.6745 * (data - median) / mad
```
wk9874: What's the significance of 0.6745? Should that be hardcoded or configurable? (maybe add a comment describing what that's doing

services/api/core/annotators.py
```diff
@@ -46,22 +151,267 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
                     time_max=float(peak_time + width),
+                    created_by=AnnotatorIds.PEAK_DETECTION,
                 )
                 regions.append(region)
 
         return regions
 
+
+class OutlierDetectionAnnotator(DataAnnotator):
+    def __init__(self, params: OutlierDetectionParams):
```
wk9874, line 273: Would be good to have a docstring for the class and methods - I dont understand how this works atm! :)

wk9874: Generally some more comments for these annotators would help too

services/api/core/annotators.py
```diff
@@ -46,22 +151,267 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
+                time_max=time[imax],
+                label="Outlier",
+                created_by=AnnotatorIds.OUTLIER_DETECTION,
+            )
+            for imin, imax in bounds
+        ]
+        return bounds
+
+
+class ChangePointDetectionAnnotator(DataAnnotator):
```
wk9874, line 344: Also requires docstrings

services/api/core/annotators.py
```diff
@@ -1,30 +1,135 @@
...
+        return time, signal
+
+    time_coarse = np.linspace(time.min(), time.max(), num_points)
+    interpolator = interp1d(time, signal, kind="linear")
+    signal = interpolator(time_coarse)
+    time = time_coarse
+    return time, signal
+
+
 class DataAnnotator(ABC):
```
wk9874, line 160: Should we also include 

```
@abstractmethod
def __init__(self, params: AnnotatorTypes):
    pass
```

Will just show people what the expected input into `__init__` is

services/api/schemas/annotators.py
```diff
@@ -14,15 +20,40 @@ class Annotator(BaseModel):
     pass
 
 
-class FindPeaksParams(Annotator):
+class PeakDetectionParams(Annotator):
```
wk9874: Perhaps `Annotator` should be renamed `AnnotatorParams`?

wk9874 (changes requested):
Have reviewed backend changes

services/api/core/annotators.py
samueljackson92: Scaling coefficient to convert MAD to follow the range of a standard z-score.

For a standard normal distribution the std is 1, and the mean absolute deviation is 0.6745. So we scale by this factor to renormalise the score to the range of a standard z-score, while being robust to outliers.

services/api/core/annotators.py
samueljackson92: I was actually thinking that we should make documentation pages for this. And the PR for mkdocs has now been merged.

services/api/routers/annotations.py
samueljackson92: Did not realise that the ids get automatically added.

services/api/schemas/annotations.py
samueljackson92: Because the front end may have an annotation that already has a project_id and sample_id associated with it that it is posting back (possibly modified).

services/api/core/annotators.py
samueljackson92: Yes, we should probably decide on some standards and put those in the dev docs/contributng.md I am happy with numpy style docstrings

services/api/schemas/annotations.py
wk9874: I think that is going to break stuff surely, since now when the API endpoints return an annotation, it will not attach their project or sample IDs. Shouldn't the frontend UI be posting the `AnnotationIn` schema if creating a new annotation, or `AnnotationUpdate` to update one as per the other PR? (but I dont thinkkk we will ever need to have AnnotationUpdate, since the backend should delete all previous annotations associated with the shots and add an entirely new set of annotations instead of updating)

samueljackson92: But it will attach the project and sample id because `Annotations` inherits from `AnnotationsIn`?

wk9874: Ohhh sorry I see, you've moved them to AnnotationIn, thought you had just removed them...

I still dont think we should allow this though - AnnotationIn should not have project_id or sample_id fields. Those should be removed from the annotation on the front end, and passed in as path parameters into the endpoint you are POST/PUT -ing to. Those fields are then automatically added back in by the endpoint based on the path params. Having them as optional parameters in the AnnotationsIn schema breaks this pattern which we follow with all other schemas, and it should remain consistent across them all :)

wk9874: Otherwise it adds ambiguity - what happens if you POST to `/projects/abc/samples/def/annotations`, but the request body has `AnnotationIn(..., project_id='uvw', sample_id='xyz')`? Which ones should it attach to the db entry?

samueljackson92: I agree. You're correct. I have changed it back.

samueljackson92:
Ok, @wk9874  I think this is ready for a re-review.

services/api/core/annotators.py
samueljackson92: I have added a annotator doc page with some nice pictures

wk9874:
Happy with the backend code!

wk9874 (commented):
Backend code looks good to me! (Haven't reviewed frontend code)

services/api/schemas/annotations.py
wk9874: Ahhhh damn, now I see the issue: Our GET endpoints return `-> AnnotationTypes`, which are `Union[TimePoint, TimeRegion, BoundingBox, VideoBoundingBox]`. Issue is that these inherit from `AnnotationIn`, not `Annotation`. Now that is correct for POSTing to the endpoints, but when GETing the endpoint will not attach the project, sample or annotation ID to the model.....

So maybe we need to do this?
```
class TimePointOut(TimePoint, Annotation):
    pass
class TimeRegionOut(TimeRegion, Annotation):
    pass
class BoundingBoxOut(BoundingBox, Annotation):
    pass
class VideoBoundingBoxOut(VideoBoundingBox, Annotation):
    pass
AnnotationOutTypes = Union[TimePointOut, TimeRegionOut, BoundingBoxOut, VideoBoundingBoxOut]
```
And then in the GET endpoint, return `-> AnnotationOutTypes`?

A bit ugly, but I think that is the only way it will work properly... thoughts @samueljackson92 ?

samueljackson92: But if you're doing that then why not just use `AnnotationTypes`? Why do we need the duplication? What is `TimePointOut` adding?

samueljackson92: ... Because `Annotation` has the `project_id` and `sample_id`. Maybe we should rename `Annotation` to `ProjectMixin` or similar.

wk9874: I guess really to be consistent, eg `TimePoint` should be renamed `TimePointIn`, and then `TimePoint(TimePointIn, Annotation`...

samueljackson92: Or we flip it around: `TimePoint` has optional ids, but `TimePointIn` overloads `TimePoint` with a non-optional IDs. `TimePointIn` is only used on incoming `PUT`/`POST` requests?

wk9874:
@samueljackson92 Something I spotted while writing some tests, why is the width in the FindPeaks annotator set to           `width = w * 3 * dt`?

This might make the distance parameter not work correctly?

```
def test_find_peaks_params():
    # Create data with clear peaks above noise
    data = numpy.random.uniform(0, 10, 100)
    data[5] = 1000
    data[30] = 500
    data[35] = 500
    data[40] = 500
    data[60] = 300
    data[80] = 100
    data[95] = 1000
    
    ts_data = TimeSeriesData(time=numpy.arange(100), values=data)
    mv_data = MultiVariateTimeSeriesData(values={"Ip": ts_data})
    
    params = FindPeaksParams(signal_name = "Ip", prominence=200, distance=10, time_min=10, time_max=90)
    annotator = annotators.FindPeaksAnnotator(params)
    time_regions = annotator.predict(mv_data)
    
    # Based on data above with these params:
    #   - Should recognise peaks at idx 30, 35, 40 as one peak since they are within 'distance'
    #   - Should find peak at idx 60
    #   - Should not find peak at idx 80 since this is below 'prominence'
    #   - Should not find peak at idx 5 since this is before 'time_min'
    #   - Should not find peak at idx 95 since this is beyond 'time_max'
    assert len(time_regions) == 2
    
    assert time_regions[0].time_min < 30 and  time_regions[1].time_min > 40
    time_mid = (time_regions[0].time_min + time_regions[0].time_max) / 2
    assert math.isclose(time_mid, 35, abs_tol=1)
```

Gets this error:
```
>       assert time_regions[0].time_min < 30 and  time_regions[1].time_min > 40
E       AssertionError: assert (31.96826144046949 < 30)
E        +  where 31.96826144046949 = TimeRegion(timestamp=datetime.datetime(2025, 8, 18, 11, 7, 46, 677024), validated=False, uncertainty=1.0, label='ELM', time_min=31.96826144046949, time_max=38.03173855953051).time_min
```

samueljackson92:
@wk9874 yes, it was a hack because the width returned was not sufficient. Here I've updated with a better implementation using `peak_widths`.

samueljackson92:
@wk9874 I also switched to enabling/disabling the tool.

services/ui/src/app/components/changepoints.tsx
jblake42, line 1: Could all of these files into an annotators folder please to keep the component folder tidy

services/ui/src/app/components/changepoints.tsx
```diff
@@ -0,0 +1,93 @@
+import { useEffect, useState } from 'react';
+import {Provider, defaultTheme, Slider, Flex, ComboBox, Item, Switch } from '@adobe/react-spectrum'
+import { Annotation, MultiVariateTimeSeriesData } from "@/types";
+
+
+type ChangePointDetectionType = {
+    project_id: string;
+    sample_id: string;
+    data: MultiVariateTimeSeriesData;
+    setAnnotations: (annotations: Annotation[] | ((prev: Annotation[]) => Annotation[])) => void;
```
jblake42: Is it worth abstracting this shared type out so that we can easily tweak it across all annotators in the future?

services/ui/src/app/components/changepoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+            });
+
+            const payload: Annotation[] = await response.json();
+            setAnnotations(previousAnnotations => {
+                const otherAnnotations = previousAnnotations.filter((annotation: Annotation) => annotation.created_by !== 'change_point_detection');
+                return otherAnnotations.concat(payload);
+            });
+        };
+        fetchData();
+    }, [signalName, penalty, method, numPoints, numComponents, isEnabled]);
```
jblake42: You're missing a few dependencies here: `data.values`, `project_id`, `sample_id`, and `setAnnotations`

services/ui/src/app/components/changepoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const signalOptions = Object.keys(data.values).map((value, index)=> ({id: index, name: value}));
+
+    const [isEnabled, setIsEnabled] = useState<boolean>(false);
+    const [signalName, setSignalName] = useState<string | null>(null);
+    const [penalty, setPenalty] = useState<number>(5);
+    const [numPoints, setNumPoints] = useState<number>(500);
+    const [method, setMethod] = useState<string>("pelt");
+    const [numComponents, setNumComponents] = useState<number>(3);
+
+    useEffect(() => {
```
jblake42: Below you have a `useEffect` that sets the annotations when the annotator is disabled however this `useEffect` seems to fetch the data and set it no matter what, only to be overwritten. This seems wasteful to me - could they be combined?

Also feel free to correct me if I've missed something?

services/ui/src/app/components/changepoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }, [signalName, penalty, method, numPoints, numComponents, isEnabled]);
+
+    useEffect(() => {
+        if (!isEnabled) {
+            setAnnotations((previousAnnotations: Annotation[]) => {
+                const otherAnnotations = previousAnnotations.filter((annotation: Annotation) => annotation.created_by !== 'change_point_detection');
+                return otherAnnotations;
+            });
+        }
+    }, [isEnabled]);
```
jblake42: `setAnnotations` is a dependency

services/ui/src/app/components/changepoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+type ChangePointDetectionType = {
+    project_id: string;
+    sample_id: string;
+    data: MultiVariateTimeSeriesData;
+    setAnnotations: (annotations: Annotation[] | ((prev: Annotation[]) => Annotation[])) => void;
+};
+
+export function ChangePointDetectionTool({ project_id, sample_id, data, setAnnotations }: ChangePointDetectionType) {
+    const methodOptions = [{id: 0, name: "pelt"}, {id: 1, name: "hmm"}];
```
jblake42: I would perhaps like to see this typing tightened up a bit. Is it worth making the names an `enum` if they are set options>

services/ui/src/app/components/changepoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+                <Switch isSelected={isEnabled} onChange={setIsEnabled}>Enable Tool</Switch>
+                <ComboBox label="Signal Name" defaultItems={signalOptions} onInputChange={setSignalName}>
+                    {x => <Item>{x.name}</Item>}
+                </ComboBox>
+                <br/>
+                <ComboBox label="Method" defaultItems={methodOptions} defaultInputValue={method} onInputChange={setMethod}>
+                    {x => <Item>{x.name}</Item>}
+                </ComboBox>
+                <br/>
+                {(method === "pelt") && (
```
jblake42: The `enum` would make this more robust imo

services/ui/src/app/components/jump.tsx
```diff
@@ -0,0 +1,80 @@
+import { useEffect, useState } from 'react';
+import {Provider, defaultTheme, Slider, Flex, ComboBox, Item, Switch } from '@adobe/react-spectrum'
+import { Annotation, Annotations, MultiVariateTimeSeriesData } from "@/types";
```
jblake42: Annotations is unused

services/ui/src/app/components/jump.tsx
```diff
@@ -0,0 +1,80 @@
...
+export function JumpDetectionTool({ project_id, sample_id, data, setAnnotations }: JumpDetectionType) {
+    const [isEnabled, setIsEnabled] = useState<boolean>(false);
+    const [signalName, setSignalName] = useState<string| null>(null);
+    const signalOptions = Object.keys(data.values).map((value, index)=> ({id: index, name: value}));
+    const [threshold, setThreshold] = useState<number>(2);
+    const [minDistance, setMinDistance] = useState<number>(5);
+    const [smoothingValue, setSmoothingValue] = useState<number>(2);
+    const [numPoints, setNumPoints] = useState<number>(2000);
+
+    useEffect(() => {
```
jblake42: I'll leave the comment one more time regarding checking for `isEnabled` - I won't repeat it for the other files

services/ui/src/app/components/jump.tsx
```diff
@@ -0,0 +1,80 @@
...
+            });
+
+            const payload: Annotation[] = await response.json();
+            setAnnotations((previousAnnotations: Annotation[]) => {
+                const otherAnnotations = previousAnnotations.filter((annotation: Annotation) => annotation.created_by !== 'jump_detection');
+                return otherAnnotations.concat(payload);
+            });
+        };
+        fetchData();
+    }, [signalName, minDistance, threshold, smoothingValue, numPoints, isEnabled]);
```
jblake42: I'll also only leave this comment once more here but I think all of your `useEffects` are missing dependencies. ESLint might help you catch some of these

services/ui/src/app/components/changepoints.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    method: method,
+                    penalty: penalty,
+                    num_points: numPoints,
+                    num_components: numComponents
+                }),
+            });
+
+            const payload: Annotation[] = await response.json();
+            setAnnotations(previousAnnotations => {
+                const otherAnnotations = previousAnnotations.filter((annotation: Annotation) => annotation.created_by !== 'change_point_detection');
```
jblake42: I think this should be a utility function it is repeated many times

services/ui/src/app/components/labels.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    const items = labels.map((label, index) => ({id: index, name: label}));
+    const [selectedKeys, setSelectedKeys] = useState<Set<string>>(new Set());
+
+    useEffect(() => {
+        const defaultAnnotations = annotations.filter((annotation: Annotation) => annotation.type === 'class_label');
+        const defaultSelectedKeys = defaultAnnotations.map((annotation: Annotation) => {
+            const index = labels.indexOf(annotation.label);
+            return index !== -1 ? index.toString() : null;
+        }).filter((key) => key !== null) as string[];
```
jblake42: I don't think we need the type assertion here - best to remove when not required for type safety

services/ui/src/app/components/labels.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    }, [annotations]);
+
+    const onSelectionChange = (keys: Selection) => {
+
+        let newKeys = new Set<string>();
+        if (keys === 'all') {
+            items.forEach(item => newKeys.add(item.id.toString()));
+        } else {
+            newKeys = new Set(Array.from(keys).map(key => key.toString()));
```
jblake42: Can we not take a similar approach to above rather than recreating sets:
`keys.forEach(key => newKeys.add(key.toString()))`

services/ui/src/app/components/labels.tsx
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        setAnnotations((prevAnnotations: Annotation[]) => {
+
+            let newAnnotations = prevAnnotations || [];
+            newAnnotations = newAnnotations.filter(annotation => annotation.type !== 'class_label');
+            newKeys.forEach((key: string) => {
+                let item = items.find(item => item.id.toString() === key) || null;
```
jblake42: `const`

services/ui/src/app/components/labels.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+                if (item === null) {
+                    console.warn(`Label with key ${key} not found in items.`);
+                    return;
+                }
+                
+                newAnnotations.push({
+                    type: 'class_label',
+                    label: item.name,
+                    created_by: 'manual',
```
jblake42: This is missing `time` which is not optional

services/ui/src/app/components/labels.tsx
```diff
@@ -0,0 +1,102 @@
...
+                });
+            });
+            return newAnnotations;
+        });
+    }
+
+    // Listen for global key presses
+    useEffect(() => {
+      const handleKeyDown = (e) => {
```
jblake42: Can we explicitly state event type please

services/ui/src/app/components/outliers.tsx
```diff
@@ -0,0 +1,88 @@
+import { useEffect, useState } from 'react';
+import {Provider, defaultTheme, Slider, Flex, ComboBox, Item, Switch } from '@adobe/react-spectrum'
+import { Annotation, Annotations, MultiVariateTimeSeriesData } from "@/types";
```
jblake42: Annotations is not used

services/ui/src/app/components/outliers.tsx
jblake42, line 1: Only one specific comment - but all of the above comments relate to this one

services/ui/src/app/components/peaks.tsx
jblake42, line 1: No specific comments - but all of the above comments relate to this one

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -81,7 +95,7 @@ function SaveButton({ project_id, sample_id, annotations }: SaveInfo) {
         timeout: 5000,
       });
     } catch (err) {
```
jblake42, line 116: Can we be specific about the error we are catching?

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -174,12 +188,13 @@ function AmplitudeSlider({
...
 }
 
+
 type ToolBarInfo = {
   project: Project;
   sample: Sample;
   data: Data;
   annotations: Annotations;
-  setAnnotations: (annotations: Annotations) => void;
+  setAnnotations: (annotations: Annotations | ((prev: Annotations) => Annotations)) => void;
```
jblake42: This is a good justification for having the common type definition in a localised place - on a side note we should aim to be consistent on the use of `Annotation[]` vs `Annotations`

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -194,67 +209,111 @@ export default function ToolBar({
...
 
-    const ampRangeTool = (
-      <AmplitudeSlider
-        data={mhdData.data}
-        viewParams={viewParams}
-        setViewParams={setViewParams}
-      />
-    );
-    tools.push(ampRangeTool);
+    let mhdData = resultSpec.data;
```
jblake42: `const`

services/ui/src/types.ts
```diff
@@ -64,6 +68,7 @@ export const CategorySchema = z.object({
 export type Category = z.infer<typeof CategorySchema>;
 
 export const ZoneSchema = z.object({
+  created_by: z.string().default('manual'),
```
jblake42: Doing this has created issues where this schema is used - `created_by` does not appear to be optional

services/ui/src/app/spectrogram/components/spectrogram.tsx
```diff
@@ -12,6 +12,8 @@ import {
   ZoneSchema,
   VSpanSchema,
   VSpan,
+  Annotation,
+  ClassLabelSchema,
```
jblake42: This doesn't seem to be being used

services/ui/src/app/utils.tsx
```diff
@@ -11,7 +11,9 @@ import {
   Zone,
   TimePoint,
   TimeRegion,
+  ClassLabelSchema,
```
jblake42: I don't think this needs to be added

services/ui/src/app/utils.tsx
```diff
@@ -11,7 +11,9 @@ import {
...
 } from "@/types";
+import { time } from "console";
```
jblake42: As above

services/ui/src/app/utils.tsx
jblake42, line 1: The `timeRegion` and `timePoint` typing in here seems off

jblake42: Lines 32 and 40

jblake42 (changes requested):
When I try to test this by adding a zone through the right click menu I get an error - I think this is due to the change to the annotation type in the types file requiring a created by field.

wk9874:
You will also need to write tests for each of the new annotators in the backend, as well as verify that the Find Peaks test passes with the changes made in this commit :)

samueljackson92:
@wk9874 @jblake42 thanks for all the feedback. I think I've addressed all the comments above. Give me a re-review when you have time. :)

samueljackson92:
Checked tests/linting locally 👍

jblake42 (approved):
LGTM from a front-end point of view - though there pbviously are pipeline issues. Retested the tooling a the zone bug does appear to have been resolved. There are some remaining unresolved comments that haven't been addressed or commented. I have left them unresolved but happy for them to be diregarded.

services/ui/src/app/components/annotators/changepoints.tsx
```diff
@@ -0,0 +1,159 @@
...
+    validSignalName,
+    setAnnotations,
+  ]);
+
+  return (
+    <Provider theme={defaultTheme}>
+      <div className="m-4">
+        <Flex direction="column">
+          <Switch isSelected={isEnabled} onChange={setIsEnabled}>
+            Enable Tool
```
wk9874, line 109: Dont know how easy/difficult this would be, but could we have it remove predictions made by any given annotator when you toggle it from Enabled to Disabled?

services/ui/src/app/components/annotators/changepoints.tsx
```diff
@@ -0,0 +1,159 @@
...
+            method: method,
+            penalty: penalty,
+            num_points: numPoints,
+            num_components: numComponents,
+          }),
+        },
+      );
+
+      const payload: Annotation[] = await response.json();
+      setAnnotations((previousAnnotations) => {
```
wk9874, line 82: Why does changepoint add a blue background over the whole plot?

wk9874 (commented):
Couple of minor questions, otherwise LGTM!

wk9874:
@samueljackson92 I think this is ready to merge, except for my two questions above :)

services/ui/src/app/components/annotators/changepoints.tsx
samueljackson92: No, it provides a (possibly many) region annotations. They will cover the whole time series and can be modified. I think in another branch I add a border to the regions to make it more clear...

samueljackson92:
@wk9874 fixed/responded to those final issues. Can we merge?

wk9874:
@samueljackson92  Looks good to me!

wk9874 (approved):

PR #85, opened 2025-08-04
Small zones bug fix for issue 59

### LINKED ISSUE #59: Issues when zones are made to be small
There are a couple of sub-issues identified relating to the zone creation tool:

- If a zone is dragged to be zero-width, it is not visible and cannot be interacted with however it is still present. Minimum width based on x-range should probably be added
- When a zone is made very small it impossible to interact using right-click as the drag handles have no listeners attached
- When the zone is made very small there is no way to directly drag the zone - interaction is only possible by initially dragging the zone to be wider. This may not be seen as being an issue howeever

abdullah-ukaea:
This PR fixes issue https://github.com/ukaea/viz-annotation/issues/59 by ensuring zones never collapse to zero width, remain right‐clickable when tiny, and preserve all existing drag/resize behaviors. We now clamp the left and right edges so that x1 – x0 never falls below 1% of the axis span, add context‐menu listeners to the resize handles so even narrow zones respond to right‐click, and keep the original full‐zone drag and edge‐resize functionality intact.

[issue59v2.webm](https://github.com/user-attachments/assets/6214a78e-88fc-4f78-9317-06ee5678faa3)

Closes #59


abdullah-ukaea:
This PR will not be merged with main until these 2 PRs are merged to main.

1) https://github.com/ukaea/viz-annotation/pull/45
2)  https://github.com/ukaea/viz-annotation/pull/74

Then a refactor will be needed to delay the clamping until the drag ends for example. That way we allow wrapping, but if the user ends the drag and the zone is too small it will clamp it to a larger size.

samueljackson92 (changes requested):
Looks good to me. I push a single line commit to make the minimum width a little smaller. 1% is too wide. 0.1% seems more reasonable to me.

There's a merge conflict with main. Can you do a `git merge main` into your branch and fix the conflict and this is ready to go.

.gitignore
```diff
@@ -172,3 +172,5 @@ data/elm-events
 node_modules
 package-lock.json
 package.json
+
+services/ui/public/frames/
```
jblake42: Is this necessary - I'd prefer us not to add stuff into the ignore file that are specific to one test that isn't included in this branch?

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -81,6 +81,11 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
 
       const xaxis = plot._fullLayout.xaxis;
 
+      // Minimum width in data units: 1% of current x-range
```
jblake42: Update the comment to reflect the current percentage or get rid of the specific value

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -90,15 +95,28 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
+                const right = Math.max(d.x0, d.x1);
+                d.x1 = right;
+                d.x0 = right - minWidth;
+              } else {
+                const left = Math.min(d.x0, d.x1);
+                d.x0 = left;
+                d.x1 = left + minWidth;
+              }
+            } else if (d.x1 < d.x0) {
+              const t = d.x0; d.x0 = d.x1; d.x1 = t;
```
jblake42, line 131: It does make more sense to have this logic carried out here rather than within `handleZoneDragFinish` as it currently is. I would say that if we are doing it here then the duplicated logic carried out in `handleZoneDragFinish` should be removed and similar logic should be put into the `end` handler when zones are created within the `zoneProvider`.

Message me for clarity if needed

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -90,15 +95,28 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
+            }
+            handleZoneUpdate();
```
jblake42: Do we think it is worth putting this update within the two logical checks to avoid unecessary updates - it is likely negligible in terms of performance but I'm not sure

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -145,12 +163,18 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
         const x1 = xaxis.d2p(zone.x1);
                 const x = Math.min(x0, x1)
                 const pointerEvent = disableToolingInteraction ? "none" : "all"
+        // render span using left-most x and absolute width (shows correctly during wrap)
+        const spanLeft = Math.min(x0, x1);
```
jblake42: What is the reason for this?

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -145,12 +163,18 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
+        const spanWidth = Math.abs(x1 - x0);
+        const handleWidth = Math.min(20, spanWidth / 2);
+
+        // Span (center drag target)
         graphGroup
           .append("rect")
-          .attr("class", "zone span cursor-grab disable-on-modifier")
+          .attr("class", "zone span cursor-grab disable-on-shift")
```
jblake42: Is there a reason for changing this?

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -145,12 +163,18 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
         const x1 = xaxis.d2p(zone.x1);
                 const x = Math.min(x0, x1)
                 const pointerEvent = disableToolingInteraction ? "none" : "all"
+        // render span using left-most x and absolute width (shows correctly during wrap)
+        const spanLeft = Math.min(x0, x1);
+        const spanWidth = Math.abs(x1 - x0);
+        const handleWidth = Math.min(20, spanWidth / 2);
```
jblake42: I understand the rationale for this but I am concerned that users need to be very accurate to drag narrower zones. There is also a lack of consistency between how accurate they have to be for large zones vs small zones which may be off putting. I would probably recommend having a fixed allowable width ouside the zone and a variable amount inside to prevent overlap @samueljackson92 might have an opinion?

image

jblake42 (changes requested):
This is looking really good - few issues to take a look at.

 I personally think we should keep in the minimum width (0.1%) just to ensure we don't get zero width annotations as we have no way of removing them - @samueljackson92 what do you think?

We should then create a new MR for modifying the rendering if the zone is very small - change to a line for example

.gitignore
abdullah-ukaea: Yep, this is not needed. Removed in https://github.com/ukaea/viz-annotation/pull/85/commits/ba486470a611fab624cf108adda0846885f60453

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: addressed in https://github.com/ukaea/viz-annotation/pull/85/commits/85cf4762d26c764786573af1894e64141978fc00

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: From my end this was an accident from when I was experimenting with span class. I have reverted this back to normal in https://github.com/ukaea/viz-annotation/pull/85/commits/a874af7e3ab7207b47b4634074b8945b4edcf780

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: Ah I didn't realise I was computing x and span left but they were doing the same thing so spanLeft is redundant here. We currently already compute const x = Math.min(x0, x1) and use it as the rect's x anchor. I decided to keep spanLeft and removed x as it’s more descriptive and to avoid confusion. Addressed in https://github.com/ukaea/viz-annotation/commit/bd2c3a83564c69cdd9d3ed214e8a3c99b6990f2c

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: I have taken a look at zones.tsx and zone-provider.tsx and I think I have addressed this comment now.

I've moved the min-width enforcement and x0/x1 normalization into the zone edge end handler in zones.tsx, and now only call handleZoneUpdate when values actually change. I removed the bulk normalization from handleZoneDragFinish and added a lightweight orientation normalize in the provider’s creation end callback so new zones are consistent without duplicating logic. 

Please check Commits:https://github.com/ukaea/viz-annotation/commit/0f522753e81f64f86cf3a8488a1906979a06f786 and https://github.com/ukaea/viz-annotation/commit/50862ffca7e1d07fa27395807316bbe9487a87be

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: Please check Commits:https://github.com/ukaea/viz-annotation/commit/0f522753e81f64f86cf3a8488a1906979a06f786 and https://github.com/ukaea/viz-annotation/commit/50862ffca7e1d07fa27395807316bbe9487a87be

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: I’ve now implemented fixed “outer” hit-areas on both ends (20 px each) plus a small “inner” portion that shrinks as the zone gets narrower, leaving a minimum center gap so the middle stays draggable. The span is still drawn from the leftmost boundary with absolute width (so wrap works), and each handle is anchored to its true x0/x1 edge, so behavior is consistent even when reversed. Hopefully now this means tiny zones are easy to grab and the required precision is consistent across sizes. Commit: https://github.com/ukaea/viz-annotation/commit/2a32bb4bfd205ab0fbd08af8a70f7d9b358dfd18

abdullah-ukaea:
@jblake42  Hi Josh, this PR is now ready for re-review I have tried to address all your comments.

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -64,9 +53,10 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
     const handleZoneUpdate = () => {
         triggerZoneUpdate()
     }
-
+    // Orientation normalization is handled per-zone in the UI drag 'end' handler.
```
jblake42: I don't think we need these comments relating to logic we have taken out. It not being present is enough

services/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -107,8 +97,14 @@ export const ZoneProvider = ({categories, initialData, children, onModifyZone} :
                 triggerZoneUpdate()
             },
             end: (x, _y) => {
-                zones.current[zones.current.length-1].x1 = x;
-                handleZoneDragFinish()
```
jblake42, line 111: We still need to call this even though it doesn't do zone cleanup anymore - it triggers backend logic which is required on zone creation

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -102,21 +102,33 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
             if (width < minWidth) {
               if (isLeft) {
                 const right = Math.max(d.x0, d.x1);
-                d.x1 = right;
-                d.x0 = right - minWidth;
+                const newX0 = right - minWidth;
+                if (d.x1 !== right || d.x0 !== newX0) {
+                  d.x1 = right;
+                  d.x0 = newX0;
+                  changed = true;
```
jblake42: I feel like setting `changed` can go within the `if (width < minWidth)` check can we know that if this is the case then the zone will be changing - this gets rid of the need to check twice for left and right.

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -102,21 +102,33 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
           .on("end", function (_event, d) {
             // on end: enforce min width and normalize orientation
+            let changed = false;
             const width = Math.abs(d.x1 - d.x0);
             if (width < minWidth) {
               if (isLeft) {
                 const right = Math.max(d.x0, d.x1);
-                d.x1 = right;
-                d.x0 = right - minWidth;
+                const newX0 = right - minWidth;
```
jblake42: This new logic does not appear to do what I would expect. We want the handle that the user is dragging to be the one that adjusts - automatically changing the handle the user has not touched seems like it could be a nuisance. I have mocked up what I expect below (with an inflated minimum width), I haven't included the code so that you can have a go but let me know if you want to go through my implementation.



With all said an done, from tests with 0.1% this is all unoticable so we can probably just get rid of any of the logic and go back to how you were intially doing it, however if you want to have the logic in there it should be behave as above imo.

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -159,20 +171,35 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
+        const px0 = xaxis.d2p(zone.x0);
+        const px1 = xaxis.d2p(zone.x1);
+        const pointerEvent = disableToolingInteraction ? "none" : "all"
+        
+        // render span using left-most x and absolute width, span in pixels
+        const spanLeft = Math.min(px0, px1);
+        const spanRight = Math.max(px0, px1);
+        const spanWidth = spanRight - spanLeft;
+
+        // handle layout: fixed outside strip + variable inside strip
```
jblake42, line 189: I think we can get away with smaller handle catchment areas (maybe 10px each side from experimentation). They are pretty large at the moment meaning the chance of overlaps causing issues is more likely:
image

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -159,20 +171,35 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
+        ));
+        const totalHandleWidth = OUTER_HANDLE_PX + inner;
+
+        const x0IsLeft = px0 <= px1;
 
         // Span (center drag target)
         graphGroup
           .append("rect")
-          .attr("class", "zone span cursor-grab disable-on-shift")
-          .attr("x", x)
```
jblake42, line 151: Nice, I think I was being silly here with duplicated logic

jblake42 (changes requested):
Really nice work, the fixed external handle bit looks really good and just a couple of points regarding the min-width bits

services/ui/src/app/components/providers/zone-provider.tsx
abdullah-ukaea: Fixed in commit https://github.com/ukaea/viz-annotation/commit/b7ecc191a69f55f57e65e3cabcb70b57f3643185

services/ui/src/app/components/providers/zone-provider.tsx
abdullah-ukaea: fixed in commit https://github.com/ukaea/viz-annotation/commit/8d708eca7e8e54dbcfe8a5ec06e995641528e782

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: I have tried to tackle this in commit https://github.com/ukaea/viz-annotation/commit/7b2c2fdd0e9d4f1d77bdea05b6db378165adc029

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: should be resolved in https://github.com/ukaea/viz-annotation/commit/7b2c2fdd0e9d4f1d77bdea05b6db378165adc029

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: should be resolved in https://github.com/ukaea/viz-annotation/commit/fa66eaf3858e45aef7116f30ecd825cce36b2bb1

abdullah-ukaea:
ready for rereview @jblake42

services/ui/src/app/components/tools/zones.tsx
```diff
@@ -95,37 +95,45 @@ export const Zones = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
+                } else {
+                  d.x0 = d.x1 - minWidth;  // normal case → clamp on the left side of x1
+                }
               } else {
-                d.x0 = left;
-                d.x1 = left + minWidth;
+                // Symmetric logic for right boundary relative to fixed left boundary (x0)
+                if (d.x1 < d.x0) {
+                  d.x1 = d.x0 - minWidth // wrapped past left → clamp on the left side of x0
+                  d.x1 = d.x0 + minWidth // normal case → clamp on the right side of x0
```
jblake42: This should be in an `else` statement

jblake42 (dismissed):
One issue needs addressing

services/ui/src/app/components/tools/zones.tsx
abdullah-ukaea: fixed in https://github.com/ukaea/viz-annotation/commit/9168edf793bc054df8a6bcb8e3785a36e1ef414f

samueljackson92:
Nice work here. I tried out the zones editing, and trying to make them very small. The UI interaction felt natural to me.I think this is good to go.

samueljackson92 (approved):

PR #86, opened 2025-08-06
Spectrogram improvements

rorlandoc:
- [x] Improve log scale colorbar
  - Add major power of 10 ticks and tick labels
  - Add minor ticks for values of n * 10^p (n = 2, .., 9)
  - Range slider now clamped to a min value of 1E-4 (same value as spectrogram plot) to remove a "dead zone"
- [x] Add segmentation features
  - Added threshold toolbox and modified data schemas
  - Simple thresholding based on percentile value
- [x] Add colormap selection
  - Also implements the idea of `PlotProps` for parameters that influence the look of a plot but leave the data unchanged - thus not requiring a back end connection - I believe more thought needs to go into this as this should be a structure that changes per plot and currently is being set from a page level (meaning that any plot within a page would see the same props object)

~Depends on #45 (merge this first)~


services/api/core/views.py
```diff
@@ -86,10 +92,18 @@ def convert_timeseries_to_spectrogram(
         ds = ds.sel(frequency=slice(frequency_min, frequency_max))
         ds = ds.clip(amplitude_min, amplitude_max)
 
+        ds_threshold = xr.DataArray(
+            threshold_values, coords=dict(frequency=freq, time=time)
+        )
+        ds_threshold = ds_threshold.sel(time=slice(time_min, time_max))
+        ds_threshold = ds_threshold.sel(frequency=slice(frequency_min, frequency_max))
+        ds_threshold = ds_threshold.clip(0, 1)
```
samueljackson92: I am not sure this makes sense here. I think in theory producing a segmentation like this should be moved to an annotator model. Similar to how we do peak detection [here](https://github.com/ukaea/viz-annotation/blob/a7bc9b050d212fbbfcf3e3e56007253c1d90c5bd/services/api/core/annotators.py#L19), except instead of returning a list of time regions, we return a mask.

That would suggest we need a new type of [annotation](https://github.com/ukaea/viz-annotation/blob/a7bc9b050d212fbbfcf3e3e56007253c1d90c5bd/services/api/schemas/annotations.py#L1), but a simple mask could potentially be quite big.

Therefore I suggest we adopt what they do for [COCO](https://cocodataset.org/#home) and use run-length encoding (RLE) to represent the mask as a string OR use a polgon list. For this PR I suggest we just do RLE.

This would mean we need a little bit of javascript to decode it client side, but it shouldn't be too hard.

samueljackson92 (changes requested):
Good start! I have not reviewed the code changes too much, just the UI. There's a couple of improvements that we can make

1. I think if the thresholding tool is active, then the slider should automatically update the plot when it is adjusted, rather than the user having to click apply each time.
2. For consistency with the time series view I would move the Amplitude Range and Thresholding tools to disclosures in an [accordian](https://react-spectrum.adobe.com/react-spectrum/Accordion.html#accordion). See [here](https://github.com/ukaea/viz-annotation/blob/dae626e91c94350eadec9854dff99552e6dcea41/services/ui/src/app/components/tools/toolbar.tsx#L242) for an example.

rorlandoc:
Made some changes.

To @samueljackson92's feedback:
- Moved backend code to the annotator model. I think it is working but I need you and @wk9874 to have a look at it - I don't know how some of the schema metadata (e.g. validated, uncertainty, etc) were relevant to this use case. Also I didn't know how to handle clearing the mask - if I use `setAnnotations([])`, am I not clearing all types of annotations - including vspans and zones?
- Also @samueljackson92 regarding the accordion / disclosure: I agree but that isn't in main at the moment correct? I don't want to duplicate code

To @jblake42's feedback:
- Threshold toolbox is now a textbox with preset +-1 and +-5 buttons; Also @jblake42 if you can please have a look at the code;

To improve overall usability:
- Integrated the mask into the main plot using opacity to separate the two signals;
- Added some styling options to the plotly layout to play nice with dark mode (sorry I know this is unrelated).

RLE does not seem to be necessary - it feels snappy to me, but let me know if this is not the case and I'll implement it.

Screenshot 2025-08-14 at 17 26 36

services/api/core/views.py
rorlandoc: Partially addressed - RLE doesn't appear to be necessary for now? Please see my latest comment

services/api/core/annotators.py
```diff
@@ -54,14 +55,43 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
 
         return regions
 
+
+class ThresholdAnnotator:
```
samueljackson92: Maybe rename `SpectrogramThresholdAnnotator`

services/api/core/annotators.py
```diff
@@ -54,14 +55,43 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
+        values = np.array(data.values["mirnov"].values)
+
+        sample_rate = 1 / (time[1] - time[0])
+
+        _, _, values = stft(
+            values,
+            fs=int(sample_rate),
+            nperseg=256,
+            noverlap=128,
+        )
```
samueljackson92: Can we reuse the code in the `SpectrogramView` rather than duplicating the FFT here.?

services/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -231,11 +376,28 @@ export default function ToolBar({
...
+        <hr className="m-4 h-px opacity-30 border-gray-200" />
+        <AmplitudeSlider
+          data={mhdData.data}
+          viewParams={viewParams}
+          setViewParams={setViewParams}
+          plotProps={plotProps}
+          setPlotProps={setPlotProps}
+        />
+        <hr className="m-4 h-px opacity-30 border-gray-200" />
+        <ThresholdTool
```
samueljackson92: Again, this is for spectrograms, so use a more specific name.

samueljackson92 (dismissed):
Looking really nice and works well. A couple of comments above.

To further comment on your questions:

 >I don't know how some of the schema metadata (e.g. validated, uncertainty, etc) were relevant to this use case.

Don't worry for now.

> Also I didn't know how to handle clearing the mask - if I use setAnnotations([]), am I not clearing all types of annotations - including vspans and zones?

Later PRs have an additional `type` on the Annotations which specifies whether it's a time region, time point etc., and whether it is `created_by` (either a tool or a manual). I suggest you don't worry about this here for now.


> Also @samueljackson92 regarding the accordion / disclosure: I agree but that isn't in main at the moment correct? I don't want to duplicate code.

As above. Don't worry unless my PRs get merged before yours!

> Threshold toolbox is now a textbox with preset +-1 and +-5 buttons; Also @jblake42 if you can please have a look at the code;
> Integrated the mask into the main plot using opacity to separate the two signals;
> Added some styling options to the plotly layout to play nice with dark mode (sorry I know this is unrelated).

Looks good to me and works well.

> RLE does not seem to be necessary - it feels snappy to me, but let me know if this is not the case and I'll implement it. 

It's not just for the user. The mask needs to be stored as an annotation in a database, so there's a lot of redundancy here. I would go ahead and implement it.

samueljackson92:
Sorry it took me a few days to come back to you on this!

samueljackson92:
Add a unit test and fix a few last minor refactoring things. UKAEA CI is currently out of credits, so I've checked all the tests/linting for the project locally.

PR #88, opened 2025-08-07
Add a tool to export annotations

### LINKED ISSUE #44: Add a download/export annotations option to sample view.
We should add the ability to directly download the annotations for:
 - a particular sample 
 - all samples within this project.

samueljackson92:
Adding a tool to export annotations. The user can either 1) export all annotations in the project 2) export just the annotations for the current sample being viewed.

image

To test:
 - try out the new tool
 - code review.

closes #44


samueljackson92:
@wk9874 Exporting annotations works fine, but the annotations only have a `sample_id` in order to know what the `shot_id` for the annotations is I must also get the sample data. I think it would be nicer if `sample_id` was actually `shot_id`.

PR #91, opened 2025-08-07
Backend work for implementing ML Models

wk9874:
Adds backend changes to support ML Models
* Creates new set of `/models` endpoints
* Reintroduces Celery workers
* Adds in model abstract base class
* Adds in a dummy CNN on some fake disruption data as a test


scripts/setup.py
```diff
@@ -7,7 +7,7 @@ def create_project(name: str, task: str, data_loader: str) -> str:
     project = {
         "name": name,
         "task": task,
-        "query_strategy": "random",
+        "query_strategy": "uncertainty",
```
samueljackson92, line 10: Can we leave the testing projects are using random for now.
Only for the mock data project should we use uncertainty sampling.

scripts/create_mock_data.py
```diff
@@ -0,0 +1,66 @@
...
+                    "times": time.tolist()   
+                },
+            },
+            "annotations": {
+                "ramp_up_start": 0.02*t_ramp_start,
+                "ramp_up_end": 0.02*t_ramp_end,
+                "disruption": 0.02*t_disrupt_start
+            }
+        }
+
```
samueljackson92, line 47: Are we sure we want to add the ability to upload data like this? I understand it's just for testing, but I am not sure I like this. We could of just made a parquet file, or added CSV support. 

If we're adding JSON data support then we should think about how that should be structured in a little more detail. I guess we are just using the format that we are passing back from the API... which makes sense

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
...
+        self,
+        db_client: MongoDBClient,
+        db_id: ObjectId,
+        project: Project, 
+        samples: list[Sample], 
+        annotations: list[list[Annotation]], 
+        train_val_test_split: typing.Tuple[float, float, float]
+        ) -> None:
+        
+        if len(samples) != len(annotations):
```
samueljackson92: I don't think this is correct? Could be that: no. samples << no. annotations

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
...
+            raise ValueError("Annotations missing for some samples!")
+        if not abs(sum(train_val_test_split) - 1) < 1e-8:
```
samueljackson92: Consider [numpy.isclose](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html) or similar.

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
...
+            raise ValueError("Ratios in train_val_test split must sum to 1!")
+        if not train_val_test_split[0]:
```
samueljackson92: Don't do a zero check like this. It's not a bool type. Something like `train_val_test_split[0] == 0` is clearer to me.

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
...
+            raise ValueError("Must be samples in the training set!")
+        
+        self.db_client = db_client
+        self.db_id = db_id
+        self.project = project
+        self.model = self._define_model()
+        
+        # If train ratio is 1, no splitting required, just set train sets and return
+        if train_val_test_split[0] == 1:
```
samueljackson92: Maybe for neater code unpack as `train_size, val_size, test_size = train_val_test_split` and use those.

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
+
+from services.api.schemas.samples import Sample
+from services.api.schemas.annotations import Annotation, AnnotationIn, TimePoint
+from services.api.schemas.projects import Project
+from services.api.schemas.data import Data, TimeSeriesData
```
samueljackson92: Please remove unused imports.

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
...
+                test_size= train_val_test_split[2] / (train_val_test_split[1] + train_val_test_split[2])
+                )
+    
+    async def _update_progress(self, progress: float):
+        updated_model = ModelUpdate(progress=progress)
+        await self.db_client.update(collection="models", model=updated_model, object_id=self.db_id)
+                
+    @abstractmethod
+    def _define_model(self):
+        pass
```
samueljackson92: Consider raising `NotImplemented` instead of `pass`?

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
...
+        annotations: list[list[Annotation]],
+        train_val_test_split: typing.Tuple[float, float, float] = (0.7, 0.2, 0.1),
+    ) -> None:
+        super().__init__(db_client=db_client, db_id=db_id, project=project, samples=samples, annotations=annotations, train_val_test_split=train_val_test_split)
+        self.dataset = dataset
+        self.train_dataset = dataset(project, self.train_samples, self.train_annotations)
+        self.val_dataset = dataset(project, self.val_samples, self.val_annotations) if self.val_samples else None
+        self.test_dataset = dataset(project, self.test_samples, self.test_annotations) if self.test_samples else None
+    
+class DisruptionDataset(TorchDataset):
```
samueljackson92: Can we move implementation specific details to a separate file/folder/module?

services/api/core/models.py
```diff
@@ -0,0 +1,360 @@
...
+        
+class DisruptionCNN(TorchModel):
+    def __init__(
+        self,
+        db_client: MongoDBClient,
+        db_id: ObjectId,
+        project: Project, 
+        samples: list[Sample], 
+        annotations: list[list[Annotation]],
+        train_val_test_split: typing.Tuple[float, float, float] = (0.7, 0.2, 0.1),
```
samueljackson92: Good first attempt, but no, I don't like this. We are mixing logic about our application with logic about running a model. A DisruptionCNN model should not need to know about `db_client`. This class needs to be split. I suggest the following:

1. A `Runner` class that knows how about `db_client` and probably `train_val_test_split`. 
2. A `Model` class than knows about DisruptionCNN and how to run it.

`Model` will need to know about samples & annotations for setting up dataloaders.
There should be some way for `Model` to callback to `Runner` to update training, or write predictions to the database etc.

samueljackson92: Think like you are a model creator who does not know anything about our system. I should not need to know what a `db_id` or `db_client` is to integrate my model. The coupling should be a loose as possible.

services/api/routers/models.py
```diff
@@ -1,114 +1,247 @@
...
+    model = {"type": model_type, "version": version, "training_status": "queued", "accuracy": 0, "progress": 0}
+    model_id = await db_client.insert(
+        collection = "models",
+        model = ModelIn(**model),
+        ids = {"project_id": ObjectId(project.id)}
+    )
+    model["project_id"] = project.id
+    model["id"] = model_id
+    # Start task with ID of this project? How will we know whether training is running? dont want multiple trainings at once? TODO
+    run_training.delay(project.model_dump(mode="python"), model)
```
samueljackson92, line 127: Separate the API from the update logic and put a method in `utils` or similar.

services/api/routers/samples.py
```diff
@@ -172,9 +172,12 @@ async def get_next_sample(
...
     project = await utils.get_project(db_client, project_id)
-    samples = await utils.get_samples(db_client, project_id)
-    print(samples)
-    annotations = await utils.get_annotations(db_client, project_id, validated=False)
+    
+    # Only consider samples that have not been human annotated
+    samples = [Sample(**sample) for sample in await utils.get_samples(db_client, project_id, validated=False, sort_by="shot_id")]
+    annotations = [Annotation(**annotation) for annotation in await utils.get_annotations(db_client, project_id, validated=False, sort_by="uncertainty")]
+    print(f"found {len(samples)} non validated samples")
+    print(f"found {len(annotations)} non validated annotations")
```
samueljackson92, line 200: Probably should remove these debug statements or make use of proper logging.

services/api/core/models.py
samueljackson92: Furthermore, we should consider breaking apart the logic for setting up data loading and actual training. Like I can see a interface for `Model` that looks something like:

```python
def train(train_dataset: TorchDataset):
    ...
def predict(dataset: TorchDataset):
    ... 
```

Or possibly:

```python
def train(train_samples: list[Samples]):
    ...
def predict(dataset: list[Samples]):
    ... 
```

samueljackson92 (changes requested):
@wk9874 some comments to get you started. My main concern here is the tight binding between model code and our database logic.

samueljackson92:
Also this branch needs a merge with main now :) ...

services/api/core/models.py
wk9874: True, but this is because `annotations` is type `list(list(Annotation))`, where the outer list is the length of `samples`, and each list corresponding to a single sample can have multiple annotations stored within

services/api/core/models.py
wk9874: Don't think this is required since it is decorated as an `abstractmethod`, so will automatically throw an error if a class doesnt override it - feel like using `pass` in that case is the standard way to do it in Python? (and is the way it is defined in eg `QueryStrategy`)

services/api/core/models.py
wk9874: Yeah that's fair, I also hated having to introduce DB logic into the Model class but couldn't think of a better way at the time and wanted to get something working!

Runner class sounds like a good idea - will look into that

samueljackson92: Yea completely understand. About 80% of this PR is good, but it's hard to see until you throw something down.

wk9874:
Going to close this, have addressed review comments in branch `wk9874/models_frontend` and will raise a new PR soon

PR #93, opened 2025-08-08
Remove NEXT_PUBLIC_API_URL as it causes CORS issues

### LINKED ISSUE #67: UI breaks when using 127.0.0.1 instead of localhost
When spinning up the server with docker compose and logging into the web UI at `127.0.0.1:3001`, the UI breaks with this error:

Image

Looking in the browser console log you see this:

```
Access to fetch at 'http://localhost:3001/backend-api/projects/?sort_by=_id&sort_direction=descending&start=0&count=5' from origin 'http://127.0.0.1:3001' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

I think this is because inside the docker compose, `NEXT_PUBLIC_API_URL: "http://localhost:3001"`. Even though `127.0.0.1` and `localhost` point to the same place, looks like the redirection has been blocked because the browser thinks they are different?

Not that much of an issue for now since you can just use `localhost` in the browser, but will probably cause a problem for #39

wk9874:
Not sure if this is an appropriate solution, but it seems to work...

The code in `services/ui/next_config.js` already overwrites the location of the backend API to point to the docker container as far as I can tell, so dont think we need to provide anything to `NEXT_PUBLIC_API_URL`. It causes CORS issues when trying to load the web UI on `127.0.0.1` or on the VM, since it tries to point to localhost from a different origin. Tried adding CORSMiddleware from FastAPI but that also doesnt work, I think since the browser doesnt add an `Origin` header for some reason and therefore CORSMiddleware doesnt kick in....

Closes #67

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,7 +62,7 @@ services:
       PORT: 3001
       NODE_ENV: development
       DATA_API_URL: "http://api_app:8002"
-      NEXT_PUBLIC_API_URL: "http://localhost:3001"
+      NEXT_PUBLIC_API_URL: ""
     volumes:
       - ./services/ui:/app
       - /app/node_modules
--- a/services/api/crud/utils.py
+++ b/services/api/crud/utils.py
@@ -6,6 +6,7 @@ from services.api.schemas.annotations import Annotation
 from services.api.schemas.projects import Project
 from services.api.schemas.samples import Sample
 
+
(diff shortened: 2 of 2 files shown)


samueljackson92 (approved):
Looks fine to me. 

When we come to do CI with your unit tests we should also include a ruff linter check for python code. Perhaps should also see what we equivalent would be for react/typescript.

PR #95, opened 2025-08-18
Backend API Tests

wk9874:
Adds backend API tests for:

* All endpoints
* Data loaders
* Query strategies
* Annotators (only find peaks since others aren't merged yet)
* Mongo DB Client
* DB utils

Have made some refactoring changes which fix issues spotted during writing these tests, but some things (eg that get_samples and get_annotations return dicts, not objects) have not been fixed since I know they are fixed in existing PRs.

Only one test fails - Find Peaks annotator. I believe the issue with this is now fixed in the time series annotators PR (?)


samueljackson92:
@wk9874 there's a merge conflict that needs fixing here.

samueljackson92 (changes requested):
This needs some instructions on how to setup and run the test. Preferably in the dev docs.

Furthermore, I found that trying to run pytest I had to additional install:

```
pytest-asyncio
httpx
```
I then get the error:

```
ImportError while loading conftest '/Users/rt2549/projects/viz-annotation/tests/conftest.py'.
tests/conftest.py:6: in <module>
    from testcontainers.mongodb import MongoDbContainer
E   ModuleNotFoundError: No module named 'testcontainers'
```

wk9874:
@samueljackson92 have added docs and added requirements into pyproject.toml, please give another go!

samueljackson92 (changes requested):
I don't have UKAEA VPN access here, so the tests with UDA fail. Can we maybe add a `skipif` check if uda can't pull data? This will be needed for CI deployment as well.

wk9874:
@samueljackson92 They should be added as dev requirements inside the pyproject.toml - did you install it with `uv sync` as written in the docs?

With UDA, yeah I was also thinking we should add a skip. Is there some way to test the connection with pyuda (besides just trying to pull data from a pulse)? Maybe a version check or something?

samueljackson92:
@wk9874 Sorry, I missed that I had not checked out the lastest commit. `uv sync` does indeed work.

samueljackson92:
@wk9874 the UDA team suggest that you can do a `client.get("help::help()")` to minimally check connection to the server is successful.

wk9874:
Added pytest skip if UDA not contactable as above :)

tests/api/core/test_data_loaders.py
```diff
@@ -0,0 +1,86 @@
...
+    
+    # Ip should have 4 values of 1000
+    assert len(numpy.where(numpy.isclose(ip_values, 1000))[0]) == 4
+    
+def test_uda_loader(uda_env_vars):
+    try:
+        import pyuda
+        pyuda.Client().get("help::help()")
+    except Exception:
+        pytest.skip("Could not contact UDA server")
```
samueljackson92, line 54: Consider using a skipif:
https://docs.pytest.org/en/stable/how-to/skipping.html#id1

samueljackson92 (changes requested):
Still getting some test failures:
image

tests/api/core/test_data_loaders.py
wk9874: How would skipif be better? Still needs the try/except right since it has to catch the error thrown by pyuda?

wk9874:
> Still getting some test failures: image

Have fixed the uncertainty query strategy one

With respect to the FindPeaks one, I think this one is currently expected to fail and should (hopefully?) pass with the changes I asked for in the latest round of annotators PR. Could you take a look at the test and check for me if I've got the expected behaviour right with the specified parameters?

```
def test_find_peaks_params():
    # Create data with clear peaks above noise
    data = numpy.random.uniform(0, 10, 100)
    data[5] = 1000
    data[30] = 500
    data[35] = 500
    data[40] = 500
    data[60] = 300
    data[80] = 100
    data[95] = 1000
    
    ts_data = TimeSeriesData(time=numpy.arange(100), values=data)
    mv_data = MultiVariateTimeSeriesData(values={"Ip": ts_data})
    
    small_params = FindPeaksParams(signal_name="Ip", prominence=50, distance=1)
    # Create an annotator with the above params - this should detect all 7 peaks
    full_annotator = annotators.FindPeaksAnnotator(small_params)
    time_regions = full_annotator.predict(mv_data)
    assert len(time_regions) == 7
    
    # Then create an annotator with more limiting params
    params = FindPeaksParams(signal_name = "Ip", prominence=200, distance=10, time_min=10, time_max=90)
    annotator = annotators.FindPeaksAnnotator(params)
    time_regions = annotator.predict(mv_data)
    
    # Based on data above with these params:
    #   - Should recognise peaks at idx 30, 35, 40 as one peak since they are within 'distance'
    #   - Should find peak at idx 60
    #   - Should not find peak at idx 80 since this is below 'prominence'
    #   - Should not find peak at idx 5 since this is before 'time_min'
    #   - Should not find peak at idx 95 since this is beyond 'time_max'
    assert len(time_regions) == 2
    
    assert time_regions[0].time_min < 30 and  time_regions[1].time_min > 40
    time_mid = (time_regions[0].time_min + time_regions[0].time_max) / 2
    assert math.isclose(time_mid, 35, abs_tol=1)

    
    assert time_regions[0].time_min < 60 and  time_regions[1].time_min > 60
    time_mid = (time_regions[0].time_min + time_regions[0].time_max) / 2
    assert math.isclose(time_mid, 60)
```

tests/api/core/test_data_loaders.py
samueljackson92: I was thinking something like:

```python
def has_uda() -> bool:
    try:
        import pyuda
        pyuda.Client().get("help::help()")
        return True
    except Exception:
        return False

...
@pytest.mark.skipif(has_uda(), reason="Cannot connect to UDA")
def my_test():
    pass
```

PR #96, opened 2025-08-19
[STALE] Annotation Toolbar and Edit Mode

jblake42:
This PR introduces **edit mode** and an **annotation tool bar**

### EDIT MODE
To prevent users from accidently changing annotations while viewing data, edit mode has been introduced. When not in edit mode (view mode) all annotation interactions are disabled allowing users to safely explore the data without the risk of editing existing annotations.

When in edit mode, all interactions are enabled including moving and resizing annotations, changing annotation types and adding new annotations either through the draw feature or through the context menu.

### ANNOTATION TOOLBAR
The annotation toolbar is a dynamic component that allows annotation tooling modes to be selected, allowing for the annotation to be added using the draw functionality. The toolbar will dynamically query all registered tooling from the annotation provider and present the tooling to the user, meaning developers need only inlcude the toolbar within the desired view - noting that it must be contained within an annotation provider like many of the other support components.

The annotation provider also provides a way to display and toggle the edit mode status.

### Remaining tasks
This PR focuses on the funcitonality of edit mode and the annotation toolbar, there is still work to be done on the styling of the toolbar


services/ui/src/app/components/tools/annotationToolbar.tsx
```diff
@@ -0,0 +1,53 @@
...
+  } = useAnnotationProvider();
+
+  const modeVariant: "accent" | "primary" = editMode ? "accent" : "primary";
+  const modeText = editMode ? "Edit Mode" : "View Mode";
+
+  return (
+    <div className="flex flex-col w-50 items-center space-y-3 ">
+      <h1 className="text-4xl font-bold text-center text-gray-900">
+        Annotation Toolbar
+      </h1>
```
samueljackson92, line 22: Can we move to using the equivilent items from react spectrum to keep the theme consistent?

For example:
https://react-spectrum.adobe.com/react-spectrum/Header.html
https://react-spectrum.adobe.com/react-spectrum/Flex.html#flex

etc.

services/ui/src/app/components/plots/time-series.tsx
```diff
@@ -1,6 +1,6 @@
 "use client";
 
-import { useContextMenuProvider } from "@/app/components/providers/annotation-provider";
+import { useAnnotationProvider } from "@/app/components/providers/annotation-provider";
```
samueljackson92, line 3: Should we call this `AnnotationProvider`. Would `ToolingProvider` be more appropriate?

services/ui/src/app/components/tools/annotationToolbar.tsx
samueljackson92: I know you want to enhance this later, but can you make the title a little nicer for dark mode, and maybe a bit smaller?:
image

services/ui/src/app/components/providers/annotation-provider.tsx
```diff
@@ -84,31 +109,61 @@ export const ContextMenuProvider = ({
...
       if (event.key === "Shift" || event.key === "Control") {
         keyHeldRef.current = false;
-        setDisableToolInteraction(false);
+        if (editMode) {
+          setDisableToolInteraction(false);
+        }
+      }
+    };
+
+    const handleEditMode = (event: KeyboardEvent) => {
```
samueljackson92, line 118: Can you please add a short documentation page, highlighting the usage and also documenting the keyboard shortcuts for this?

samueljackson92 (changes requested):
I think this is great work, and I know you're planning to refine it further, so I won't push for lots of extra stuff.

However, a few minor comments before it can go in.

samueljackson92:
@jblake42 is the surpassed by #177 ?Can we delete this branch and close the PR?

PR #97, opened 2025-08-19
Add CI config

samueljackson92:
Add a CI config for github. This adds two jobs:

1. Add a CI job which runs the new unit tests with `pytest`
2. Add a ruff job which check the quality of the python code

Once merged these will be enforced for all PRs


wk9874:
Looks good to me - but could we also add a `.pre-commit-config.yaml` as part of this PR which runs the ruff linter before commit? 

I also personally find pdb hook very useful, since I accidentally leave these dangling in the code all the time

```
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.12.9
    hooks:
      - id: ruff
        args: [ --fix, --exit-non-zero-on-fix, "--ignore=C901" ]
      - id: ruff-format
  - repo: https://github.com/conorfalvey/check_pdb_hook
    rev: 0.0.9
    hooks:
      - id: check_pdb_hook
        pass_filenames: false
```

And any others you think would be useful!

samueljackson92:
@wk9874 I also think it's worth adding a `ESLint` and `Prettier` check before this is merged.

samueljackson92:
@wk9874 I've added the suggested pre-commit hook, and added support for eslint and prettier checks in the CI. As you can see, all are now passing.

Note: for some of the issues in `time-series.tsx` I had to skip the check. Plotly doesn't actually specify a type which includes the right elements and you get all sorts of mad type errors when I try to resolve.

Give it a retest when you can?

samueljackson92:
@wk9874 ok, check if this has improved things?

wk9874:
Looks good to me!

wk9874 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,105 @@
...
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  lint:
+    name: Ruff Lint
+    runs-on: ubuntu-latest
```
wk9874, line 13: We should add a short timeout to all steps in case it gets stuck again :)

wk9874 (changes requested):

.github/workflows/ci.yml
samueljackson92: Done

wk9874 (approved):
lgtm!

PR #102, opened 2025-09-02
Hotfix for setup script and UI

wk9874:
In the PR adding the tests, I changed the create new samples endpoint from PUT to POST. Forgot to change the setup script, so this was just getting 405 errors back when trying to create samples

Also erroneously changed the update annotations endpoint from PUT to POST in the UI code - this annotations endpoint remained a PUT since it deletes the old annotations before creating new ones. Fixed this in the UI code

This demonstrates that we really need full end-to-end tests of the UI and backend API - raised https://github.com/ukaea/viz-annotation/issues/101

### DIFF
--- a/scripts/setup.py
+++ b/scripts/setup.py
@@ -32,7 +32,7 @@ def create_uda_samples(project_id: str, shot_ids: list[int]):
         }
         samples.append(sample)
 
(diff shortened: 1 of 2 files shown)


samueljackson92 (approved):
LGTM

PR #107, opened 2025-09-16
Add Models training & prediction backend + UI

wk9874:
Adds schemas, routers, celery worker tasks, and UI elements for model training and predictions

TODO:
* Investigate options which do not require Redis
* Investigate websockets instead of polling
* Add tests for model endpoints
* Run UI linting locally


services/ui/src/types.ts
```diff
@@ -108,6 +109,19 @@ export const SampleSchema = z.object({
...
 export type Sample = z.infer<typeof SampleSchema>;
 
+export const ModelSchema = z.object({
+  _id: z.string(),
+  timestamp: z.string(),
+  project_id: z.string(),
+  type: z.string(),
+  training_status: z.string(),
+  progress: z.number(),
+  accuracy: z.number(),
```
samueljackson92, line 119: This should be more generic. Maybe `metric` or `score`

services/api/routers/models.py
```diff
@@ -1,114 +1,425 @@
...
-@router.get("/{model_id}/train")
-async def get_training_info(project_id: str, model_id: str):
-    # Get current status of model training
-    pass
+    # Delete from DB
+    for model in models_to_delete:
+        await request.app.state.db_client.delete_filtered_documents(
+            collection="models",
+            filters={"_id": model["_id"], "project_id": project_obj_id},
+        )
```
samueljackson92, line 93: Maybe we need a `utils.delete_models`?

services/api/routers/models.py
```diff
@@ -1,114 +1,425 @@
...
+    annotations = await utils.get_annotations(db_client, project.id, validated=True)
+    samples = await utils.get_samples(db_client, project.id, validated=True)
+
+    # Add the training task to the queue
+    task = run_training.delay(
+        project.model_dump(mode="python"),
+        model,
+        [sample.model_dump(mode="python") for sample in samples],
+        [annotation.model_dump(mode="python") for annotation in annotations],
+    )
```
samueljackson92, line 174: Can we refactor this code out of the endpoint? Perhaps a `ModelRunner` class?
I think we should try and separate the endpoint definitions from the underlying logic. Rather than mixing it all together.
Consider this for other endpoints

services/ui/src/app/components/tools/modelPredictSample.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    setAnnotations((previousAnnotations) => { // add type for annotations, wait for Sams PR
+                        const otherAnnotations = previousAnnotations.filter((annotation) => annotation.created_by !== selectedModel);
+                        return otherAnnotations.concat(payload);
+                    });
+                    clearInterval(interval)
+                    setIsLoading(false);
+                } else {
+                    setErrorMessage(payload.detail)      
+                    clearInterval(interval)
+                    setIsLoading(false);
```
samueljackson92, line 78: Maybe break this out into a separate file for handling model stuff.

services/api/worker.py
```diff
@@ -0,0 +1,162 @@
...
+        annotations_2d = [
+            [ann for ann in group]
+            for _, group in itertools.groupby(
+                annotations, key=lambda annotation: annotation.sample_id
+            )
+        ]
+
+        # TODO: Where should epochs, batch size be passed in???
+        BATCH_SIZE = 32
+        NUM_EPOCHS = 10
```
samueljackson92, line 55: Train endpoint should receive a `TrainParameters` class with these items?

services/api/models/base.py
```diff
@@ -0,0 +1,165 @@
...
+class Model(ABC):
+    def __init__(
+        self,
+        model_id: str,
+        project: Project,
+        samples: list[Sample],
+        annotations: list[list[Annotation]],
+        train_val_test_split: typing.Tuple[float, float, float],
+        num_epochs: int = 100,
+    ) -> None:
```
samueljackson92, line 22: This is better than before, but I still don't like this.

It forces the user to modify their model code quite a bit to fit with our framework. I would like to see this more decoupled...

Given the direction you're heading, I would like to see how Ray proposed to do model serving. Maybe it makes this class somewhat redundant?

services/api/worker.py
```diff
@@ -0,0 +1,162 @@
...
+    annotations: list[AnnotationOutTypes],
+):  # TODO: do we want to support retraining where we only get annotations not previously put into model?
+    try:
+        print(f"Running model training for project {project.id}")
+        model_dir = pathlib.Path(os.environ["MODEL_STORAGE"])
+        model_dir.mkdir(exist_ok=True)  # Do i need to do this every time?
+
+        # Get all validated samples and annotations for this project
+        print(f"Collected {len(annotations)} annotations.")
+        print(f"Collected {len(samples)} samples.")
```
samueljackson92, line 43: Maybe we need to adopt a logging framework? I quite like `loguru`

samueljackson92 (changes requested):
A few comments before I fly to keep you going.
I want to see how this changes when using Ray and how Ray does model serving?
https://docs.ray.io/en/latest/serve/develop-and-deploy.html
https://docs.ray.io/en/latest/train/getting-started-pytorch-lightning.html

services/api/models/base.py
wk9874: I'm not convinced whether Ray will help us here... Ray Serve is cool, but as far as I can tell it basically requires you to define a deployment class like:

```

@serve.deployment
class MyFirstDeployment:
    # Take the message to return as an argument to the constructor.
    def __init__(self, msg):
        self.msg = msg

    def __call__(self):
        return self.msg
```

And then sort of binds your model to an endpoint, and whenever that endpoint is pinged it does the stuff in `__call__`

I so far haven't used Ray Serve and have just stuck to base Ray. I'll get that merged in later and you can see what it looks like - but it hasn't changed the base structure of this aspect of the code very much. So maybe I need to rethink what format this base class should be in / how the model should receive and output data

Any suggestions are very welcome :)

wk9874:
@samueljackson92 I've created a new PR for the changover to using Ray and the refactoring which came along with that - thought it might be easier for you to review to see how things changed :) 

https://github.com/ukaea/viz-annotation/pull/108

Or if easier, just merge that straight into this branch, and review it all at the same time!

samueljackson92:
@wk9874 Is #108 an extension of this branch? In which case I may just review #108 and delete this one?

wk9874:
> @wk9874 Is #108 an extension of this branch? In which case I may just review #108 and delete this one?

Yep it is - I can fix up the comments from this PR on that branch, then can close this PR and change the branch #108 merges into to main if that's easiest

services/api/routers/models.py
wk9874: Addressed in `wk9874/ray`

services/ui/src/types.ts
wk9874: I'd be inclined we keep this specific, as otherwise we wont know how to display it in the UI? Eg if we had 'score' and the person making the model could either provide accuracy or loss (or MSE or whatever else), then we have no way of displaying this consistently across all models in a table in the UI

We could add 'loss' in addition to accuracy?

samueljackson92: I appreciate that it's ambiguous, but this needs to change. Accuracy doesn't make sense for regression problems and in classification problems one typically cares about other metrics more such as F1, IoU, precision/recall scores. Loss would be at least consistent, but is less human interpretable.

samueljackson92:
@wk9874 Sounds good. I shall review the changes in the other PR.

PR #114, opened 2025-11-05
Add extensible data loaders

wk9874:
Adds the ability for users to add their own data loaders to the server after pip installing the package.

1. Adds a loader registry class which sets the available loaders globally
2. Changes validation on the Projects schema to not use an Enum, but to validate based on the loaders available in the registry
3. Refactor main.py and cli.py so that Server is now in its own class and can be called by the user in a script
4. Rename 'services' to 'viz-annotation' so that the package can be accessed easily once pip installed
5. Add basic user documentation


.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         run: uv sync
 
       - name: Run Ruff
-        run: uv run ruff check services tests --ignore=C901 && uv run ruff format --check services tests  
+        run: uv run ruff check viz_annotation tests --ignore=C901 && uv run ruff format --check viz_annotation tests
```
samueljackson92: Why did we rename the package here? Yes it would be better if it was called something else, but `viz_annotation` is not an improvement.

services/api/core/data_loaders.py
```diff
@@ -19,6 +23,34 @@ def get_sample(self, sample: Sample) -> Data:
         pass
 
 
+class LoaderRegistry:
```
samueljackson92, line 21: I like this pattern and having a decorator is good 👍

.github/workflows/ci.yml
wk9874: Made decision in meeting to change this to `toktagger`

samueljackson92:
[like]  Jackson, Samuel (Computing) reacted to your message:
________________________________

PR #118, opened 2025-11-13
Extensible Data Loaders

wk9874:
Adds extensible data loaders

New PR but same code, since I split the renaming of the module into a separate PR (and had to fix a bunch of weird git merge issues)


toktagger/api/core/project.py
```diff
@@ -30,11 +29,11 @@ def set_data_pool(self):
...
     def get_data_loader(
-        self, samples: List[Sample], data_loader_type: DataLoaderType
+        self, samples: List[Sample], data_loader_type: str
     ) -> DataLoader:
-        if data_loader_type == DataLoaderType.PARQUET:
+        if data_loader_type == "parquet":
             data_loader = ParquetDataLoader(samples)
-        elif data_loader_type == DataLoaderType.UDA:
+        elif data_loader_type == "uda":
             data_loader = UDADataLoader(samples)
```
samueljackson92: Why did we remove this enum? We can change the return type, but we should keep the enum...

samueljackson92 (changes requested):
One small comment....Otherwise this is good to go...

toktagger/api/core/project.py
wk9874: That was removed because the available DataLoaders are no longer fixed, since the user may have added new ones. And since enums are immutable, it no longer makes sense to keep this (imo). Its replaced by looking up the string in the registry and finding the associated DataLoader class. This if statement combo should be replaced by looking up in the registry, as it does in the endpoints...

However more general comment, as far as I know, we dont actually use this Project class anywhere? Should we just remove the entire thing?

samueljackson92: Yea... actually just remove.

Issue #54, opened 2025-06-29
Don't automatically save annotations on Next / jump to shot


wk9874:
Currently when annotating a sample, if you press Next or search for another shot without pressing save, the current annotations will automatically be saved. I don't think we should do this without warning the user as they may be moving to a next shot intentionally without saving the annotations from the current one

Could add a popup box or something if the user presses Next / searches without pressing save which asks them if they want to save their annotations, or continue without saving?

samueljackson92:
I think we should just save automatically when the user click next.

wk9874:
Im not convinced we should silently autosave when someone presses previous/next buttons without pressing save... what if someone is just flicking between different shots to see what the auto annotator tools look like? Theyll unknowingly be adding 'validated' annotations into the database and marking shots as validated

Maybe it could popup with 'save shots'? With a yes/no, and a 'set default / dont ask again' option?

samueljackson92:
In #120 I added a checkbox to validate on save, which is essentially the same feature. I don't think the user should be asked every time or need to always click "Save" then "Next".

samueljackson92:
Address to a certain extent in #120

PR #120, opened 2025-11-19
Improved time series annotation interface

samueljackson92:
This makes a number of changes to improve the time series annotation interface.

1. Adding better zooming functionality. User can now toggle "edit mode"/ "zoom mode" and scroll zoom on traces
2. Annotator settings are persistent between shots
3. Batch select, edit, and deletion of annotations has been implemented
4. Import and export annotation buttons have been added
5. Previous/next navigation buttons have been added.


toktagger/api/core/annotators.py
```diff
@@ -234,7 +234,7 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
             peak_time = time[idx]
             if peak_time >= tmin and peak_time <= tmax:
                 region = TimeRegion(
-                    label="Peak",
+                    label="Unknown",
```
wk9874, line 237: Why have these changed to unknown?

toktagger/api/core/query_strategy.py
```diff
@@ -8,28 +10,49 @@
...
+        previous_index = index - 1
+
+        if previous_index < 0:
+            raise RuntimeError("No previous sample available!")
 
-        index = random.randint(0, len(self.samples) - 1)
-        return self.samples.pop(index)
+        return self.samples[previous_index]
+
+    def _get_matching_sample(self, current_sample_id: Optional[str]) -> int:
```
wk9874: Why is `current_sample_id` optional here, but required in higher level functions which call this?

toktagger/api/core/query_strategy.py
```diff
@@ -38,38 +61,52 @@ class SequentialQueryStrategy(QueryStrategy):
...
+        super().__init__(samples, annotations)
+        # simply shuffle the samples at the start
+        # seed is used to ensure consistent shuffling between calls
+        self._random_shuffle_samples(seed)
+
+    def _random_shuffle_samples(self, seed: int):
+        idx = np.arange(len(self.samples))
+        rng = np.random.default_rng(seed=seed)
+        rng.shuffle(idx)
+        self.samples = [self.samples[i] for i in idx]
```
wk9874: Wouldnt it be easier to just do `random.shuffle(self.samples)`?

toktagger/api/core/query_strategy.py
```diff
@@ -38,38 +61,52 @@ class SequentialQueryStrategy(QueryStrategy):
...
+
+class UncertaintyQueryStrategy(RandomQueryStrategy):
+    """Uncertainty-based query strategy
+
+    Chooses the next sample based on uncertainty scores from existing annotations.
+    If no annotations exist, falls back to random sampling.
+    """
+
+    def __init__(self, samples, annotations, seed: int = 42):
+        super().__init__(samples, annotations)
```
wk9874: This will already shuffle the samples since it inherits from RandomQueryStrategy, making the following if statement redundant

toktagger/api/routers/samples.py
```diff
@@ -187,14 +188,53 @@ async def get_next_sample(
...
+        },
+        409: {"description": "Server is not setup to use the selected project."},
+    },
+)
+async def get_previous_sample(
+    request: Request,
+    project_id: str = Path(
+        description="The project to return the previous sample from."
+    ),
+    current_sample_id: Optional[str] = Query(
```
wk9874, line 262: This shouldnt be optional here as it is required for getting previous sample from query strategy

toktagger/api/routers/samples.py
```diff
@@ -187,14 +188,53 @@ async def get_next_sample(
...
+    ------------------------------------------------------------------------------
+    """
+    # Return the previous sample for human validation for this project
+    # Should use the query strategy, which access the database to determine the previous sample to annotate
+    # This should then be passed in to the /data endpoint to get required data for visualisation
+    # And the /annotation endpoint to get initial prediction (if available)
+    db_client = request.app.state.db_client
+    project = await utils.get_project(db_client, project_id)
+    samples = await utils.get_samples(db_client, project_id)
+    query_strategy = QUERY_STRATEGIES[project.query_strategy](samples)
```
wk9874: Since this is reinitialized each time, the 'previous' button may be unreliable:

* If new samples are added during a labelling sessions with Random query strategy, and then someone presses previous, the order of the shuffled samples will be different, and so next will send them to a random sample

* If new annotations are added, eg by a model finishing predictions, and someone presses next with Uncertainty QS, the order of samples by uncertainty may have changed due to new annotations, and so itll send them to a new pulse, not the previous one they were looking at

toktagger/api/routers/samples.py
```diff
@@ -187,14 +188,53 @@ async def get_next_sample(
...
 
-    data_pool = DataPool(
-        data_loader=DATA_LOADERS[project.data_loader](),
-        query_strategy=QUERY_STRATEGIES[project.query_strategy](samples, annotations),
-    )
     try:
-        sample = data_pool.query_strategy.get_next_sample()
+        sample = query_strategy.get_previous_sample(current_sample_id)
     except RuntimeError:
         raise HTTPException(status_code=204, detail="No more samples available!")
```
wk9874: Error message should be more like 'no previous sample available'

toktagger/api/schemas/annotations.py
```diff
@@ -9,6 +9,8 @@ class AnnotationIn(ConfiguredModel):
     created_by: AnnotatorTypes
     validated: bool = False
     uncertainty: Optional[float] = None
+    sample_id: Optional[str] = None
```
wk9874: Why did these move into AnnotationIn? The endpoint should fill these in rather then them be required in the actual Annotation's schema

toktagger/api/crud/utils.py
```diff
@@ -252,3 +253,37 @@ async def get_sample_summary(
...
+        return
+
+    sample_groups = defaultdict(list)
+    for annotation in annotations:
+        sample_groups[annotation.sample_id].append(annotation)
+
+    for sample_id, sample_annotations in sample_groups.items():
+        sample_obj_id = convert_to_objectid(sample_id, "samples")
+
+        if not await db_client.get_document_by_id("samples", sample_obj_id):
```
wk9874: Do we want this to fail on the first invalid sample ID? Or continue trying all others, then throw one error at the end with a list of any invalid sample IDs?

toktagger/api/routers/annotations.py
```diff
@@ -67,6 +67,28 @@ async def get_all_annotations(
...
+@router.put(
+    "/annotations",
+    responses={
+        200: {"description": "Annotations for this project updated successfully."},
+        404: {"description": "Project not found with that ID."},
+    },
+)
+async def import_annotations(
+    request: Request,
+    annotations: list[AnnotationTypes],
```
wk9874: Ahh now I see why you moved `sample_id` into `AnnotationIn` instead of keeping it in `Annotation` only - you need a sample ID attached to each annotation...

What I did instead in the models stuff was add a new schema:

```
class AnnotationBatchItem(ConfiguredModel):
    sample_id: str
    annotations: list[AnnotationTypes]
```

Your endpoint would then expect (I assume):
```
annotations: list[AnnotationBatchItem]
```

I personally think this is better since it still means you can define an Annotation without having to care about the project or sample ID, which is consistent with all our other schemas

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -34,36 +31,39 @@ import {
...
+import { NavigationBar } from "./nav";
 
 async function saveAnnotations(
   project_id: string,
   sample_id: string,
   annotations: Annotation[],
 ) {
+  // user has validated the annotations, so set created_by to "manual"
+  const updatedAnnotations = annotations.map((annotation: Annotation) => {
+    annotation.created_by = "manual";
```
wk9874: Do you also need to set validated = True?

toktagger/ui/src/app/core.tsx
```diff
@@ -102,3 +102,99 @@ export const deleteProject = async (project_id: string) => {
...
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify(annotations),
+  });
+  if (!response.ok) {
+    throw new Error(`Failed to save annotations: ${response.statusText}`);
+  }
+}
+
+export async function saveAnnotations(
```
wk9874, line 177: how does this work without being provided a sample ID to save to?

tests/api/core/test_query_strategy.py
```diff
@@ -86,13 +88,13 @@ def test_uncertainty_strategy(samples, annotations):
     # These correspond to samples number 4, 1, 2
     # It should then fallback to random selection of remaining samples - 3
     next_sample = strategy.get_next_sample()
-    assert next_sample.id == "sample_4"
-    next_sample = strategy.get_next_sample()
-    assert next_sample.id == "sample_1"
-    next_sample = strategy.get_next_sample()
     assert next_sample.id == "sample_2"
```
wk9874, line 128: Why have these reversed order? Indicates uncertainty query strat isnt working properly, should be giving us most uncertain first (annotation 5, linked to sample 4)

tests/api/core/test_query_strategy.py
```diff
@@ -54,27 +52,31 @@ def test_sequential_strategy(samples, annotations):
...
+        random_samples.append(next_sample)
+        next_sample_id = next_sample.id
 
-    # Check that two instances have not returned the same order of results
-    assert returned_samples_1 != returned_samples_2
+    assert len(random_samples) == len(samples)
+    assert random_samples != samples
 
 
 def test_uncertainty_strategy(samples, annotations):
```
wk9874, line 120: Should add 'preivous sample' versions of these tests for each query strategy

toktagger/api/routers/samples.py
```diff
@@ -187,14 +188,53 @@ async def get_next_sample(
...
+    Get the next sample to annotate for this project, according to query strategy.
+    ------------------------------------------------------------------------------
+    """
+    # Return the previous sample for human validation for this project
+    # Should use the query strategy, which access the database to determine the previous sample to annotate
+    # This should then be passed in to the /data endpoint to get required data for visualisation
+    # And the /annotation endpoint to get initial prediction (if available)
+    db_client = request.app.state.db_client
+    project = await utils.get_project(db_client, project_id)
+    samples = await utils.get_samples(db_client, project_id)
```
wk9874: By default this will return samples in the order in which they were created. Do we instead potentially want it to sort by shot ID?

Currently with the Local ELM demo project, when I press previous / next it unexpectedly goes in the 'wrong' order - instinctively I would expect it to go in shot ID order, but it does the reverse (because the samples were created highest shot ID first)

wk9874: Or split into two Query Strategies? One which is sequential by shot ID, one which is sequential by time created?

toktagger/ui/src/app/components/tools/export.tsx
```diff
@@ -0,0 +1,81 @@
...
+}: ExportToolInfo) {
+  const exportItems = ["All", "Current Sample"].map((item) => ({
+    id: item,
+    name: item,
+  }));
+  const [exportOption, setExportOption] = useState<string>("All");
+
+  const handleExport = () => {
+    if (exportOption === "All") {
+      exportAnnotations(project);
```
wk9874, line 36: The annotations returned differs between 'all' versus 'current:

With 'all':
```
  {
    "timestamp": "2025-11-21T12:07:03.399789",
    "label": "Sawtooth",
    "created_by": "manual",
    "validated": false,
    "uncertainty": 1,
    "sample_id": "692052ff8d2504a1af2843a5",
    "project_id": "692052ff8d2504a1af2843a3",
    "_id": "692055e78d2504a1af2843ae",
    "type": "time_region",
    "time_min": 0.4394439237225475,
    "time_max": 0.4589800451965979
  },
```

With current:
```
  {
    "created_by": "manual",
    "type": "time_region",
    "time_min": 0.4394439237225475,
    "time_max": 0.4589800451965979,
    "label": "Sawtooth"
  }
```

toktagger/ui/src/app/components/tools/export.tsx
```diff
@@ -0,0 +1,81 @@
...
+  project: Project;
+  sample: Sample;
+  current_annotations: Annotation[];
+};
+export function ExportTool({
+  project,
+  sample,
+  current_annotations,
+}: ExportToolInfo) {
+  const exportItems = ["All", "Current Sample"].map((item) => ({
```
wk9874, line 28: I'm not sure it makes sense to give the option of 'all' or 'current' on the sample screen? Shouldnt we have two places where you can export annotations:

* Sample screen, exports that one sample only
* Project screen (table with list of samples), exports all samples

?

(could be a later PR, we should probably add a toolbar to the projects screen to add things like this, along with eg the models training / predict stuff)

toktagger/ui/src/app/components/tools/import.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <>
+      <FileTrigger
+        onSelect={handleFileChange}
+        acceptedFileTypes={["application/json"]}
+      >
+        <Flex justifyContent="center" alignItems="center">
+          <Button variant="primary">
+            <Import />
+            <Text>Import Annotations</Text>
```
wk9874, line 57: Tried uploading the same form of data downloaded from export, eg:

```
[
  {
    "created_by": "manual",
    "type": "time_region",
    "time_min": 0.4394439237225475,
    "time_max": 0.4589800451965979,
    "label": "Sawtooth"
  }
]
```

And didnt seem to work:
```
INFO:     127.0.0.1:55906 - "PUT /projects/692052ff8d2504a1af2843a3/annotations HTTP/1.1" 404 Not Found
```

wk9874, line 57: Again may be good to split this out into two:

* Import annotations on sample screen - just a list of annotations for this sample, dont need to provide the sample ID
* Import annotations on project screen: Annotations for a range of samples, must provide sample IDs in some way

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -0,0 +1,194 @@
...
+  project_id: string;
+  sample_id: string;
+  annotations: Annotation[];
+};
+
+function NextButton({ project_id, sample_id, annotations }: ButtonInfo) {
+  const navigate = useNavigate();
+
+  const moveNextShot = useCallback(async () => {
+    await saveSampleAnnotations(project_id, sample_id, annotations);
```
wk9874: Im not convinced we should silently autosave when someone presses previous/next buttons without pressing save... what if someone is just flicking between different shots to see what the auto annotator tools look like? Theyll unknowingly be adding 'validated' annotations into the database and marking shots as validated

Maybe it could popup with 'save shots'? With a yes/no, and a 'set default / dont ask again' option?

toktagger/ui/src/app/components/plots/time-series.tsx
```diff
@@ -65,9 +83,15 @@ export const TimeSeries = ({
     setUpdateTools((current) => (current + 1) % 100);
   };
 
+  if (!disableToolingInteraction) {
```
wk9874: How do i enable zoom mode on the UI?

wk9874 (changes requested):

toktagger/api/core/annotators.py
samueljackson92: Label is the type e.g. `UFO`, `ELM`, `IRE` etc. We now have a default option of `Unknown`.

toktagger/api/core/query_strategy.py
samueljackson92: Random needs to be seeded so we keep the same (random) ordering with each call to the API. Otherwise `next` and `previous` will not work as expected.

toktagger/api/schemas/annotations.py
samueljackson92: What about round-tripping a import/export. If I export then my annotations file contains these properties, if I then try and import it again it will break if we don't allow this?

toktagger/api/crud/utils.py
samueljackson92: I think fail here. Something has gone wrong at this point.

toktagger/api/routers/annotations.py
samueljackson92: Ah ok. I think this makes sense.

toktagger/ui/src/app/core.tsx
samueljackson92: sample id's are in the file.

toktagger/api/routers/samples.py
samueljackson92: Keep it simple: sort by shot id.

toktagger/ui/src/app/components/tools/nav.tsx
samueljackson92: Improvement for another PR...

toktagger/api/routers/annotations.py
samueljackson92: I don't see how this will work when importing annotations for multiple samples?

toktagger/ui/src/app/components/plots/time-series.tsx
samueljackson92: Hold down "Alt" and scroll to scroll zoom. We should have a page showing keyboard shortcuts.... Another PR...

toktagger/api/routers/annotations.py
samueljackson92: I think you'd have to have a list of batched items and similarly for exporting. But I don't like this, thinking about how I want to use this *as a user*, I just want a flat list of annotations that I can load into a pandas dataframe...

With the approach I have implemented you still don't have to provide `sample_id` and `project_id` when creating annotations, but if you do have it then it can be optionally attached.

toktagger/ui/src/app/components/tools/nav.tsx
wk9874: Will leave this as a comment on #54

toktagger/ui/src/app/components/plots/time-series.tsx
wk9874: Raised #149

toktagger/api/core/query_strategy.py
```diff
@@ -11,29 +13,56 @@
...
-            raise RuntimeError("No more samples to label!")
+        index = self._get_matching_sample(current_sample_id)
+        previous_index = index - 1
+        previous_index = previous_index % len(self.samples)
 
-        index = random.randint(0, len(self.samples) - 1)
-        return self.samples.pop(index)
+        return self.samples[previous_index]
+
+    def _get_matching_sample(self, current_sample_id: Optional[str]) -> int:
```
wk9874: Picky comment: I think this method should be defined before `get_next_sample` and `get_previous_sample`, since those methods use this one

Also the typehint on `current_sample_id` is wrong here since it is a required param

toktagger/api/routers/samples.py
wk9874: Think this is still unresolved, but happy for it to just be an edge case we accept might cause issues since I can't think of a way to recitify this atm...

toktagger/api/core/query_strategy.py
```diff
@@ -42,40 +71,91 @@ class SequentialQueryStrategy(QueryStrategy):
...
+        self.annotations = annotations if annotations is not None else []
+        # simply shuffle the samples at the start
+        # seed is used to ensure consistent shuffling between calls
+        self._random_shuffle_samples(seed)
+
+    def _random_shuffle_samples(self, seed: int):
+        rng = np.random.default_rng(seed=seed)
+        self.samples = rng.permutation(self.samples)
+
+    def get_next_sample(self, current_sample_id: Optional[str] = None) -> Sample:
```
wk9874: This is redundant since it should be inherited from QueryStrategy (base method is identical to this as far as I can tell)

toktagger/api/core/query_strategy.py
```diff
@@ -42,40 +71,91 @@ class SequentialQueryStrategy(QueryStrategy):
...
+            return self.samples[0]
+
+        index = self._get_matching_sample(current_sample_id)
+        next_index = index + 1
+        next_index = next_index % len(self.samples)
 
-        return self.samples.pop(0)
+        return self.samples[next_index]
 
+    def get_previous_sample(self, current_sample_id: Optional[str] = None) -> Sample:
```
wk9874: Same here - should be inherited from QueryStrategy

toktagger/api/core/query_strategy.py
```diff
@@ -42,40 +71,91 @@ class SequentialQueryStrategy(QueryStrategy):
...
+
+    Chooses the next sample based on uncertainty scores from existing annotations.
+    If no annotations exist, falls back to random sampling.
+    """
 
-class UncertaintyQueryStrategy(QueryStrategy):
-    def get_next_sample(self) -> Sample:
-        if len(self.samples) == 0:
-            raise RuntimeError("No more samples to label!")
+    def __init__(self, samples, annotations, seed: int = 42):
```
wk9874: Should add type hints here

toktagger/api/core/query_strategy.py
```diff
@@ -42,40 +71,91 @@ class SequentialQueryStrategy(QueryStrategy):
...
-                    if sample.id == sample_id
-                ),
-                None,
+            sample_ids = [annotation.sample_id for annotation in self.annotations]
+            self.samples = sorted(
+                self.samples,
+                key=lambda sample: sample_ids.index(sample.id)
+                if sample.id in sample_ids
+                else -1,
+                reverse=True,
```
wk9874, line 124: I think we might have one too many `reverse=True`'s in here. `self.annotations` returns in reverse order already, so highest uncertainty first. If we then reverse it again here, wont we get them in uncertainty lowest -> highest order?

tests/api/core/test_query_strategy.py
wk9874: Yeah this test is wrong now, the correct order (highest uncertainty first) is sample 4, then 1, then 2, then fallback to random selection (so sample 3 as that's the only one left)

See comment above about an extra reverse() in the UncertaintyQueryStrategy for what might cause this

tests/api/routers/test_annotations.py
```diff
@@ -312,6 +387,7 @@ async def test_batch_update_annotations(api_client, setup_db, db_client):
         sort_by="time",
         sort_direction="ascending",
     )
+    print(annotations_sample_1)
```
wk9874: Remove

tests/api/routers/test_models.py
```diff
@@ -264,6 +264,7 @@ async def test_model_get_sample_prediction_wrong_sample(
     get_response = await api_client.get(
         f"/projects/{setup_model_db['project_id']}/samples/{setup_model_db['sample_ids'][-2]}/models/mock_disruption_cnn/predict/{task_id}"
     )
+    print(get_response.json())
```
wk9874: Remove

tests/api/routers/test_projects.py
```diff
@@ -37,13 +39,13 @@ async def test_get_all_projects_sortby(api_client, setup_db):
     response = await api_client.get("/projects?sort_by=task")
     # Should sort alphabetically by task
     # Sorts by case first (all uppers before any lowers)
-    # So ELM (project 1), then UFO (project 2), then disruption (project 0)
+    # So time series (project 1), then video (project 2), then time series (project 0)
```
wk9874: Update this comment to be correct

Also might want to either change the sort_by parameter, or change one of the projects to be a spectogram task? Because currently its arbitrary which of projects 0 or 1 it returns first since they have the same task

toktagger/api/core/data_loaders.py
```diff
@@ -38,7 +45,9 @@ def sample_data_type(cls) -> Type[ShotData | FileData | TimeSeriesFileData]:
 
     @abstractmethod
     def get_sample(
-        self, shot_id: int, sample_data: ShotData | FileData | TimeSeriesFileData
+        self,
+        sample: ShotData | FileData | TimeSeriesFileData,
```
wk9874: This typing is now wrong

toktagger/api/core/data_loaders.py
```diff
@@ -97,7 +106,14 @@ def sample_data_type(self) -> Type[FileData]:
         return FileData
 
     @pydantic.validate_call
-    def get_sample(self, shot_id: int, sample_data: FileData) -> ImageData:
+    def get_sample(self, sample: Sample, **kwargs) -> ImageData:
```
wk9874, line 109: Was this intentionally changed back to the old format? I split the arguments out into `shot_id` and `sample_data` so that pydantic could automatically validate the sample data model whenever this method was called

toktagger/api/core/data_loaders.py
```diff
@@ -97,7 +106,14 @@ def sample_data_type(self) -> Type[FileData]:
...
+        sample_data: FileData = sample.data
+
+        if not isinstance(sample_data, FileData):
```
wk9874: So eg splitting `sample_data` out in the arguments to the data loader makes this check redundant, as pydantic will automatically validate it on method call

toktagger/api/core/data_loaders.py
wk9874: (not essential anymore though, I think i initially did that when the loader registry relied on the type of `sample_data` to get the model schema, but realised that was a bad idea and added the `sample_data_type` classmethod instead)

toktagger/api/schemas/annotations.py
```diff
@@ -99,21 +97,22 @@ class ModelAnnotation(AnnotationIn):
...
-
-class AnnotationBatchItem(ConfiguredModel):
-    sample_id: str
-    annotations: list[AnnotationTypes]
-
+AnnotationBatchInputTypes = AnnotationOutTypes
 
 AnnotationTypeAdapter = TypeAdapter(AnnotationTypes)
 AnnotationOutTypeAdapter = TypeAdapter(AnnotationOutTypes)
+AnnotationBatchInputTypeAdapter = TypeAdapter(AnnotationBatchInputTypes)
```
wk9874: This is now redundant as it is identical to AnnotationOutTypeAdapter

toktagger/api/schemas/annotations.py
```diff
@@ -99,21 +97,22 @@ class ModelAnnotation(AnnotationIn):
...
-    Annotation, TimePointOut, TimeRegionOut, BoundingBoxOut, VideoBoundingBoxOut
+    TimePointOut, TimeRegionOut, BoundingBoxOut, VideoBoundingBoxOut, SpectrogramMaskOut
 ]
-
-
-class AnnotationBatchItem(ConfiguredModel):
-    sample_id: str
-    annotations: list[AnnotationTypes]
-
+AnnotationBatchInputTypes = AnnotationOutTypes
```
wk9874: Makes this redundant

toktagger/api/schemas/annotations.py
wk9874: Wouldn't it make more sense to have the export to be in form:

```
{
    sample_id: [annotations]
}
```

And then if you import them again via the UI itd use the batch update endpoint?

toktagger/api/routers/annotations.py
```diff
@@ -69,76 +68,58 @@ async def get_all_annotations(
...
-        200: {"description": "Annotations for this project deleted successfully."},
+        200: {"description": "Annotations for this project updated successfully."},
         404: {"description": "Project not found with that ID."},
+        422: {"description": "Invalid annotation data provided."},
     },
 )
-async def delete_all_annotations(
+async def import_annotations(
     request: Request,
+    annotations: list[AnnotationBatchInputTypes],
```
wk9874: Currently `import_annotations` and `update_annotations` in the annotations router page both expect an `AnnotationBatchInputTypes` list, which (might? since they're optional) contain the `sample_id` and `project_id` within the Annotation.

`import_annotations` needs to contain the sample ID in some way alongside the annotations, which was why I previous had:

```
class AnnotationBatchItem(ConfiguredModel):
    sample_id: str
    annotations: list[AnnotationTypes]
```
and `import_annotations` expects a list of those. It should not need `project_id` to be provided in the annotation itself

toktagger/api/routers/annotations.py
```diff
@@ -200,13 +181,13 @@ async def get_annotations(
...
     responses={
-        200: {"description": "Annotations for this sample retrieved successfully."},
+        200: {"description": "Annotations for this sample updated successfully."},
         404: {"description": "Project or Sample not found with that ID."},
     },
 )
 async def update_annotations(
     request: Request,
-    annotations: list[AnnotationTypes],
+    annotations: list[AnnotationBatchInputTypes],
```
wk9874: Meanwhile `update_annotations` shouldn't have a project_id or sample_id provided within the annotation. What if the values provided in the annotation arent the same as those in the path of the API request? The user might see unexpected behaviour here


Generally, this is why I'm going back to `AnnotationIn` not having `sample_id` or `project_id` as optional parameters. It means that we know exactly what format and what parameters are expected by anything which accepts an `AnnotationType`. Currently we have the confusing situation where two different endpoints accept the same type, but get the sample IDs from different places

toktagger/api/schemas/annotations.py
```diff
@@ -70,26 +69,25 @@ class VideoBoundingBox(BoundingBox):
...
 
 
-class AnnotationOut(BaseModel):
-    id: str = Field(..., alias="_id")
-    created_by: AnnotatorTypes
-    project_id: str
-    sample_id: str
+class AnnotationOut(AnnotationIn):
+    id: Optional[str] = Field(None, alias="_id")
+    project_id: Optional[str] = None
```
wk9874: Why are these optional now?

toktagger/ui/src/app/projects/project_id/page.tsx
```diff
@@ -175,15 +197,21 @@ export default function ProjectView() {
...
-                label="Search By Shot ID"
-                // SearchField should be able to do validation when provided a 'pattern' inside a Form element
-                // But I could not for the life of me get that to work, so will do it manually...
-                onSubmit={onSearchSubmit}
-                validationState={errorMessage ? "invalid" : undefined}
-                errorMessage={errorMessage}
-              />
+              <Flex gap="size-100" alignItems="center" justifyContent="start">
+                <AddSamplesEditor project={project} onModify={refreshSamples} />
+                <SearchField
```
wk9874, line 208: In terms of page layout, the search bar on the Projects and Samples page should be in the same location:

Image

Image

scripts/setup.py
```diff
@@ -4,12 +4,21 @@
...
+    query_strategy: str,
+    min_time_step: float = 0.0001,
+) -> str:
     project = {
         "name": name,
         "task": task,
         "query_strategy": query_strategy,
         "data_loader": data_loader,
+        "time_min": -0.1,
+        "time_max": 0.8,
```
wk9874: Need to expose these time settings as options to `create_project`, and then add `create_project(..., time_max=times[-1])` into the `create_mock_data.py` script

toktagger/api/routers/samples.py
```diff
@@ -170,77 +157,84 @@ async def add_samples(
...
-    )
-    annotations = await utils.get_annotations(
-        db_client, project_id, validated=False, sort_by="uncertainty"
-    )
-    logger.debug(f"found {len(samples)} non validated samples")
-    logger.debug(f"found {len(annotations)} non validated annotations")
-
-    query_strategy = QUERY_STRATEGIES[project.query_strategy](samples, annotations)
+    samples = await utils.get_samples(db_client, project_id)
+    query_strategy = QUERY_STRATEGIES[project.query_strategy](samples)
```
wk9874: This never passes in annotations to the Query Strategy

toktagger/api/routers/samples.py
```diff
@@ -170,77 +157,84 @@ async def add_samples(
...
     """
+    # Return the next sample for human validation for this project
+    # Should use the query strategy, which access the database to determine the next sample to annotate
+    # This should then be passed in to the /data endpoint to get required data for visualisation
+    # And the /annotation endpoint to get initial prediction (if available)
     db_client = request.app.state.db_client
-    await utils.get_project(db_client, project_id)
+    project = await utils.get_project(db_client, project_id)
+    samples = await utils.get_samples(db_client, project_id)
+    query_strategy = QUERY_STRATEGIES[project.query_strategy](samples)
```
wk9874: This never passes annotations into the query strategy, so Uncertainty based QS wont work

toktagger/api/core/query_strategy.py
wk9874: Also `annotations` should be optional, like all other query strategies

toktagger/ui/src/app/components/tools/export.tsx
```diff
@@ -0,0 +1,81 @@
...
+        `${project.name}_${sample.shot_id}_annotations.json`,
+      );
+    }
+  };
+
+  return (
+    <div className="m-4">
+      <Flex direction="column">
+        <Flex direction="row" alignItems="center" gap="size-100">
+          <ComboBox
```
wk9874, line 46: Personally I dont even think we should provide the option here... If you want to export all annotations for a project, do it from the project screen. If you want just a specific sample, do it from the sample screen

toktagger/ui/src/app/time_series/components/time-series.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { Zones } from "@/app/components/tools/zones";
+import "react-contexify/ReactContexify.css";
+
+import {
+  createAnnotationToDisplayAnnotationFunc,
+  updateAnnotations,
+} from "@/app/utils";
+import { useEffect, useMemo, useState } from "react";
+
+const zoneCategories: Category[] = [
```
wk9874, line 34: Where have eg `RampUp`, `FlatTop` etc gone?

toktagger/ui/src/app/time_series/components/time_series.tsx
```diff
@@ -0,0 +1,178 @@
...
+    <div className="flex space-y-3">
+      <div className="flex-1 text-center items-center">
+        <ContextMenuProvider menuId="elm-menu">
+          <ZoneProvider
+            categories={zoneCategories}
+            initialData={zones}
+            onModifyZone={updateZones}
+          >
+            <TimeSeries
+              plotId="ELMs"
```
wk9874: Should this still reference ELMs?

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -63,67 +60,73 @@ type SampleViewInfo = {
...
-  if (project.task == "disruption") {
-    const result = MultiVariateTimeSeriesDataSchema.safeParse(data);
-    if (!result.success) {
-      throw new Error("Invalid data for disruption view");
+  const [result, setResult] = useState<
+    MultiVariateTimeSeriesData | CompositeData | SpectrogramData | null
+  >(null);
+
+  useEffect(() => {
+    if (project.task == "disruption") {
```
wk9874: This 'disruption' task no longer exists

wk9874: In fact none of these project tasks exist...

toktagger/ui/src/app/components/providers/vpsan-provider.tsx
```diff
@@ -69,8 +69,9 @@ export const VSpanProvider = ({
     onModifyVSpan?.(spans.current);
   };
 
-  const handleDelete = (input: unknown) => {
+  const handleVSpanDelete = (input: unknown) => {
```
wk9874: What is input here?

toktagger/ui/src/app/components/providers/vpsan-provider.tsx
```diff
@@ -69,8 +69,9 @@ export const VSpanProvider = ({
...
     spans.current = spans.current.filter((span) => span !== input);
+    spans.current = spans.current.filter((span) => !span.selected);
```
wk9874, line 74: Do you intend to do two updates to this here?

toktagger/ui/src/app/components/providers/vpsan-provider.tsx
```diff
@@ -193,6 +199,17 @@ export const VSpanProvider = ({
     );
   });
 
+  // Delete selected zones on Delete/Backspace keypress
+  document.addEventListener("keydown", (e) => {
+    if (e.key === "Delete" || e.key == "Backspace") {
```
wk9874: How do i select a zone for this to work? For me either clicking, or clicking and holding, and then pressing delete or backspace does nothing...

toktagger/ui/src/app/components/providers/zone-provider.tsx
```diff
@@ -209,6 +215,19 @@ export const ZoneProvider = ({
     );
   });
 
+  // Handle keyboard events
+
+  // Delete selected zones on Delete/Backspace keypress
+  document.addEventListener("keydown", (e) => {
+    if (e.key === "Delete" || e.key == "Backspace") {
```
wk9874: Same here, how do i use this?

toktagger/ui/src/app/components/providers/annotation-provider.tsx
```diff
@@ -73,7 +73,7 @@ export const ContextMenuProvider = ({
   // Set up listeners to toggle tooling interaction
   useEffect(() => {
     const disableInteraction = (event: KeyboardEvent) => {
-      if (event.key === "Shift" || event.key === "Control") {
+      if (event.key === "Alt") {
```
wk9874, line 76: its not clear to me that Alt is doing anything?

toktagger/ui/src/app/components/tools/export.tsx
wk9874: Ok now theyre consistent, but project_id and sample_id are filled with nulls when exporting only the current sample:
```
[
  {
    "created_by": "manual",
    "type": "time_region",
    "project_id": null,
    "sample_id": null,
    "validated": false,
    "uncertainty": 1,
    "time_min": 5.311458196992046,
    "time_max": 5.902732015804313,
    "label": "ELM"
  },
  {
    "project_id": null,
    "sample_id": null,
    "validated": false,
    "uncertainty": 1,
    "created_by": "disruption_cnn",
    "type": "time_point",
    "time": 7.837500307692308,
    "label": "Disruption"
  },
  {
    "project_id": null,
    "sample_id": null,
    "validated": false,
    "uncertainty": 1,
    "created_by": "manual",
    "type": "time_point",
    "time": 9.026386117988075,
    "label": "Disruption"
  }
]
```

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -0,0 +1,312 @@
...
+} from "@/app/core";
+import { useNavigate } from "react-router-dom";
+import { useSample } from "@/app/contexts/SampleContext";
+
+const TOAST_TIMEOUT = 5000;
+
+async function getNextSample(project_id: string, current_sample_id: string) {
+  const NEXT_URL = `${BACKEND_API_URL}/projects/${project_id}/samples/next?current_sample_id=${current_sample_id}`;
+  const sampleResult = await fetch(NEXT_URL);
+  if (sampleResult.status === 204) {
```
wk9874, line 33: Should catch other errors and display to user

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -0,0 +1,312 @@
...
+  return sample;
+}
+
+async function getPreviousSample(
+  project_id: string,
+  current_sample_id: string,
+) {
+  const PREVIOUS_URL = `${BACKEND_API_URL}/projects/${project_id}/samples/previous?current_sample_id=${current_sample_id}`;
+  const sampleResult = await fetch(PREVIOUS_URL);
+  if (sampleResult.status === 204) {
```
wk9874, line 50: Same here, should catch other errors too and display error to user

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -0,0 +1,312 @@
...
+      }
+    }
+
+    window.addEventListener("keydown", handleKeyDown);
+    return () => window.removeEventListener("keydown", handleKeyDown);
+  }, [annotations, sample_id, project_id, navigate, moveNextShot]);
+
+  return (
+    <View marginStart="size-100">
+      <ActionButton onPress={moveNextShot}>
```
wk9874: Should add an appropriate aria-label to this so that it can be identified by screen readers (and Playwright for end to end tests!)

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -0,0 +1,312 @@
...
+      }
+    }
+
+    window.addEventListener("keydown", handleKeyDown);
+    return () => window.removeEventListener("keydown", handleKeyDown);
+  }, [annotations, sample_id, project_id, navigate, movePreviousShot]);
+
+  return (
+    <View marginStart="size-100">
+      <ActionButton onPress={movePreviousShot}>
```
wk9874: Again, aria-label here

wk9874: (Note that for most elements React / React Spectrum handle this for us, and the text label automatically becomes the aria label. But anywhere that there are buttons with an icon only (no text), we should add a text aria-label to describe what the button is)

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -0,0 +1,312 @@
...
+          validateOnNavigate={validateOnNavigate}
+        />
+        <ClearButton setAnnotations={setAnnotations} />
+      </ButtonGroup>
+      <TooltipTrigger delay={1000} placement="bottom">
+        <Checkbox
+          isSelected={validateOnNavigate}
+          onChange={setValidateOnNavigate}
+        >
+          Validate on Navigate
```
wk9874: Should this just change to 'Save on Navigate'? 'Validate' is potentially a confusing term for users

toktagger/ui/src/app/components/tools/vspans.tsx
```diff
@@ -116,6 +124,20 @@ export const VSpans = ({ plotId, plotReady, forceUpdate }: ToolingProps) => {
...
       for (const vspan of vspans) {
+        let opacity = 0.5;
+
+        // change opacity if the zone is selected
+        if (
+          selectedXRange &&
+          vspan.x > selectedXRange[0] &&
+          vspan.x < selectedXRange[1]
+        ) {
+          vspan.selected = true;
```
wk9874, line 135: Cant test this as I dont know how to select a Vspan or Zone... (I assume clicking on it? If so, this isnt working)

toktagger/ui/src/app/time_series/components/time-series.tsx
```diff
@@ -0,0 +1,218 @@
...
+  createAnnotationToDisplayAnnotationFunc,
+  updateAnnotations,
+} from "@/app/utils";
+import { useEffect, useMemo, useState } from "react";
+import { VSpanProvider } from "@/app/components/providers/vpsan-provider";
+import { VSpans } from "@/app/components/tools/vspans";
+import { useSample } from "@/app/contexts/SampleContext";
+import { Flex, View } from "@adobe/react-spectrum";
+
+const zoneCategories: Category[] = [
```
wk9874, line 34: disruption things like 'RampUp', 'FlatTop' missing

toktagger/ui/src/app/time_series/components/time-series.tsx
```diff
@@ -0,0 +1,218 @@
...
+    const domainHeight = 1 / numRows;
+    // Dynamically generate y-axis domains based on numRows
+    const yAxisDomains = Array.from({ length: numRows }, (_, i) => {
+      const start = i * domainHeight;
+      const end = (i + 1) * domainHeight;
+      return [start, end];
+    });
+
+    // Build yaxis layout object dynamically
+    const yAxesLayout = yAxisDomains.reduce(
```
wk9874, line 156: Where are the axes titles set? I dont see a title when there is only one plot (for multiple plots the key is displayed which acts as the title)

Image

toktagger/ui/src/app/core.tsx
```diff
@@ -103,6 +109,121 @@ export const deleteProject = async (project_id: string) => {
...
+  }
+  const annotations = await response.json();
+  return annotations;
+}
+
+export async function saveSampleAnnotations(
+  project_id: string,
+  sample_id: string,
+  annotations: Annotation[],
+  validateOnSave: boolean = true,
```
wk9874: This should probably be called validateOnNavigate

toktagger/ui/src/app/core.tsx
```diff
@@ -103,6 +109,121 @@ export const deleteProject = async (project_id: string) => {
...
+) {
+  // user has validated the annotations, so set created_by to "manual"
+  const updatedAnnotations = annotations.map((annotation: Annotation) => {
+    annotation.created_by = "manual";
+    annotation.validated = validateOnSave ? true : annotation.validated;
+    return annotation;
+  });
+
+  const ANNOTATIONS_URL = `${BACKEND_API_URL}/projects/${project_id}/samples/${sample_id}/annotations`;
```
wk9874, line 164: I wouldve thought we just wouldnt save if validateOnNavigate=False?

toktagger/ui/src/app/time_series/components/time-series.tsx
```diff
@@ -0,0 +1,218 @@
...
+              <TimeSeries
+                plotId="TimesSeriesView"
+                plotConfig={{ data: plotData, layout: plotLayout }}
+              >
+                <Zones onUpdate={updateZones} />
+                <VSpans onUpdate={updateVSpans} />
+              </TimeSeries>
+            </VSpanProvider>
+          </ZoneProvider>
+        </ContextMenuProvider>
```
wk9874, line 233: This means we've lost the little table which used to appear at the bottom of the disruption view, right?

I'm wondering whether we want to still have the old tasks which exist on the backend, so that we can display different things under the banner of 'time-series'.

Eg, if task == 'disruption', dont add some of those tools to the toolbar as they wouldnt be useful, and render the table like before. Also limit which Vspans and Zones are available.

We could either on the backend or the frontend store which type of data it is (ie disruption = TimeSeries, ELM = TimeSeries, UFO = Image, etc)

wk9874 (changes requested):

toktagger/api/routers/samples.py
samueljackson92: I think it's a little better because you can at least cycle through every sample. However, I think it's still an issue. I think we should fix this by having the concept of a "validated" or "visited" sample and properly tracking state. That's a bigger task than we can fit into this PR, so I think it's a job for another issue.

I created: https://github.com/ukaea/toktagger/issues/153 for this

toktagger/api/schemas/annotations.py
samueljackson92: I really don't like this I would much rather have a flat list of annotations which can be easily imported into a pandas dataframe or similar

toktagger/ui/src/app/components/providers/annotation-provider.tsx
samueljackson92: Have you tried zooming while hovering over an annotation?

toktagger/ui/src/app/components/tools/import.tsx
samueljackson92: I think it's nice to have the option of either all or just this sample on the sample page?

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
samueljackson92: Yup gone. Diff was out of date...

toktagger/ui/src/app/time_series/components/time-series.tsx
samueljackson92: Maybe, I think for now keeping it generic is ok for now. We can bring more specific things later if required.

I also only removed the table because I think it needs more work, but actually I can quickly provide some updates here to improve it and re-add it. Here's how it looks now:

image

toktagger/ui/src/app/components/tools/import.tsx
wk9874: I think for this, I would have the 'Import Annotations' button on the sample page not require sample / project IDs in the schema which is uploaded. You could have it still accept both I guess, but:


1. If the annotations uploaded by the user dont contain a project ID and/or sample ID, automatically fill it with the proejct and sample currently being looked at
2. If the annotations do contain a project or sample ID, these take priority over the page the user is currently looking at (?)

toktagger/ui/src/app/components/tools/vspans.tsx
wk9874: Used box select and this works

tests/api/routers/test_annotations.py
wk9874: Remove

tests/api/routers/test_models.py
wk9874: Remove

tests/api/routers/test_projects.py
wk9874: Update this comment to be correct

Also might want to either change the sort_by parameter, or change one of the projects to be a spectogram task? Because currently its arbitrary which of projects 0 or 1 it returns first since they have the same task

toktagger/api/core/data_loaders.py
wk9874: This typing is now wrong

toktagger/api/core/data_loaders.py
wk9874: ^

toktagger/api/schemas/annotations.py
wk9874: Why are these optional now?

toktagger/ui/src/app/projects/project_id/page.tsx
wk9874: In terms of page layout, the search bar on the Projects and Samples page should be in the same location:

Image

Image

toktagger/ui/src/app/components/providers/vpsan-provider.tsx
wk9874: What is input here?

toktagger/ui/src/app/components/providers/vpsan-provider.tsx
wk9874: Do you intend to do two updates to this here?

toktagger/ui/src/app/components/tools/nav.tsx
wk9874: I think more generally - what do we want this to do? Currently it will always save (And hence overwrite any preexisting) annotations when you navigate, its just they wont be marked as validated. I think personally if this box is not ticked, it shouldnt save the annotations on navigation at all... But I leave it up to you :)

toktagger/ui/src/app/projects/project_id/page.tsx
```diff
@@ -175,15 +197,21 @@ export default function ProjectView() {
...
+                  label="Search By Shot ID"
+                  // SearchField should be able to do validation when provided a 'pattern' inside a Form element
+                  // But I could not for the life of me get that to work, so will do it manually...
+                  onSubmit={onSearchSubmit}
+                  validationState={errorMessage ? "invalid" : undefined}
+                  errorMessage={errorMessage}
+                />
+              </Flex>
+              <Flex gap="size-100">
+                <ImportButton project_id={project_id} />
```
wk9874: When importing annotations, I suspect from the user's POV it would be better if they could provide the annotations without the sample IDs, and instead just provide the shot IDs. The UI could then do the mapping between the two somehow?

But this could be raised as an issue and looked at in a future PR

toktagger/ui/src/app/components/plots/time-series.tsx
```diff
@@ -244,6 +244,12 @@ export const TimeSeries = ({
       });
     };
 
+    // if (!disableToolingInteraction) {
```
wk9874: Is this intentionally commented out?

toktagger/ui/src/app/components/tools/import.tsx
wk9874: Think this hasn't been addressed - if I upload an annotation via the 'Import Annotations' button on the sample page (not the samples table) like this:

```
[
  {
    "label": "Disruption",
    "created_by": "manual",
    "validated": true,
    "uncertainty": 0,
    "type": "time_point",
    "time": 8.4
  }
]
```
I think it should fill the shot ID, sample ID, project ID automatically for me based on the sample I am currently looking at

toktagger/api/routers/annotations.py
```diff
@@ -208,10 +208,15 @@ async def update_annotations(
...
     await utils.get_project(db_client=db_client, project_id=project_id)
-    await utils.get_sample(
+    sample = await utils.get_sample(
         db_client=db_client, project_id=project_id, sample_id=sample_id
     )
 
+    # Set shot_id for each annotation
+    for annotation in annotations:
+        annotation.shot_id = sample.shot_id
+        print(annotation.shot_id)
```
wk9874: Remove debug print statement

toktagger/ui/main.jsx
```diff
@@ -2,6 +2,23 @@ import React from "react";
...
+function updateFavicon() {
+  const isDark = window.matchMedia("(prefers-color-scheme: dark)").matches;
+  const favicon = document.querySelector('link[rel="icon"]');
+  if (favicon) {
+    favicon.href = isDark ? "/favicon_light.png" : "/favicon.png";
+  }
+}
+
+// Update on load
+updateFavicon();
```
wk9874: I still see the old favicon when using `toktagger` to run the simulation - but this might be browser caching...

wk9874: Finallyyyy fixed

samueljackson92:
[celebrate]     Jackson, Samuel (Computing) reacted to your message:
________________________________

PR #130, opened 2025-12-01
Backend changes for Image DataLoader

wk9874:
Includes backend changes to make image data loader work for Abdullah's annotation UI. Includes:

* Changing format of schemas to match COCO
* Adding DataParams (similar to ViewParams but for the data loader) which contain the frame being asked for
* Propagating this change through to endpoints in the UI, including /data and /annotators endpoints
* Change image data to be sent in base64 encoded string to increase performance
* Add a basic UI page for UFOs to check that images are served correctly (to be replaced by Abdullah's UI)
* Add tests


samueljackson92 (approved):
Looks good to me. Make sure we remove the UFOs page when Abdullah's stuff comes in.

Issue #32, opened 2025-06-12
Video Annotation


jblake42:
Tooling to annotate video should be implemented. Some tools do already exist that we could use:

- https://www.npmjs.com/package/react-video-annotation-tool
- https://github.com/HumanSignal/label-studio-frontend/tree/master/examples

abdullah-ukaea:
**Update on video annotation**

Instead of a single video for example in .mp4 format we are now looking at a series of individual frames. The UI would have previous/next buttons to cycle through frames of a video.

Example tooling could be:
- https://annotorious.dev/

jblake42:
[Here ](https://www.figma.com/proto/aJJ0cCPI7ETgGHwarRCsWh/MAST-Data-Visulisation?page-id=1%3A2&node-id=21-152&viewport=407%2C-228%2C0.35&t=7AQ22FIhtAKDnQ2b-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=21%3A152)is an initial idea of what we could do for things like the 'scrubbing bar' - we'll discuss more in the weekly catchup

PR #137, opened 2025-12-17
Frame Annotation Tooling

abdullah-ukaea:
This PR introduces frame-based annotation for UFO image data and integrates it with existing UI tooling and the backend annotation APIs.

We currently support rectangle-only annotations (polygons to follow in a later PR) with per-instance tracking using track_id.

Addresses issue https://github.com/ukaea/toktagger/issues/32 

# What this adds

**Frame-level annotation view for UFO tasks**
* Adds a new UFOView that wraps FrameView for per-frame annotation.
* Renders frames using base64-encoded image data from the backend.
* Frame navigation via Prev / Next / Jump wired through the existing dataParams.frame plumbing.

**Rectangle-only Annotorious integration with tracking**
* Uses @annotorious/react to draw rectangles only.
* Each annotation is stamped with class label + track_id to enable instance-style tracking.
* Drawing is enabled only when a selected instance profile (class + track) exists in the UFO toolbar.

**Per-frame localStorage persistence + aggregation**
* Per-frame annotations are stored as W3C image annotations in localStorage (keyed by app://p/<project>/s/<sample>/f/<frame>).
* A background auto-save loop flushes edits from Annotorious into localStorage.
* window.ufoCollectForSave() sweeps all frames for a sample and returns a flat W3C list for saving.

Toolbar integration (class + instance profiles)
* Left-hand UFO toolbar provides:
     - Class picker (hard-coded label map for now)
     - Instance panel listing (class_name, track_id) profiles
     - Cross-frame count badges per instance
* Toolbar + FrameView stay in sync via window.* globals + ufo:state events.
* Cross-frame instance counts are computed with a chunked localStorage scanner (scanInstanceCountsChunked).

**Backend save + navigation wiring**
* UFO Save converts: W3C → COCO frames → VideoBoundingBox[] and PUTs to: PUT /projects/{project_id}/samples/{sample_id}/annotations (API on :8002)
* Important: backend expects created_by in AnnotatorTypes; for UI annotations we send created_by: "manual".
* Next sample: saves current UFO payload then loads next sample via: GET /projects/{project_id}/samples/next
* Jump to shot: saves multi-frame annotations before navigating to the target sample from shot_id.

**Bulk delete & clear flows**
* Right-click an instance → delete that instance’s annotations across all frames (with counts preview).
* Delete All Instances → wipes all instance profiles and annotations for the sample across every frame.
* Toolbar buttons delegate to frame-level helpers:
   - ufoClearAllFrames() — removes all frame keys for the sample from localStorage
   - ufoClearCurrent() — clears current frame overlay + storage

# Implementation notes

* Annotorious ImageAnnotator is used as the drawing surface; tool is restricted to rectangles.
* An imperative AnnoBridge wraps useAnnotator() and:
   - Normalizes rectangle annotations via normalizeWithMode, stamping class/track metadata from toolbar selection.
   - Tracks dirty state and exposes: persistWorkingNow, hydrateOverlay, clearOverlaySilently, hasUnsaved, markSaved.
   - Listens to create/update/delete events and re-normalizes overlay.
   - Supports “auto quick add”: if a second shape for the same instance is drawn in a frame, a new profile can be created and the duplicate is retagged.

# Files added / modified

* toktagger/ui/src/app/frames/components/adapters.ts
* toktagger/ui/src/app/frames/components/bridge.tsx
* toktagger/ui/src/app/frames/components/frames.tsx
* toktagger/ui/src/app/frames/components/lib.ts
* toktagger/ui/src/app/frames/components/ui.tsx
* toktagger/ui/src/app/components/tools/toolbar.tsx
* toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx

# Future Work

* Add polygon tool support (alongside rectangles)
* Pan/zoom/drag UX for the image view
* Timeline/transport controls for faster navigation through frames
* Deeper backend integration (e.g., loading existing stored frame annotations into localStorage/hydration path
* Improve jump-to-frame UX (direct frame number input + enter)


abdullah-ukaea:
Currently working on passing the linters (I was very liberal with the any type and I can do better with some usestates and useeffects so minor refactoring will be done), then after that this PR will be review ready, perhaps while I work on this @samueljackson92  @wk9874  you guys can check the functionality of the frame annotation page and confirm everything works. 
image


image

wk9874:
Will leave some comments for things I find as I play around with it!

First thing: If I change the class selected in the dropdown on the left before drawing any box, itll still create an instance of it in the list below. Do we want this? I would suspect that it should only create the instance in the list when a box is drawn



**Fixed!**

wk9874:
If I hit 'Clear All', the instances still exist in the column on the left hand side, even though all the UFOs within them have been deleted. Can we also delete the instances?

**Fixed!**

abdullah-ukaea:
> If I hit 'Clear All', the instances still exist in the column on the left hand side, even though all the UFOs within them have been deleted. Can we also delete the instances?

So currently I have 3 options to clear: Clear Current, Clear ALL, Delete ALL Instances

clear current deletes any annotations on the frame, but does not touch any of the instance profiles

clear all deletes any annotations on all frames, but does not touch any of the instances profiles

delete all instances, clears all annotations and deletes all instance profiles.

You can also delete annotations by using the popup directly on the annotation when clicked. I do have a check here to delete the instance profile if deleting this annotation means there are no annotations in this instance

Would you still like Clear ALL to also delete the empty instances?

wk9874:
If I delete a UFO box from a previous frame, then go to the next frame and draw a new box, it adds it to the same instance which I 'deleted' the box from previously. I think this should create a new instance instead? See video below for what I mean:




We probably should also support the functionality where eg I want to add a UFO in the same instance to non continuous frames (Eg if it goes off the screen and comes back, but youre fairly sure its the same one). In that case I would expect youd need to click the instance in the list first, before drawing a new box. Just drawing a new box directly onto the canvas should create a new instance

wk9874:
> > If I hit 'Clear All', the instances still exist in the column on the left hand side, even though all the UFOs within them have been deleted. Can we also delete the instances?
> 
> So currently I have 3 options to clear: Clear Current, Clear ALL, Delete ALL Instances
> 
> clear current deletes any annotations on the frame, but does not touch any of the instance profiles
> 
> clear all deletes any annotations on all frames, but does not touch any of the instances profiles
> 
> delete all instances, clears all annotations and deletes all instance profiles.
> 
> You can also delete annotations by using the popup directly on the annotation when clicked. I do have a check here to delete the instance profile if deleting this annotation means there are no annotations in this instance
> 
> Would you still like Clear ALL to also delete the empty instances?

Ah, didnt see 'delete all instances'! I can't think of a reason though why you would want to clear all annotations, but leave the instances intact. I see the logic for clear current - remove annotations from the current frame, but since annotations may exist on previous frames for instances, they have to be left intact. But for clear all, there's basically no value keeping the instances around. 

So personally I would get rid of either the clear all / delete all instances button, and condense the functionality into one (which deletes all annotations and all instances). 

But happy to be convinced the other way if you can think of a case where you might wanna delete all boxes without deleting the instances attached?

-> Probably just remove the 'Clear ALL' button, centre the 'Clear Current' button (and rename 'Clear Current Frame'), and then the delete all instances button can stay where it is since that does exactly what I want already! :)

wk9874:
Final comment for now - its sometimes quite hard to click on the boxes you've drawn to move / resize / see the details about them. I wonder if we can make it so if you click anywhere on the box it selects it, instead of what currently seems to happen where you can only click on the lines at the edge... Dunno if this is a limitation of Annotorious tho, in which case dont worry about it!

That would I suppose prevent you from drawing a box within a box, but i dont know if that would ever be useful or not. I guess if you had a big cloud of UFOs you may want to draw a box around them, and then draw around each individual UFO within that box... hmmm

Maybe a solution would be that if you click on the instance in the list on the left hand side, it selects that object on the image (if it exists on the frame youre currently looking at). But if that'd be a significant amount of work dont worry about it for this PR, we can raise an issue and come back to it later :)

wk9874:
Just tried it out with loading data live from JET and works perfectly!

Only tiny comment - it defaults to getting frame 292 when there are no annotations - assume this is a hangover from the set of 10 test images I gave you. I actually dont know what frame number to start with... maybe 200? Thats normally near the beginning of the pulse, but not black...

**Fixed!**

wk9874:
Feature request: Would be great if you could navigate previous / next frame with the arrow keys, and delete a selected UFO box with the Delete key!

abdullah-ukaea:
I now pass eslinting and prettier + no merge conflicts. I think a pytest is failing ( weird as it passed in a commit before) but it might not be related to this PR? 

image

Now I will work on @wk9874  Feature requests

Please do not review just yet as some features will be added shortly.

abdullah-ukaea:
@wk9874  When you get a chance, please pull the latest changes and let me know if the new updates meet the features / functionalities you requested above. (annotorious annotations click ability and arrow key shortcuts will be done outside of this PR) If so, we can start the code review process I believe.

wk9874:
Have marked the bugs above which are now fixed, the two outstanding comments above are feature requests so can go into Issues and be addressed later :)

Next thing I've spotted - good news is that the UI is clearly storing and recovering annotations correctly, as if I add some boxes, save, and refresh the page, the boxes which I previously drew are placed back on the frames!

Issue is that the list on the left hand side doesn't propagate these correctly, so it looks as if they're missing. Can you make it so that previous annotations are added back into the instances list on the left?

wk9874:
It'd also be good to have a 'propagate boxes' toggle - If you had a toggle that just meant the boxes dont propagate over frames, then users could still do 'tracking' if they wanted by clicking on the instance in the list and drawing a new box on the next frame... it would just be quicker to use if you dont have many consecutive UFO events rather than having to right click -> delete each box. 

Keep it in 'tracking' mode, ie keep it storing instance labels and stuff, just make the toggle enable / disable the boxes moving across frames automatically

docker-compose.dev.yml
```diff
@@ -42,6 +42,7 @@ services:
       - mongo
     volumes:
       - ./data/:/app/data
+      - ./data/:/data
```
jblake42: Is this definitely needed? It seems weird we are mapping this host volume twice for this service. The `cwd` of the service is `/app` and looking at the dataloader this is where it looks for locally stored files/directories. I could be wrong though, I've only glanced over the backend code

toktagger/api/core/data_loaders.py
```diff
@@ -105,8 +105,10 @@ def get_sample(self, shot_id: int, sample_data: FileData) -> ImageData:
                 f"Could not find directory at '{dir_path}', relative to {pathlib.Path().cwd()} - {list(pathlib.Path().cwd().iterdir())}"
             )
         # Open image which represents frame selected
-        if self.params.name != "image":  # TODO do we want this?
-            file_path = next(dir_path.iterdir())
+        if self.params.name != "image":
+            raise ValueError("Must provide image data parameters!")
+        elif self.params.frame is None:
+            file_path = sorted(dir_path.iterdir())[0]
```
jblake42: Might be an edge case but this looks unsafe to me - do we risk an `IndexError` exception here if an empty directory is provided?

toktagger/ui/src/app/components/tools/toolbar.tsx
jblake42, line 1: This toolbar file is getting pretty nasty with lots of conditional logic etc. I think we need to look at abstracting this so that projects can load in their own components when needed rather than making this massive file that does different things depending on the project task.

Happy for this to be made an issue and we look at it outside the PR

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
 
-  if (project.task == "ELM") {
-    const result = MultiVariateTimeSeriesDataSchema.safeParse(data);
+  let nonUfoDataOk = true;
 
-    if (!result.success) {
-      console.warn("ELM data is not available");
-      return;
-    }
+  if (!isUfo) {
```
jblake42: This is the sort of conditional logic I am talking about - it makes the code very hard to follow. This isn't your fault, but I think we need to give this a bit of thought as I don't think this is the way personally

toktagger/ui/src/app/components/tools/toolbar.tsx
jblake42, line 1: There also seem to be comments in here that look like debugging/tracking changes - these should not be left in

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+   * Listen to FrameView state events so toolbar stays in sync (instances, selection, counts, registry).
+   */
+  useEffect(() => {
+    if (!isUfo) return;
+    if (typeof window === "undefined") return;
+
+    const onState = (e: Event) => {
+      const detail = (e as CustomEvent<unknown>).detail;
+      const d = (
+        detail && typeof detail === "object" ? (detail as UfoStateDetail) : {}
```
jblake42: The internal type assertion is not needed as you do it again to the whole ternary result.

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+  const [selectedInstanceId, setSelectedInstanceId] = useState<string | null>(
+    null,
+  );
+
+  const [instanceCounts, setInstanceCounts] = useState<Record<string, number>>(
+    {},
+  );
+
+  /**
+   * Listen to FrameView state events so toolbar stays in sync (instances, selection, counts, registry).
```
jblake42: I am not the biggest fan of using these custom events to keep things synchronised. I feel like this would be a good use for contexts like I used for the other tooling. Since the toolbar is a fairly high level it may be a challenge as to where we put the context, but I can't see it causing a massive issue. This does work, I just worry about if this adds complexity and strange global variables as opposed to the tested context approach.

jblake42: The major issue I see with relying on an event system is that if an event is missed / a component loads after a component is loaded then it breaks the synchronisation

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+            const inst = next.find((p) => instanceKey(p) === d.selectedKey);
+            setSelectedInstanceId(inst ? inst.id : null);
+          } else if (d.selectedKey == null) {
+            // allow FrameView to clear selection
+            setSelectedInstanceId(null);
+          }
+        }
+      }
+
+      if ("selectedClassName" in d) {
```
jblake42: I think the `nullish coalescing operator` means we do not really need the outer conditional check

abdullah-ukaea:
I think we are ready for code review now on this branch @jblake42 @wk9874 @samueljackson92 
No more features / functionality will be added in this PR unless it is necessary. 
Happy to do any bug fixing / minor refactoring and software updates to address and pass the code review for this tooling.

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+  /**
+   * Per-instance usage scanner (keeps the instance count badges updated).
+   */
+  useEffect(() => {
+    if (!isUfo) return;
+    if (typeof window === "undefined") return;
+
+    const keyPrefix = "anno::w3c::" + `app://p/${project_id}/s/${sample_id}/`;
+
+    let stopScan: (() => void) | null = null;
```
jblake42, line 731: I'm a little bit confused about the whole start/stop thing - could you explain what it does please?

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -44,6 +46,55 @@ import { JumpDetectionTool } from "../annotators/jump";
...
+type UfoStateDetail = {
+  profiles?: UfoWireProfile[];
+  selectedKey?: string;
+  selectedClassName?: string | null;
+  lastClassName?: string;
+  classRegistry?: Record<string, number>;
+  profileCounts?: Record<string, number>;
+};
+
+declare global {
```
jblake42: Overwriting the Window type is a little confusing especially in the global scope - if there is a way to avoid it I would. There may not be though

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+
+    const raw = await collect();
+    const w3cList: ImageAnnotation[] = Array.isArray(raw)
+      ? (raw as ImageAnnotation[])
+      : [];
+
+    const cocoFrames = w3cToCocoFrames(w3cList, true);
+    const videoBBoxes = cocoFramesToVideoBBoxes(cocoFrames);
+
+    return videoBBoxes as unknown as Annotation[];
```
jblake42: How come you cast via `unknown` is it a linter thing?

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+  };
+
+  const handleUfoSave = async () => {
+    try {
+      const payload = await collectUfoPayloadForBackend();
+      const response = await saveAnnotations(project_id, sample_id, payload);
```
jblake42: `project_id` can be `undefined`

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
-          viewParams={viewParams}
-          setViewParams={setViewParams}
-          plotProps={plotProps}
-        />
-      ),
-    });
+  const handleUfoNextSample = async () => {
+    try {
+      const payload = await collectUfoPayloadForBackend();
+      await saveAnnotations(project_id, sample_id, payload);
```
jblake42: Same for this `project_id` and the one below

jblake42: I won't comment everyone but there are quite few more below

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+          console.error("Failed to fetch data:", err);
+          setErrorMessage("Failed to fetch shot.");
+        }
+      } else {
+        setErrorMessage("Please enter a number.");
+      }
+    };
+
+    return (
+      <SearchField
```
jblake42: On my system I cannot input any data into the box - it looks as if it is immediately overwritten with a blank string

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
     <Provider theme={defaultTheme} height="100vh">
-      <View overflow="auto" height="100vh">
-        <Flex
-          direction="column"
-          alignItems="center"
-          justifyContent="center"
-          gap="size-100"
-          width="100%"
-        >
+      {isUfo ? (
```
jblake42: Again, this is the place where if we put some work into how we best compose this component it would make it so much clearer imo - we wouldn't need layers of conditions

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+
+          <div className="pl-4 pr-4 pb-4 pt-2">
+            <UfoShotSearch />
+          </div>
+
+          <div className="pl-4 pr-4 pb-4">
+            <div className="max-w-[16rem] mx-auto mb-4">
+              <div className="mb-2">
+                <Flex gap="size-100" alignItems="center" wrap>
+                  <Button
```
jblake42: This button component is malformed - missing `variant`. Also, do we need this? It looks a bit out of place?

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+  const selectedInstanceKey = (() => {
+    const inst = instanceProfiles.find((p) => p.id === selectedInstanceId);
+    return inst ? instanceKey(inst) : null;
+  })();
+
+  if (!isUfo && !nonUfoDataOk) {
+    return null;
+  }
 
   return (
```
jblake42, line 883: The UFO toolbar is not very light mode friendly

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+          <div className="pl-4 pr-4 pb-4 pt-2">
+            <UfoShotSearch />
+          </div>
+
+          <div className="pl-4 pr-4 pb-4">
+            <div className="max-w-[16rem] mx-auto mb-4">
+              <div className="mb-2">
+                <Flex gap="size-100" alignItems="center" wrap>
+                  <Button
+                    isQuiet
```
jblake42: This is depracated

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+
+              <div className="mb-1">
+                <Flex
+                  gap="size-100"
+                  alignItems="center"
+                  justifyContent="center"
+                  wrap
+                >
+                  <Button
+                    isQuiet
```
jblake42: Deprecated and teh component is missing `variant`

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -419,195 +466,655 @@ export default function ToolBar({
...
+                    UNSAFE_className="!px-2.5 !py-1.5 text-xs"
+                    onPress={async () => {
+                      try {
+                        await ufoClearCurrent();
+                        ToastQueue.positive("Cleared current frame.", {
```
jblake42: I'm unsure if we need positive feedback for events that provide immediate feedback to the user - for example it is clear that the action happened because the annotations dissappear. Realistically the toast is discreet enough that it wouldn't be an annoyance

toktagger/ui/src/app/frames/components/adapters.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * LocalStorage-backed store for W3C-style image annotations for a single image/frame.
+ *
+ * - Keyed by a stable `sourceKey` (see `buildSourceKey`).
+ * - Returns an empty array on any failure or malformed data.
+ *
+ * Storage layout:
+ *   localStorage["anno::w3c::" + sourceKey] = JSON.stringify(ImageAnnotation[])
+ */
+export function W3CImageFormat(sourceKey: string) {
```
jblake42, line 28: Nice adapter pattern

jblake42 (commented):
I've made it though some of the files but still have a way to go. Pushing these now so they don't get lost. I think the local storage approach you have taken is very clever, I just wonder if a more standardised approach to align to other tooling could be simpler? I haven't really gotten into depth yet so I am far from confident in my judgement

docker-compose.dev.yml
jblake42: I think you may have already corrected this

toktagger/api/core/data_loaders.py
abdullah-ukaea: @wk9874

docker-compose.dev.yml
abdullah-ukaea: I agree, addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/0c9888c27deb3bf3dd1895646a1e742b2101a2e5

toktagger/api/core/data_loaders.py
wk9874: Good shout

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+  if (!a || (typeof a !== "object" && typeof a !== "function")) return null;
+  return a as AnnotatorApi;
+}
+
+function isImageAnnotation(v: unknown): v is ImageAnnotation {
+  return (
+    !!v &&
+    typeof v === "object" &&
+    "id" in v &&
+    typeof (v as { id: unknown }).id === "string"
```
jblake42: The line above type-narrows in a way that we don't need to cast to `{ id: unknown }` - `v.id` is guarenteed

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+} | null;
+
+type AnnotatorApi = {
+  getAnnotations?: () => unknown;
+  setAnnotations?: (anns: ImageAnnotation[], replace?: boolean) => void;
+  on?: (event: string, cb: (...args: unknown[]) => void) => void;
+  off?: (event: string, cb: (...args: unknown[]) => void) => void;
+};
+
+function asAnnotatorApi(a: unknown): AnnotatorApi | null {
```
jblake42: This slightly worries me as the logic in this function does not guarentee it is of type `AnnotatorApi` for an unknown type so this does nothing to improve the type safety. It seems like you are only using this to cast something of type `Annotator` which already satisfies the `AnnotatorApi` so could be a direct cast - I fear having this generic function would encourage people to use the it on a different type and assume it would work when it is not guarenteed.

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+  if (!a || (typeof a !== "object" && typeof a !== "function")) return null;
+  return a as AnnotatorApi;
+}
+
+function isImageAnnotation(v: unknown): v is ImageAnnotation {
```
jblake42, line 115: As before, this type guard implies you can pass anything in to it and it will confirm if it is an `ImageAnnotation` but that is not the case. For example `ImageAnnotation` requires a `target` type which is not confirmed in this guard. I think we should be more accurate if we are including these sorts of guards as it makes it look like we are type safe when we are not

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+  return (
+    !!v &&
+    typeof v === "object" &&
+    "id" in v &&
+    typeof (v as { id: unknown }).id === "string"
+  );
+}
+
+function toAnnoList(got: unknown): ImageAnnotation[] {
```
jblake42, line 137: This is a good example of why these type guards are dangerous. In this function if I pass an array that is filled with `{id: string}` objects it will output a `ImageAnnotation` array - but this list does not satisfy the type assigned to it which can lead to UI crashes if we try to access the `target` attribute

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+  return [];
+}
+
+function deepClone<T>(value: T): T {
+  return JSON.parse(JSON.stringify(value)) as T;
+}
+
+// --- Rectangle-only guard (filters out polygons or unknown shapes) ---
+function isRectangleAnno(a: unknown): boolean {
+  const sel = (a as { target?: { selector?: unknown } } | null)?.target
```
jblake42: You only ever pass in `ImageAnnotation` types into this function. Why not change the input type so that you can make use of the typing e.g.:
```
function isRectangleAnno(a: ImageAnnotation): boolean {
  const sel = a.target.selector
  return sel.type === ShapeType.RECTANGLE
}
```
This seems a lot cleaner. The only risk I see is if we are being type unsafe in other places - but that should be resolved there rather than here

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+  const s = sel as { type?: unknown; value?: unknown };
+  if (s.type === "RECTANGLE") return true;
+  if (typeof s.value === "string")
+    return /xywh=(pixel|percent):/i.test(s.value);
+  return false;
+}
+
+export const AnnoBridge = Object.assign(
+  forwardRef<
```
jblake42: `forwardRef` is a deprecated approach: [react docs](https://react.dev/reference/react/forwardRef)

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+    const dirtyRef = useRef(false);
+
+    // Buffered overlay apply (one setAnnotations per animation frame).
+    const pendingRef = useRef<ImageAnnotation[] | null>(null);
+    const flushingRef = useRef(false);
+    const lastAppliedRef = useRef<string>(""); // JSON signature of last applied overlay
+
+    const sig = (anns: ImageAnnotation[]) =>
+      JSON.stringify(
+        anns.map((a) => ({ id: a.id, t: a.type, tSrc: a.target?.source })),
```
jblake42: The typing is off here. I think it stems from the unsafe type assertions leading to it looking like you are passing in an `ImageAnnotation` array when you actually are not. The solution I think is to add additional typing

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+      );
+
+    // Double-RAF to ensure Annotorious has settled before re-enabling event handling
+    const doubleRAF = useCallback(async () => {
+      await new Promise<void>((r) =>
+        requestAnimationFrame(() => requestAnimationFrame(() => r())),
```
jblake42, line 238: I'm not a huge fan of this, it seems like a bit of a code smell however I understand sometimes things like this are necessary

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+        requestAnimationFrame(() => resolve());
+      });
+
+      const batch = pendingRef.current;
+      pendingRef.current = null;
+
+      if (batch && anno) {
+        const s = sig(batch);
+        if (s !== lastAppliedRef.current) {
+          const api = asAnnotatorApi(anno);
```
jblake42: What is the reason for this cast? anno is of type `Annotator` which has the funciton you want to call? e.g.:
```
if (batch && anno) {
        const s = sig(batch);
        if (s !== lastAppliedRef.current) {
          await runSilently(() => anno.setAnnotations(batch, true))
          lastAppliedRef.current = s;
        }
      }
```
I can't see where a typing or linting error would come from? If this is possible I would prefer it as it avoids the need for the unsafe type guard

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+      }
+
+      flushingRef.current = false;
+    }, [anno, runSilently]);
+
+    // Shared writer: normalize, keep lastByIdRef in sync, and return the normalized list (rectangles only)
+    const normalizeWrite = useCallback(
+      (raw: ImageAnnotation[]) => {
+        // Rectangle-only: drop anything not a rectangle
+        const rects = (raw || []).filter(isRectangleAnno);
```
jblake42: If `raw` is of type `ImageAnnotation[]` why is the `|| []` needed?

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+          if (!registry[keyLower]) {
+            const profileId = getSelectedProfile?.() ?? null;
+
+            const fixedId = FIXED_CLASS_REG[keyLower] ?? 1;
+            registry = {
+              ...registry,
+              [keyLower]: {
+                id: String(fixedId),
+                name: selectedClass,
+                profileId,
```
jblake42, line 367: The typing of this is wrong

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+      [getSelectedClassName, getSelectedProfile],
+    );
+
+    useImperativeHandle(
+      ref,
+      (): BridgeHandle => ({
+        isAnnotatorReady: () => !!anno,
+
+        persistWorkingNow: async (currentKey: string) => {
+          const api = asAnnotatorApi(anno);
```
jblake42: What is the reason for this cast when all you do is call a funciton that is already contained within the type?

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+          return normalizeWrite(stamped);
+        },
+
+        clearOverlaySilently: async () => {
+          const api = asAnnotatorApi(anno);
+          if (!api) return;
+
+          await runSilently(() => {
+            api.setAnnotations?.([], true);
```
jblake42: Same question, surely this method call is contained within `anno` without the type cast?

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+          });
+        },
+
+        hydrateOverlay: async (list: ImageAnnotation[], currentKey: string) => {
+          const api = asAnnotatorApi(anno);
+          if (!api) return false;
+
+          currentKeyRef.current = currentKey;
+
+          const src = Array.isArray(list) ? list.filter(isRectangleAnno) : [];
```
jblake42: `list` is of type `ImageAnnotation[]` is it not guarenteed to be an array?

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+          const stamped: ImageAnnotation[] = src.map((a) => ({
+            ...deepClone(a),
+            target: { ...(a.target ?? {}), source: currentKey },
```
jblake42: `a.target` should always be defined - if we need this check it is because we have been type unsafe

jblake42 (commented):
Sorry, still working my way through

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+            saveClassRegistry(registry);
+          }
+        } catch (err) {
+          console.warn("Auto quick-add class failed:", err);
+        }
+      },
+      [getSelectedClassName, getSelectedProfile],
+    );
+
+    useImperativeHandle(
```
jblake42, line 386: The use of this hook is interesting. There are a few places where non-imperitive behaviour is being carried out - specifically the getters towards the end. It is preferred where possible to pass declaritive information like this through props. This may not be necessary but it is something to consider

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+      if (!anno) return;
+
+      const onAnyChange = async () => {
+        if (suppressPersistRef.current) return;
+
+        // Any create/update/delete from the user marks this sample dirty
+        dirtyRef.current = true;
+
+        const api = asAnnotatorApi(anno);
+        const rawFull = toAnnoList(api?.getAnnotations?.());
```
jblake42: From inspection I don't see why this wouldn't work:
```
const rawFull = toAnnoList(anno.getAnnotations());
```
Removing the need for the guard

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+          includeTrackIds,
+          classRegistry,
+        );
+
+        // Auto quick-add logic; no-op when onAutoQuickAdd is not provided
+        if (includeTrackIds && normalized.length > 0 && onAutoQuickAdd) {
+          const firstByInstance = new Map<string, string>();
+          let duplicateAnno: ImageAnnotation | null = null;
+
+          for (const a of normalized) {
```
jblake42: There is a lot of use of single letter variable names in fair complex situations, I think using more explanatory names would be beneficial throughout this PR. Things like `a` and `aid` are quite opaque from inspection

toktagger/ui/src/app/frames/components/bridge.tsx
```diff
@@ -0,0 +1,454 @@
...
+            await onAutoQuickAdd({ class_name: selectedClass });
+          }
+        }
+
+        await handleCreate(w3c);
+        await onAnyChange();
+      };
+
+      const api = asAnnotatorApi(anno);
+      api?.on?.("createAnnotation", onCreate);
```
jblake42: This is not type safe which is why TypeScript complains about this line and the destructor below. Based on the event definition any number of unknown inputs can be provided to the event, however the callback only accepts a specific type of parameter. Whilst this may be the only thing it can receive, the typing does not show this.

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+  SearchField,
+  Flex,
+  Text,
+  Picker,
+} from "@adobe/react-spectrum";
+
+/** ------------------------------------------------------------------
+ *  Toolbar — simple frame navigation controls (Prev/Next/Jump)
+ *  ------------------------------------------------------------------ */
+export function Toolbar({
```
jblake42: This doesn't appear to be used?

jblake42 (commented):
Another day, another small number of reviews

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+      <Text UNSAFE_className="text-sm font-medium mb-2 block text-white">
+        Class
+      </Text>
+      <Picker
+        aria-label="Class"
+        selectedKey={selectedClassName}
+        onSelectionChange={(key) =>
+          setSelectedClassName((key as string) || null)
+        }
+        placeholder="— Select class —"
```
jblake42, line 66: I am in two minds between just defaulting to the top class as I worry it will confuse users when they cannot draw onto the frame. It is either that or we should have a toast that warns them that no class is selected

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+        width="100%"
+      >
+        {LABEL_MAP.categories.map((c) => (
+          <Item key={c.name}>{c.name}</Item>
+        ))}
+      </Picker>
```
jblake42, line 70: Spectrum docs seem to prefer this approach for clarity:
```
<Picker
     aria-label="Class"
     items={LABEL_MAP.categories}
     selectedKey={selectedClassName}
     onSelectionChange={(key) =>
          setSelectedClassName((key as string) || null)
     }
     placeholder="Select class"
     width="100%"
>
     {item => <Item key={item.name}>{item.name}</Item>}
</Picker>
```
I'm happy if you think the other approach is cleaner

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+      <Picker
+        aria-label="Class"
+        selectedKey={selectedClassName}
+        onSelectionChange={(key) =>
+          setSelectedClassName((key as string) || null)
+        }
+        placeholder="— Select class —"
+        width="100%"
+      >
+        {LABEL_MAP.categories.map((c) => (
```
jblake42: I worry a little about the disconnect between this component and its parent due to you directly referencing the label map here. If for example it is changed so that the parent pulls data from an API to fetch the mapping this logic would break. Would it make more sense to pass the items as a prop into this component?

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+        Drawing is enabled after you pick a class.
+      </Text>
+    </div>
+  );
+}
+
+/** Profile type used by InstancePanel.
+ *  NOTE: This is a UI-level type describing per-instance tracking profiles.
+ */
+export type Profile = {
```
jblake42: Does this need to be exported? I can't see it being used anywhere else? Minimising exports helps to avoid conflicts

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+  profileCounts?: Record<string, number>;
+  showCreator?: boolean;
+}) {
+  const [open, setOpen] = useState(false);
+  const [className, setClassName] = useState<string>(
+    LABEL_MAP.categories[0].name,
+  );
+  const [trackId, setTrackId] = useState<string>("");
+
+  useEffect(() => {
```
jblake42: `useEffect` for pure react state-based logic can be a code smell. I think this could easily be handled in the event handler for the `Add Profile` button.

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+  onRequestDeleteAllInstances: () => void;
+  /** Cross-frame annotation counts for each profile.key */
+  profileCounts?: Record<string, number>;
+  showCreator?: boolean;
+}) {
+  const [open, setOpen] = useState(false);
+  const [className, setClassName] = useState<string>(
+    LABEL_MAP.categories[0].name,
+  );
+  const [trackId, setTrackId] = useState<string>("");
```
jblake42: Is it safer to initialise this to a valid track id rather than just blank - it is likely a non-issue but it seems like a more defensive approach

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+        })}
+      </div>
+    </div>
+  );
+}
+
+/** ------------------------------------------------------------------
+ *  ConfirmModal — generic confirm dialog used by destructive flows
+ *  ------------------------------------------------------------------ */
+export function ConfirmModal({
```
jblake42, line 317: Nice abstracted component

toktagger/ui/src/app/frames/components/ui.tsx
```diff
@@ -0,0 +1,432 @@
...
+    fromList.track_id.length > 0
+      ? fromList.track_id
+      : ownLabel?.track_id) || "—";
+
+  const dims = rectToDims(annotation);
+
+  const handleDelete = async () => {
+    await anno?.removeAnnotation?.(annotation.id);
+    // Two RAFs to ensure Annotorious state has settled before we notify
+    await new Promise<void>((r) =>
```
jblake42: I have tested without this delay - what behaviour requires these because I couldn't see any issues with it? Is it a rare race condition?

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
jblake42, line 1: I think a lot of the need for these guards and casts to unknown types stems from the fact we are incorrectly casting to other types elsewhere. I don't think this probably stems from just this PR but this is the reason I try and push back against casting unless it is absolutely necessary as it causes issues later down the line and removes the inherent safety that comes with typescript. I think we need to look to refactor when possible to move back to type safety and remove the needs for these guards.

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -174,76 +228,255 @@ export default function SamplePage() {
   });
   const [plotProps, setPlotProps] = useState<PlotProps>({
     colorMap: "Cividis",
-  }); // Set default color map
+  });
+
+  // ------------------------------
+  // UFO frame navigation driver
```
jblake42: It seems strange to me that this specific frame navigation logic is in this generic section. This sort of navigation logic is surely similar to the navigation logic of other sample types which are able to move the logic within the sample specific code. Do you think it would be possible to do a similar thing for the frame code or is there a specific requirement for it to be here?

toktagger/ui/src/app/frames/components/frames.tsx
```diff
@@ -0,0 +1,1436 @@
...
+    ufoInstanceProfiles?: InstanceProfile[];
+    ufoSelectedProfileId?: string | null;
+    ufoSelectedClassName?: string | null;
+    ufoSelectedTrackId?: string | null;
+    ufoSelectionSource?: "auto" | "explicit" | null;
+
+    ufoHasUnsavedChanges?: () => boolean;
+    ufoMarkSaved?: () => void;
+
+    ufoCollectForSave?: () => Promise<ImageAnnotation[]>;
```
jblake42: Another good example of why global variables can be dangerous. This appears to be defined in two places with different typing

toktagger/ui/src/app/frames/components/frames.tsx
```diff
@@ -0,0 +1,1436 @@
...
+export function FrameSearch({ onJump }: { onJump: (n: number) => void }) {
+  const [errorMessage, setErrorMessage] = useState<string>("");
+
+  const onSearchSubmit = (newValue: string) => {
+    if (newValue === "") {
+      setErrorMessage("");
+      return;
+    }
+
+    if (/^[0-9]*$/.test(newValue)) {
```
jblake42: There appears to be some redundancy here. The regex and `isFinite` check seem to do quite similar things. I personally would get rid of the regex and use `isInteger` to drive the `onJump` call or generate an error message

toktagger/ui/src/app/frames/components/frames.tsx
```diff
@@ -0,0 +1,1436 @@
...
+
+  // Local toast for notifications (quick-add, delete, etc.)
+  const [toastOpen, setToastOpen] = useState(false);
+  const [toastMessage, setToastMessage] = useState<string>("");
+
+  const showToast = useCallback((msg: string) => {
+    setToastMessage(msg);
+    setToastOpen(true);
+
+    if (typeof window !== "undefined") {
```
jblake42: Do you need the window, can you not just use?:
```
setTimeout(() => setToastOpen(false), 2000)
```

toktagger/ui/src/app/frames/components/frames.tsx
```diff
@@ -0,0 +1,1436 @@
...
+  const includeTrackIds = true;
+
+  // Use the persisted class registry (shared with the toolbar via localStorage)
+  const classRegistry: ClassRegistry = useMemo(() => loadClassRegistry(), []);
+
+  // Enable drawing if user has selected a class (armed), even if no instance exists yet.
+  const drawingEnabled = !!getSelectedProfile() || !!getSelectedClassName();
+
+  const frameNumber: number = typeof data.frame === "number" ? data.frame : 0;
+  const frameLabel = Number.isFinite(frameNumber) ? frameNumber : "?";
```
jblake42, line 283: This dual level type checking to confirm something that is typed as `number` is actually a number is either not necessary or a result of dodgy type assertion elsewhere

toktagger/ui/src/app/frames/components/lib.ts
```diff
@@ -0,0 +1,1214 @@
...
+  keyPrefix?: string; // e.g. "anno::w3c::app://p/<proj>/s/<sample>/"
+  onUpdate: (counts: InstanceCounts) => void;
+  chunkSize?: number;
+}): () => void {
+  const storage = getStorage();
+  if (!storage) return () => {};
+
+  const allKeys = Object.keys(storage);
+  const keys = options.keyPrefix
+    ? allKeys.filter((k) => k.startsWith(options.keyPrefix))
```
jblake42: Typescript complains about this due to type narrowing across closures. To get rid of this a better approach is:
```
const prefix = options.keyPrefix
  const keys = prefix
    ? allKeys.filter((k) => k.startsWith(prefix))
    : allKeys;
```

toktagger/ui/src/app/frames/components/lib.ts
```diff
@@ -0,0 +1,1214 @@
...
+  const bodyOut = [...p.mapped];
+
+  // If no existing class body was found, append one to both body & bodies
+  if (!pb.found && !p.found) {
+    const cb = classBody();
+    bodiesOut.push(cb);
+    bodyOut.push(cb);
+  }
+
+  return {
```
jblake42: This is the sort of type casting which I believe is causing the need to do type guarding in other places. It would be nice to try and refactor this

toktagger/ui/src/app/frames/components/lib.ts
```diff
@@ -0,0 +1,1214 @@
...
+  const bodiesOut = patch(asArray((a as unknown as UnknownRecord).bodies));
+  const bodyOut = patch(asArray((a as unknown as UnknownRecord).body));
+
+  if (!found) {
+    const cb = classBody();
+    bodiesOut.push(cb);
+    bodyOut.push(cb);
+  }
+
+  return {
```
jblake42: ...and here

toktagger/ui/src/app/frames/components/lib.ts
```diff
@@ -0,0 +1,1214 @@
...
+}
+
+export function w3cToCocoFrames(
+  list: ImageAnnotation[],
+  includeTracks = true,
+): CocoFrame[] {
+  const byFrame: Record<number, CocoFrame> = {};
+
+  for (const a of Array.isArray(list) ? list : []) {
+    const src = a?.target?.source;
```
jblake42: The typing is weird here again, `source` does not exist in `target`

jblake42 (changes requested):
This is really good work to get this tool working to this level. From testing it performs really well. It was a big PR so there are a reasonable number of comments which is to be expected. I do have a couple of fears around the approach which might be unjustified. It is very complicated and makes use of approaches which the rest of the codebase does not use such as browser local storage, custom events and global variables. This makes things quite complex and I worry what it will do for maintainability in the future. From my initial interpretation of this (which is limited as I have only really looked at it for a couple of days) I believe that use of a similar approach to other tooling using contexts would work and would bring it into alignment with the rest of the code and perhaps simplify things. I think it would be worth a discussion about whether we could refactor at some point to try and simplify.
This doesn't take away from the fact this is very impressive and will help us start to prove we can annotate frames effectively. Well done!

toktagger/ui/src/app/frames/components/lib.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/61f34ce29bc2e273df56be2fe5b7ead5247cec54

toktagger/ui/src/app/frames/components/lib.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/13a5f28844fc5bee438821de0528ba1bf5f188eb

toktagger/ui/src/app/frames/components/lib.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/fa18f13c01bbf0f06b22598795082500b0877725

toktagger/ui/src/app/frames/components/lib.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/673d920f7ba5ac4f99fd299a052f278d1476e96d

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: Thanks for spotting this. Since  useAnnotator() already returns an Annotator | null,  the isAnnotatorApi check here isn’t doing anything. I’ll simplify this to a plain null-check and call getAnnotations directly.

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: Yep I agreed. useAnnotator() already gives us an Annotator, so the extra isAnnotatorApi check here is redundant; we can just null-check anno and call setAnnotations directly.

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: yep, list is already typed as ImageAnnotation[], so the Array.isArray guard is redundant. I’ll simplify it to list.filter() to match the type

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/254c87c4d40c34587db3fd2e462e293e76788a6c

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/b952446209ef657500bb82764ffc38515a45e191

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/b952446209ef657500bb82764ffc38515a45e191

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/b952446209ef657500bb82764ffc38515a45e191

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/2ae3f624b9c921e163429803362044d2049c894c

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/5243604ecf3e017a172c01b3c57cb7c04da03ab1 and https://github.com/ukaea/toktagger/pull/137/commits/2b7e3743c05b3f39eadf2212d9f3500100a7ae91

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/4a2748f0b31473609683e987df1b365e18883f9f

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/b72ca0357cb7449d0f1411cf025a4e85d806b054

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/2e4200314f12ff8f3863c98b8c586ba59483d7b4

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/964a9739e1097520a5593a94050fa305f0c3c3f8

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/402aa726de9b32c0db5f5eb0d1ccdd533fbdbab1

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/ae3eef1d069e564924744fe80149b1c0dae9c8b8

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/07031bdccea16dd6ba454ec830099ac86522bb23

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/6cca80340bd70c3e40fc6ab50f515c76fbf56d6b

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/aa3efc120a01410d3f3bcfeb0773d1bdde7e1676

toktagger/ui/src/app/frames/components/frames.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/cbe807fb99c33af4cb20634534474834c73c83a7

toktagger/ui/src/app/frames/components/frames.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/65c1363d11e3c93ae438ba53e3d70c049e6a9bd8

toktagger/ui/src/app/frames/components/frames.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/899dfb78ca57567e1d4068b785adf730bd210cf3

toktagger/ui/src/app/frames/components/frames.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/4eb4c7b37e5a8d7cd572081ec134fdd92c720f48

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/2f04a0c2c3772f6ad4a004823e33894581ed9b18

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/2218d66ef788c9e5bd7674f7adf6593b11f72ccd

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/b92b2018bf14509efb22c2567594bbc6f2c10540

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: The “start/stop” pattern is just there to manage a chunked async scan safely. scanInstanceCountsChunked walks localStorage in small setTimeout(0) chunks and returns a stopScan() cancel function; we call stopScan() before starting a new scan (and on unmount) so we don’t end up with multiple overlapping scans racing to call setInstanceCounts with stale/partial results while a new scan is already in progress. The setInterval simply re-triggers a fresh scan every second so the instance profile count badges stay accurate as FrameView writes/clears annotations in localStorage.

abdullah-ukaea: added some comments to help with readability of this section https://github.com/ukaea/toktagger/pull/137/commits/0c22d56fcc59dba06d0cc0bb854288aaaff258e9 I plan to add more comments to all the files when i am at the end of this code review

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: not sure to be honest, its not a linter thing Direct as Annotation[] compiles cleanly, so I’ve removed the unknown cast in https://github.com/ukaea/toktagger/pull/137/commits/80cbd99544f2e96ca7d092a08b1afde5faf05060

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/3233ea60c3dd3d5aac37a39cffa37e11576221f8

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: I agree about defaulting to the top class to make the user experience better, should I add this now?

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/45dd6b5c4ef8fb7b5826e4df040b7e430b5ca4f5

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/a6d5cb3d51189dbdaae2ed33bef4b00f79004a02

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/67c87f8dace1b603888d79ac52ab327fec49d4d1

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: I agree, addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/96dd0fe9f3f9855e696214ef431d96ab645d4025

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: Addressed in commit  https://github.com/ukaea/toktagger/pull/137/commits/aaff5089098b22e74d6845e4698394c6b7158bac

toktagger/ui/src/app/frames/components/ui.tsx
abdullah-ukaea: Thanks for spotting this. I played around with this and couldn’t reproduce any state issues so I removed the double RAF.

Addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/e993d2fb95c37f1dab6e285965e0c83fb2ecc896

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
abdullah-ukaea: Good points in commit https://github.com/ukaea/toktagger/commit/a7bbeb57faa7903b45e6315aa890ca34ad2f7b77 I tried to reduce the need for casts/guards by aligning the frontend with the actual backend contract. For UFO I removed the frame-bounds typing/parsing because the backend doesn’t return min/max today, and kept frame: null so the backend remains the source of truth for initial frame selection. Happy to reintroduce strict bounds/types once the API exposes them so we can remove the remaining defensive glue.

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
abdullah-ukaea: Addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/192f236dd25c9930997f609d88935a770c820ad7

You are right, SamplePage should stay task-agnostic so I moved the UFO-only navigation into the UFO module ufoview in frames.tsx. Kept backend-driven initial frame selection in SamplePage to preserve existing fetch semantics.

toktagger/ui/src/app/frames/components/bridge.tsx
abdullah-ukaea: Will try to address in the refactor PR

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: Will try to address in the refactor PR - future work

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: Will try to address in the refactor PR

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: Will try to address in the refactor PR / future work

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: Will try to address in the refactor PR / future work

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: Will try to address in the refactor PR / future work

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: future work / refactor PR

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/e7f02c8e7718e2f95c3610dc41c72f18fc2942bf

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/1614f2b9083b7a77c672cf5e3f912ce25dbe77dd

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/1614f2b9083b7a77c672cf5e3f912ce25dbe77dd

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/1614f2b9083b7a77c672cf5e3f912ce25dbe77dd

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/eceb97165e38862d71c96e06111fac1f2c763039

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/137/commits/880b2532ab2966efdd9b37276311d3c13fc60927

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea: addressed in commit http://github.com/ukaea/toktagger/pull/137/commits/9b1b0f92952e9f1449ebedb5ad2c0e3553fe44ff

toktagger/ui/src/app/components/tools/toolbar.tsx
jblake42: Created an issue

toktagger/ui/src/app/frames/components/bridge.tsx
jblake42: Will see if it is needed after refactor

jblake42 (approved):
I am happy for this to be merged based on the functionality side of it and with the caveat that we should look to refactor the implementation in order to simplify

samueljackson92:
I shall give this a quick review & road test tomorrow.

samueljackson92 (changes requested):
@abdullah-ukaea I am not going to test as deeply as @jblake42 as I trust he has reviewed the details (probably better than I would). However, one quick thing I've noticed is that the images (on my screen) are very small and there's no way to zoom. 

Zooming might be beyond this PR, but perhaps we can make the size of the images reactive to the users window size? I cannot really see the UFOs to be able to tag them with the current implementation as the image is so tiny.

image

samueljackson92:
Also it looks and reacts really nicely... 🚀

samueljackson92:
@abdullah-ukaea We've merged the other PR now. Seems there are a few conflicts with toolbar & page (as to be expected) but it does look too terrible. Can you take a look?

abdullah-ukaea:
Note: this branch doesnt work at the moment, currently in the middle of fixing it after solving the merge conflict. Do not merge.

abdullah-ukaea:
> @abdullah-ukaea I am not going to test as deeply as @jblake42 as I trust he has reviewed the details (probably better than I would). However, one quick thing I've noticed is that the images (on my screen) are very small and there's no way to zoom.
> 
> Zooming might be beyond this PR, but perhaps we can make the size of the images reactive to the users window size? I cannot really see the UFOs to be able to tag them with the current implementation as the image is so tiny.
> 
> image

I have tried to address this in commit https://github.com/ukaea/toktagger/commit/26dfdbb5434d78d7515cfc083f98abd7dc4c5c50

samueljackson92:
Nice. Looks more reasonable. One further thing I've spotted:

If I keep click forward then I get an error:

image

Same happens if I click backwards as well. I think that if we cannot get the next frame (because we're at the end or the beginning) we should just stay on the current frame, but alert the user.

abdullah-ukaea:
> Nice. Looks more reasonable. One further thing I've spotted:
> 
> If I keep click forward then I get an error:
> 
> image
> Same happens if I click backwards as well. I think that if we cannot get the next frame (because we're at the end or the beginning) we should just stay on the current frame, but alert the user.

should be fixed now

samueljackson92 (approved):

PR #146, opened 2026-01-12
Add draft pypi workflow

samueljackson92:
Adds a github workflow for pushing the package to pypi.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,9 +1,10 @@
 name: Publish to PyPI
 
 on:
-  release:
-    types: [published]
-  workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
+    workflow_dispatch:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Build package
+        run: uv build
+
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
+
```
wk9874, line 57: I would add a step which publishes to test-pypi first, and so doesn't try to push broken distros to real pypi if it fails

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install uv
+        run: pip install uv
+
+      - name: Build package
+        run: uv build
+
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v4
+        with:
```
wk9874, line 31: Can probably add a short retention time to this

wk9874 (changes requested):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
```
wk9874: Could have it dispatched on a tag? Then when we're ready for a release we can tag it and thisll deploy automatically

wk9874 (changes requested):

samueljackson92:
@wk9874 I've made the changes as suggested above.

wk9874 (approved):
Looks good to me!

PR #162, opened 2026-01-26
Video Annotation Tooling Refactor

### LINKED ISSUE #156: Refactor Frame Annotation
It would be nice to remove the need for complexities such as custom events, local storage and imperative handles as far as possible inorder to simplify and improve maintainability.

abdullah-ukaea:
This PR (162) is a refactor of PR https://github.com/ukaea/toktagger/pull/137/. The original video annotation tooling worked, but it was hard to maintain and extend because it relied on custom lifecycle glue, localStorage-driven state, and a growing set of “defensive guards” to keep Annotorious + React in sync. Addressing issue https://github.com/ukaea/toktagger/issues/156

This refactor brings the video tooling in line with the conventions used across TokTagger (especially the time-series tooling): React context for shared state, Annotorious-native lifecycle events, smaller components with clearer responsibilities, and a session model that makes it easier to add future UX/features without reworking core plumbing.

# What I did in my version 1 tooling (PR 137)
In v1, the video workflow was implemented as a set of “frames tooling” components (now deleted) that:
* Used a custom “bridge”/adapter layer to talk to Annotorious.
* Stored and restored per-frame annotations primarily via localStorage, which meant the “source of truth” could drift between:
**  Annotorious internal overlay state,
** React component state,
** and persisted localStorage copies.
* Needed additional runtime guards and custom lifecycle wiring to prevent feedback loops when programmatically setting annotations.
* Had larger files with more cross-coupled responsibilities (navigation, saving, syncing, UI actions, overlay normalization all intermingled), making iteration slower.

It worked, but it became fragile as soon as we started adding “real” requirements like multi-frame instance tracking, forward propagation, and enforcing consistent annotation metadata.

#  What I tried to do now
Rebuild the video tooling so it is:
* Annotorious-first (use Annotorious events like createAnnotation, updateAnnotation, deleteAnnotation, selectionChanged)
* React-context driven (a single in-memory session store for all frames)
* Invariant-based (normalize and enforce metadata on every change)
* Componentized (small, focused files: session, utils, host, sidebar, popup)

# What Each file does

## Core “session + state” model

**toktagger/ui/src/app/video/components/video-session.tsx**
This file defines the VideoSessionProvider and useVideoSession() hook, which act as the single in-memory source of truth for video annotation state. It stores per-frame Annotorious overlays, the current selection (class + track), and whether there are unsaved changes. It also provides helpers for common operations like clearing frames, deleting instances across frames, forward-propagating annotations, and collecting backend-ready payloads for saving. It includes a conservative “seed from DB once” helper so the session can start from backend annotations without overwriting user edits.

**toktagger/ui/src/app/video/components/video-utils.tsx**

This file contains pure helper utilities used by the session layer. It handles immutable Map operations (set/clear frames), deriving instance summaries across all frames for the sidebar, and performing bulk transforms like deleting one instance everywhere or forward-propagating into an empty next frame. It also includes track-id helpers like canonicalization and track-id generation (including readable IDs) so instance handling remains consistent and predictable.

**toktagger/ui/src/app/video/components/types.ts**

This file defines the shared types for the entire video annotation subsystem. It includes frame/index types, the session selection shape, instance summary types used by the sidebar, and the backend payload shape for video bounding boxes. It also defines the supported label set (V2_LABELS) and small helpers like building stable instance keys and stable frame source keys.

## Annotorious integration & UI

**toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx**

This file is the main integration point between Annotorious and the session store. It mounts the Annotorious context and keeps the current frame’s overlay synced between Annotorious and the session map. Any Annotorious change (create/update/delete) is routed through a single commit path that clamps geometry, enforces required metadata (label + track id), stamps a stable frame source key, and keeps the session state normalized. It also drives the selection popup UX and supports deleting the currently selected box.

**toktagger/ui/src/app/video/components/anno-utils.ts**

This file contains the small but important annotation normalization and conversion helpers. It enforces the invariant that every stored annotation has a label body (purpose="tagging") and a track id body (purpose="identifying"), and that we only keep rectangle annotations. It also provides conversion functions between Annotorious rectangles and the backend video_bounding_box format. In practice, this is the “make everything consistent” layer that makes saving and exporting reliable.

**toktagger/ui/src/app/video/components/AnnotationPopup.tsx**

This is a lightweight presentational component for the floating popup shown when a rectangle is selected. It displays the current label/track id (and optional geometry debug info) and exposes a small set of actions like deleting the selected box. Positioning is computed in the host so this component stays simple and UI-only. It’s intentionally minimal so it doesn’t couple to Annotorious internals or session logic.

## View chrome / navigation

**toktagger/ui/src/app/video/components/VideoView.tsx**

This component provides the “chrome” around the annotator: frame display, Prev/Next controls, and jump-to-frame. It also seeds the session from backend annotations once (only if the session is empty) so existing saved work appears immediately. When navigating forward, it optionally forward-propagates current annotations into the next frame if that frame has no annotations yet. It wires navigation into dataParams updates so the host page triggers the backend frame fetch

## Sidebar controls (tools panel)

**toktagger/ui/src/app/video/components/VideoSidebar.tsx**

This file defines the left sidebar UI used for video projects, replacing the generic toolbox used by other task types. It gives the user save + navigation controls (next sample, jump to shot) and calls “save if dirty” before navigation to avoid losing work. It also provides class selection (which arms drawing), shows a derived list of instances across frames, and supports destructive actions with confirmation flows. The only small piece it persists in localStorage is the last selected class, purely to reduce clicks between samples.

**toktagger/ui/src/app/video/components/ui_elements.tsx**

This file contains UI-only reusable components used by the video sidebar. It provides the class picker panel, instance list panel, and a lightweight confirm modal used for destructive actions. These components are intentionally decoupled from session implementation details so they stay easy to reuse and modify. The logic-heavy behavior stays in the session and sidebar, while this file focuses on rendering.

# Files modified in this PR
* toktagger/ui/src/app/components/tools/toolbar.tsx
* toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
* toktagger/ui/src/types.ts
* toktagger/ui/src/app/contexts/SampleContext.tsx

# Files created/deleted in this PR
* (created) toktagger/ui/src/app/video/components/anno-utils.ts
* (created) toktagger/ui/src/app/video/components/AnnotationPopup.tsx
*  (created)toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
* (created) toktagger/ui/src/app/video/components/types.ts
* (created) toktagger/ui/src/app/video/components/ui_elements.tsx
* (created) toktagger/ui/src/app/video/components/video-session.tsx
* (created) toktagger/ui/src/app/video/components/video-utils.ts
* (created) toktagger/ui/src/app/video/components/VideoSidebar.tsx
* (created) toktagger/ui/src/app/video/components/VideoToolbar.tsx
* (created) toktagger/ui/src/app/video/components/VideoView.tsx

* (deleted) toktagger/ui/src/app/frames/components/adapters.ts
* (deleted) toktagger/ui/src/app/frames/components/bridge.tsx
* (deleted) toktagger/ui/src/app/frames/components/frames.tsx
* (deleted) toktagger/ui/src/app/frames/components/lib.ts
* (deleted) toktagger/ui/src/app/frames/components/ui.tsx

# Future Work in later PRs

* Add polygon tool support (alongside rectangles)
* Pan/zoom/drag UX for the image view
* Timeline/transport controls for faster navigation through frames
* Deeper backend integration (e.g., loading existing stored frame annotations into localStorage/hydration path
* Improve jump-to-frame UX (direct frame number input + enter)

closes #156


toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -79,278 +48,132 @@ const SampleDataBreadCrumbs = ({
...
+    dataParams,
+    setDataParams,
+    isLoading,
+    error,
+  } = useSample();
+
+  const isVideo = project?.task === TaskType.Video;
+
+  // Hooks must be called unconditionally (no early returns before this).
+  const [videoFrame, setVideoFrame] = useState<number | null>(null);
```
jblake42: Well done for cleaning up this file however this logic should still not be located here. I think the best place is perhaps in the overall frame annotation context. This file should have no specific view logic in

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -79,278 +48,132 @@ const SampleDataBreadCrumbs = ({
...
-      let nextViewParams: ViewParams = vp;
-
-      if (project.task === "MHD") {
-        nextViewParams = {
-          ...vp,
-          name: "spectrogram",
-          nperseg: 256,
-        } as SpectrogramViewParams;
-      }
+  // Only hard-block on loading for non-video.
```
jblake42: What is the rationalle for having different loading handling for video? Is it to do with annotorious loading?

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -79,278 +48,132 @@ const SampleDataBreadCrumbs = ({
...
-              annotations={annotations}
-              setAnnotations={setAnnotations}
-              dataParams={dataParams}
-              setDataParams={setDataParams}
-              plotProps={plotProps}
-            />
-          </div>
-        </div>
+        <Flex>
+          {isVideo ? (
```
jblake42, line 114: This aligns with what we discussed with having the toolbar located within the context however we will need to standardise with the context ordering refactor

toktagger/ui/src/types.ts
```diff
@@ -192,7 +220,7 @@ export const DataParamsSchema = z.object({
 });
 export type DataParams = z.infer<typeof DataParamsSchema>;
 export const ImageDataParamsSchema = DataParamsSchema.extend({
-  frame: z.number(),
+  frame: z.number().nullable(), // the backend passes frame can be null?
```
jblake42: @wk9874 is this right - I would expect the backend to always provide a frame personally but maybe not if we can have single images

toktagger/ui/src/app/video/components/anno-utils.ts
jblake42, line 1: Can we be consistent with file naming please

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -0,0 +1,401 @@
...
+  const [byFrame, setByFrame] = useState<ByFrameMap>(() => new Map());
+  const [dirty, setDirty] = useState(false);
+
+  const [selection, setSelectionState] = useState<Selection>({
+    className: null,
+    trackId: null,
+    source: null,
+  });
+
+  const frameKey = useMemo(
```
jblake42, line 201: What is the reason for inclusing `useMemo` and `useCallback` - react automatically handles this but it is fine if there is a specific reason?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+  const markSaved = useCallback(() => setDirty(false), []);
+
+  const setSelection = useCallback((next: Selection) => {
+    setSelectionState(next);
+  }, []);
+
+  const createNewInstanceForClass = useCallback(
+    (className: string) => {
+      const cname = (className || "").trim() || "UFO";
```
jblake42: Reference to UFO

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+  const next = new Map(byFrame);
+  next.set(nextFrame, seeded);
+  return next;
+}
+
+/**
+ * Allocate the next numeric track id for a class by scanning existing instances.
+ * This is used when we want strictly increasing numeric ids (legacy-style).
+ */
+export function nextTrackIdForClass(
```
jblake42, line 324: Would it be simpler just to keep a track of the last assigned track ID in the context and increment that rather than scan every time? When we first load exisiting annotations this value could be initialised. The only issue would be they would not be consecutive but I don't really see why that is an issue.

If this doesn't take very long to complete I am happy with this approach though

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+  byFrame: ByFrameMap,
+  className: string,
+  getLabelTrack: (a: ImageAnnotation) => {
+    className?: string | null;
+    trackId?: string | null;
+  },
+): string {
+  const key = (className || "").trim();
```
jblake42: Do you need this check for a nullish value - we have to pass in a string

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+  if (!key) return "1";
+
+  let max = 0;
+  for (const list of byFrame.values()) {
+    for (const a of list ?? []) {
```
jblake42: Again, list must be an array - it can't be null

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+  return s.replace(/\s+/g, "-").toLowerCase();
+}
+
+/**
+ * Best-effort coercion to a non-empty track id.
+ * Used primarily in equality checks and deletion matching.
+ */
+export function ensureTrackId(trackId: string): string {
+  const c = canonicalizeTrackId(trackId);
+  return c || "1";
```
jblake42: `c` must be a string

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+export function deepClone<T>(v: T): T {
+  return JSON.parse(JSON.stringify(v)) as T;
+}
+
+/**
+ * Normalize a track id into a stable, readable key for comparisons and lookups.
+ * We keep this intentionally conservative (no heavy validation), just formatting.
+ */
+export function canonicalizeTrackId(trackId: string): string {
+  const s = String(trackId ?? "").trim();
```
jblake42: `trackId` is a string

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -0,0 +1,401 @@
...
+      const nextId = nextTrackIdForClass(byFrame, cname, getLabelTrack);
+      const trackId = String(nextId);
```
jblake42: What is the reason for parsing a string as a string?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -0,0 +1,401 @@
...
+  }, [deleteInstanceAcrossFrames, selection.className, selection.trackId]);
+
+  /**
+   * Copies current frame annotations into `nextFrame` if it's empty.
+   * Any copied annotations get their target.source updated to match the destination frame.
+   */
+  const forwardPropToNextIfEmpty = useCallback(
+    (nextFrame: FrameIndex) => {
+      setByFrame((prev) => {
+        const next = forwardPropagateIfEmpty(
```
jblake42: This function is only called here. Is there a reason to pass a callback into the function rather than just pass the data directly and have all of the logic within the `forwardPropogateIfEmpty` function?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+      const byF = new Map<number, ImageAnnotation[]>();
+
+      for (const a of dbAnnotations) {
+        const parsed = VideoBoundingBoxSchema.safeParse(a);
+        if (!parsed.success) continue;
+
+        const vb = parsed.data;
+        const key = buildSourceKey({ projectId, sampleId, frame: vb.frame });
+        const anno = videoBBoxToAnno(vb, key);
```
jblake42: The typing is a little off here. It actually suprised me but `null` is not compatible with `undefined`

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -0,0 +1,401 @@
...
+    (dbAnnotations: Annotation[]) => {
+      if (dirty) return;
+      if (byFrame.size > 0) return;
+      if (!dbAnnotations || dbAnnotations.length === 0) return;
+
+      const byF = new Map<number, ImageAnnotation[]>();
+
+      for (const a of dbAnnotations) {
+        const parsed = VideoBoundingBoxSchema.safeParse(a);
+        if (!parsed.success) continue;
```
jblake42: Do we need to flag if the backend gives incompatible data somehow?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -0,0 +1,401 @@
...
+/**
+ * Normalizes a raw overlay list into our session invariants:
+ * - rectangles only
+ * - stamps the frame source key
+ * - ensures class/track bodies exist (allocating track ids when needed)
+ * - optional per-instance de-duplication within the frame
+ *
+ * The caller is responsible for writing the returned list into session state.
+ */
+export function commitOverlayToSession(args: {
```
jblake42: To me the naming is off here as this is really just a translation utility function - it doesn't commit anything. If this is the case I think it would make more sense for this to be in the utility file

toktagger/ui/src/app/video/components/VideoView.tsx
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * Chrome and controls around the frame annotator:
+ * - Displays the current frame number
+ * - Prev/Next navigation and jump-to-frame
+ * - Seeds the session overlay from backend annotations (one-shot)
+ *
+ * Note: the frame number shown here comes from the session. The outer page is
+ * responsible for updating the session frame when the backend image changes.
+ */
+function VideoChromeV2(props: {
```
jblake42: I'm not a massive fan of this naming

toktagger/ui/src/app/video/components/VideoView.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  onPrev?: () => void;
+  onNext?: () => void;
+  onJump?: (n: number) => void;
+}) {
+  const session = useVideoSession();
+  const { seedFromDbIfEmpty } = session;
+
+  // Seed session state from backend annotations once (no-op if the session already has data).
+  useEffect(() => {
```
jblake42: It seems a little bit long winded to pass these annotations down through the components only for it to be handled by the context provider. Can we pass directly to the context and have it handled there?

toktagger/ui/src/app/video/components/VideoView.tsx
```diff
@@ -0,0 +1,223 @@
...
+ * - Prev/Next navigation and jump-to-frame
+ * - Seeds the session overlay from backend annotations (one-shot)
+ *
+ * Note: the frame number shown here comes from the session. The outer page is
+ * responsible for updating the session frame when the backend image changes.
+ */
+function VideoChromeV2(props: {
+  imageBase64: string;
+  dbAnnotations: Annotation[];
+  onSaveBackend: (payload: Annotation[]) => Promise<void>;
```
jblake42: Am I missing something or is this not used?

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+
+  return changed ? out : list;
+}
+
+/**
+ * Top-level host that provides the Annotorious context and renders the annotator.
+ */
+export function FrameAnnotatorHost(props: { imageBase64: string }) {
+  return (
+    <Annotorious>
```
jblake42: In my opinion the annotorious context is one of first things we should have in the DOM (within the video annotation view). That way the custom context will have direct access to annotations rather than passing around via props which I could image adds chances for errors or performance issues. It also means the tool bar would more easily have access to the context.

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+/**
+ * Core annotator integration:
+ * - syncs per-frame overlay between Annotorious and the session store
+ * - assigns track ids on create (based on current class selection)
+ * - keeps a small popup anchored to the selected rectangle
+ */
+function Inner({ imageBase64 }: { imageBase64: string }) {
+  const session = useVideoSession();
+  const annoRaw = useAnnotator();
+  const api = useMemo(() => asAnnotatorApi(annoRaw), [annoRaw]);
```
jblake42: The `useAnnotator` hook is a generic. If we call is like this:
```
const api = useAnnotator<Annotator<ImageAnnotation, ImageAnnotation>>();
```
it will provide us with the correctly typed inputs and outputs to functions meaning we don't need this type guarding and casting

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+              width: displayWidth ? `${displayWidth}px` : undefined,
+              maxWidth: "100%",
+              height: "auto",
+              maxHeight: "calc(100dvh - 240px)",
+            }}
+          />
+        </ImageAnnotator>
+
+        {selectedId && selectedMeta && popupPos && (
+          <AnnotationPopup
```
jblake42: Can we not make use of annatorious' provided popup API to avoid the complexity this adds with tracking position etc? Let me know if there is a reason for going down the custom route

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+      const cr = entries[0]?.contentRect;
+      if (!cr) return;
+      setContainerW(cr.width);
+    });
+
+    ro.observe(el);
+    return () => ro.disconnect();
+  }, []);
+
+  const displayWidth = useMemo(() => {
```
jblake42: Again, react should be memoizing anyway

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+    if (!containerW) return undefined;
+
+    // Tune these three values as needed
+    const TARGET = 900; // "comfortable" tagging width
```
jblake42: Good, in the future it might be worth exposing these somewhere to allow it to be changed by the user for zooming etc. Your solution seems to be nicely responsive though!

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+
+  /**
+   * Track id allocator for "AUTO" mode (no explicit trackId selected).
+   * The allocator memoizes per-class used ids across calls.
+   */
+  const makeAllocator = useCallback(() => {
+    const usedByClass = new Map<string, Set<string>>();
+
+    return (className: string) => {
+      const cls = (className || "").trim() || "UFO";
```
jblake42: className is a string

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+    () => session.getFrameList(session.frame),
+    [session],
+  );
+
+  /**
+   * Track id allocator for "AUTO" mode (no explicit trackId selected).
+   * The allocator memoizes per-class used ids across calls.
+   */
+  const makeAllocator = useCallback(() => {
+    const usedByClass = new Map<string, Set<string>>();
```
jblake42: Does this work? You seem to be using a variable defined inside the scope of a function within the callback. To me it doesn't look right and I worry about the robustness

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+      used.add(canonicalizeTrackId(next));
+      return next;
+    };
+  }, [session.byFrame]);
+
+  /**
+   * Convert annotation geometry in natural pixels to a popup position
+   * in the displayed image's client space.
+   */
+  const computePopupPos = useCallback(
```
jblake42: This is the sort of complexity I think the annotorious api would avoid

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+      setPopupPos(geom ? computePopupPos(geom) : null);
+    },
+    [api, clearPopup, computePopupPos],
+  );
+
+  /**
+   * Single commit point for all Annotorious mutations (create/update/delete).
+   * This normalizes bodies, clamps geometry to image bounds, and syncs the session store.
+   */
+  const onAnyChange = useCallback(
```
jblake42: If the context had access to annotorious as suggested, all of this could be handled there I think. This would avoid having to grab stuff from the context and then push stuff back

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+
+      if (selectedId) void refreshPopupForId(selectedId);
+    },
+    [api, session, makeAllocator, selectedId, refreshPopupForId],
+  );
+
+  /**
+   * When the frame changes, replace the annotator overlay from the session store.
+   */
+  useEffect(() => {
```
jblake42: What causes these to diverge? All I can see is the user draws an annotation, in which the context can catch that using the annotorious event or backend data is fetched in which case the updateAnnotations would be called which automatically updates the image render? I struggle to see the need for this

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+
+  /**
+   * Event wiring:
+   * - selectionChanged updates popup state
+   * - create/update/delete all funnel through onAnyChange
+   */
+  useEffect(() => {
+    if (!api?.on || !api?.off || !api?.getAnnotations) return;
+
+    const onSelectionChanged = (selected: unknown) => {
```
jblake42: Would using the built in pop up also simplify this as the only thing that would need to track selection is the context manager

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+      if (!cls) return;
+
+      const created = args[0] as ImageAnnotation | undefined;
+      const createdId = created?.id;
+      if (!createdId) {
+        onAnyChange();
+        return;
+      }
+
+      const raw = toAnnoList(api.getAnnotations());
```
jblake42: Typing is off here - using the suggestion for `useAnnotator` would fix this

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+  existingTrackIdsForClass,
+  uniqueReadableTrackId,
+} from "./video-utils";
+import { AnnotationPopup } from "./AnnotationPopup";
+
+/**
+ * Narrow API surface we rely on from Annotorious.
+ * We keep it small so the component doesn't depend on internal implementation details.
+ */
+type AnnotatorApi = {
```
jblake42: This shouldn't be needed either if you properly type the templated hook

toktagger/ui/src/app/video/components/FrameAnnotatorHost.tsx
```diff
@@ -0,0 +1,673 @@
...
+
+function isFunction(v: unknown): v is (...args: unknown[]) => unknown {
+  return typeof v === "function";
+}
+
+/**
+ * Best-effort adapter for the object returned by `useAnnotator()`.
+ * If the expected methods aren't present, we treat the API as unavailable.
+ */
+function asAnnotatorApi(a: unknown): AnnotatorApi | null {
```
jblake42: All of this type checking also wouldn't be needed I don't think

toktagger/ui/src/app/video/components/annotation-popup.tsx
jblake42, line 1: This is good and will be needed even if we migrate to the annotorious popup

toktagger/ui/src/app/video/components/VideoToolbar.tsx
jblake42, line 1: Could be a hangover from last version, but is this used?

toktagger/ui/src/app/video/components/VideoSidebar.tsx
```diff
@@ -0,0 +1,521 @@
...
+  }, []);
+
+  const classIdByName = useMemo(() => {
+    const map = new Map<string, number>();
+    for (const c of V2_LABELS) map.set(c.name, c.id);
+    return map;
+  }, []);
+
+  // Shape the derived session instances into the UI panel format.
+  const profiles = useMemo(() => {
```
jblake42: I think an unecessary memo again?

toktagger/ui/src/app/video/components/VideoSidebar.tsx
```diff
@@ -0,0 +1,521 @@
...
+      await _props.onSaved?.();
+    } catch (err) {
+      const msg =
+        err instanceof Error ? err.message : "Failed to save annotations.";
+      ToastQueue.negative(msg, { timeout: 5000 });
+    }
+  };
+
+  // Save only when needed (used before navigation).
+  const maybeSave = async () => {
```
jblake42: `saveIfDirty` - otherwise I have to refer back to the definition to know what it does

toktagger/ui/src/app/video/components/VideoSidebar.tsx
```diff
@@ -0,0 +1,521 @@
...
+
+  const confirmClearAll = () => {
+    setConfirmClearAllOpen(false);
+    session.clearAllFrames();
+  };
+
+  const cancelClearAll = () => setConfirmClearAllOpen(false);
+
+  return (
+    <Provider theme={defaultTheme} height="100vh">
```
jblake42: Do we not already set the provider earlier?

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Normalize an overlay list into a backend-safe shape:
+ * - keep rectangles only
+ * - stamp target.source with frameKey
+ * - ensure class label exists (uses fallback if missing)
+ * - ensure track id exists (uses fallback or allocator)
+ * - optionally dedupe to one annotation per (class, trackId) in this frame
+ *
+ * `list` is `unknown` because Annotorious event payloads are not always arrays.
```
jblake42: What situation are they not arrays - this suprises me

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+  const idx = list.findIndex((b) => b?.purpose === purpose);
+  if (idx >= 0) {
+    list[idx] = {
+      ...(list[idx] as AnnotationBody),
+      purpose,
+      value,
+    } as AnnotationBody;
+    return list;
+  }
+  list.push({ purpose, value } as unknown as AnnotationBody);
```
jblake42: Why do you go via unknown?

jblake42: Looking at the documentation do the bodies not need a unique identifier?

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+/** Read rectangle geometry (returns null if missing/invalid). */
+export function readRectGeometry(
+  a: ImageAnnotation,
+): { x: number; y: number; w: number; h: number } | null {
+  const g = a.target.selector.geometry as unknown;
+
+  if (!g || typeof g !== "object") return null;
+  const rec = g as UnknownRecord;
```
jblake42: I would personally do the rectangle check here otherwise we rely on the user calling the check before running this function.

I might also be inclined to cast the geometry to a custom type that contains the dimension fields so that they can then be accessed easily later on

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+
+  const x = rec["x"];
+  const y = rec["y"];
+  const w = rec["w"];
+  const h = rec["h"];
+
+  if (![x, y, w, h].every((v) => typeof v === "number" && Number.isFinite(v)))
+    return null;
+  if (w <= 0 || h <= 0) return null;
```
jblake42, line 184: The previous comment would then sort these typing issues

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+    ? list.filter(isImageAnnotationLike)
+    : [];
+  const out: ImageAnnotation[] = [];
+
+  for (const a of src) {
+    if (!isRectangleAnno(a)) continue;
+
+    const withSource: ImageAnnotation = {
+      ...a,
+      target: { ...a.target, source: frameKey },
```
jblake42: `source` is not part of the annotorious type definition - can be tighten this up at all?

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -0,0 +1,262 @@
...
+  const selector = {
+    type: ShapeType.RECTANGLE,
+    geometry,
+  } as unknown as ImageAnnotation["target"]["selector"];
+
+  const anno: ImageAnnotation = {
+    id,
+    bodies: [],
+    target: {
+      source: frameKey,
```
jblake42, line 313: Same here for `source`

toktagger/ui/src/app/video/components/ui_elements.tsx
jblake42, line 1: File naming again

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+ */
+export function existingTrackIdsForClass(
+  byFrame: ByFrameMap,
+  className: string,
+  getLabelTrack: (a: ImageAnnotation) => {
+    className?: string | null;
+    trackId?: string | null;
+  },
+): string[] {
+  const cls = (className || "").trim();
```
jblake42: Already a string

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+      if (tid) out.push(tid);
+    }
+  }
+  return out;
+}
+
+/**
+ * Allocate a fresh human-readable track id for this class, unique within the session.
+ */
+export function nextReadableTrackIdForClass(
```
jblake42: Is this used?

toktagger/ui/src/app/video/components/video-utils.ts
```diff
@@ -0,0 +1,352 @@
...
+}
+
+/**
+ * Collect all track ids for a given class across all frames.
+ * This is used to avoid allocating duplicates when creating new instances.
+ */
+export function existingTrackIdsForClass(
+  byFrame: ByFrameMap,
+  className: string,
+  getLabelTrack: (a: ImageAnnotation) => {
```
jblake42: The only two actual uses of this appear to use the same function here - do you envisage needing other functions? If not I think we simplify this

jblake42 (changes requested):
This is looking really nice, good work

jblake42:
I seem to be getting issues where if I update a frame, go back a frame and then return to the frame sometimes the frame does not maintain the update. I might have to try and show you it happening in our meeting

samueljackson92:
I've just been working on merging the video annotation stuff that is now on main into my spectrogram branch. One thing that has come up that I completely missed in the code review was that this should now use the `NavigationBar` for moving between shots, saving annotation, etc. instead of the video view having it's own logic.

See here: https://github.com/ukaea/toktagger/blob/main/toktagger/ui/src/app/components/tools/nav.tsx

Can we please also make this change here?

samueljackson92:
One more thing: The video annotation toolbar should also include the Import/Export and Shot Label widgets as well.

image

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
abdullah-ukaea: Addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/4eb9f61b21db985d194d84a1ac412b89ba9d177a. I moved the video frame-sync logic (frame state + backend/dataParams sync) out of sample_id/page.tsx and into VideoSessionProvider, keeping the sample page view-agnostic

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
abdullah-ukaea: I originally had video-specific loading as a quick patch because when navigating frames we refetch data, and hard-blocking on isLoading unmounted the video subtree which reset the session (losing per-frame overlays + derived instances). After testing, we can use the same loading approach as the other views by gating it on !data: if (isLoading && !data) return <LoadingView />; if (!data) return null; 

This keeps the video UI mounted during refetches (and avoids the session reset). Addressed also in commit https://github.com/ukaea/toktagger/pull/162/changes/4eb9f61b21db985d194d84a1ac412b89ba9d177a.

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: Addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/bc0a2b3a730ac9f0f0b968fe7037cec607ddf774. I’ve renamed the video components/util files to kebab-case for consistency and updated all affected imports.

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: Addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/bc0a2b3a730ac9f0f0b968fe7037cec607ddf774. I’ve renamed the video components/util files to kebab-case for consistency and updated all affected imports.

toktagger/ui/src/types.ts
abdullah-ukaea: I think your right, I will assume the backend always passes a frame. https://github.com/ukaea/toktagger/pull/162/changes/b215152ba6a7e3fa75c860800fa65c27be6e0815

toktagger/ui/src/app/video/components/VideoView.tsx
abdullah-ukaea: agreed. addressed in commit https://github.com/ukaea/toktagger/pull/162/commits/9125993e93af58e7457bfe036707a200ad6526b6

toktagger/ui/src/app/components/tools/toolbar.tsx
jblake42: This needs to be checked for `null` somewhere

jblake42: The sample context added by @samueljackson92 provides a way to retreive annotations - we should migrate to using this rather than a custom function that, from what I can tell, does the same thing

jblake42: It would be nice for this to be a tool like the other things added to the toolbar - since the tool is just defined as a react component I don't really see the issue. It would save you having to redo the navigation etc.

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -198,8 +233,48 @@ export function SampleProvider({
...
+            payload = await response.json();
+          } catch {
+            // ignore; payload stays null
+          }
+
+          const detail = extractDetail(payload);
+
+          // Video-only: treat missing frame as "boundary" and stay on last good frame.
+          if (projectData.task === TaskType.Video) {
+            const requestedFrame = (effectiveDataParams as { frame?: unknown })
```
jblake42: I do not like the typing of `dataParams` it seems to be mutated using casts rather than actually being defined

toktagger/ui/src/app/contexts/SampleContext.tsx
jblake42, line 212: @samueljackson92 just seen this whilst reviewing the PR - I personally think exposing a context function that allows the view to set the data parameters would make this cleaner as it will remove view-specific logic from this shared context. What are your thoughts?

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -43,23 +63,41 @@ export function getBodyValue(
...
       purpose,
       value,
     } as AnnotationBody;
+
     return list;
   }
-  list.push({ purpose, value } as unknown as AnnotationBody);
+
+  // New body: assign an id explicitly.
+  list.push({ id: newBodyId(), purpose, value } as unknown as AnnotationBody);
```
jblake42: Why via `unknown`? This is dangerous as it means when you define the object it doesn't have to follow the `AnnotationBody` definition

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -94,30 +132,60 @@ export function getLabelTrack(a: ImageAnnotation): {
...
+}
+
 /** True if the annotation target is a rectangle selector. */
 export function isRectangleAnno(a: ImageAnnotation): boolean {
   // ShapeType.RECTANGLE is already the string literal "RECTANGLE",
   // so we don't need a separate check for the string form.
   return a.target.selector.type === ShapeType.RECTANGLE;
 }
 
+function isFiniteNumber(v: unknown): v is number {
```
jblake42, line 146: Is just using `Number.isFinite(v)` not enough? It checks if it is a number as part of the test as far as I am aware:
```
Returns true if passed value is finite. Unlike the global isFinite, Number.isFinite doesn't forcibly convert the parameter to a number. Only finite values of the type number, result in true.
```

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -94,30 +132,60 @@ export function getLabelTrack(a: ImageAnnotation): {
...
 
+  const g = a.target.selector.geometry;
   if (!g || typeof g !== "object") return null;
-  const rec = g as UnknownRecord;
 
-  const x = rec["x"];
-  const y = rec["y"];
-  const w = rec["w"];
-  const h = rec["h"];
+  const rg = g as Partial<RectGeometry>;
```
jblake42, line 168: Good

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -146,18 +211,14 @@ export function normalizeOverlay(
   const enforceBothBodies = opts?.enforceBothBodies ?? true;
   const dedupeByInstance = opts?.dedupeByInstance ?? true;
 
-  const src: ImageAnnotation[] = Array.isArray(list)
-    ? list.filter(isImageAnnotationLike)
-    : [];
+  // Defensive filter: keep only annotation-ish object.
+  const src: ImageAnnotation[] = list.filter(isImageAnnotationLike);
```
jblake42: I don't understand how the input type is an `ImageAnnotation` and the output type is an `ImageAnnotation`. If it is unknown and we need to check it, `list` should not be of type `ImageAnnotation`. If it is known that list is `ImageAnnotation[]` then why are we performing the check?

toktagger/ui/src/types.ts
```diff
@@ -220,7 +220,7 @@ export const DataParamsSchema = z.object({
 });
 export type DataParams = z.infer<typeof DataParamsSchema>;
 export const ImageDataParamsSchema = DataParamsSchema.extend({
-  frame: z.number().nullable(), // the backend passes frame can be null?
+  frame: z.number(),
```
jblake42: Just to confirm, this aligns with the backend right?

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -0,0 +1,152 @@
...
+              width: displayWidth ? `${displayWidth}px` : undefined,
+              maxWidth: "100%",
+              height: "auto",
+              maxHeight: "calc(100dvh - 240px)",
+            }}
+          />
+        </ImageAnnotator>
+
+        {/* Built-in Annotorious popup positioning (no manual geometry->pixel mapping) */}
+        <ImageAnnotationPopup
```
jblake42, line 146: Nice, this seems to clean things up

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
```diff
@@ -0,0 +1,149 @@
+import type { ImageAnnotation } from "@annotorious/react";
+import { getLabelTrack, readRectGeometry } from "./anno-utils";
+
+type UnknownRecord = Record<string, unknown>;
+
+// Our app stores a frame key on target.source (not present in upstream Annotorious types).
+type VideoImageAnnotation = ImageAnnotation & {
```
jblake42: Have we not already defined this - why not make it a shared definition?

toktagger/ui/src/app/video/components/video-navigation-bar.tsx
jblake42, line 1: Is this not has been replaced with a standardised component for navigation

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -315,6 +385,233 @@ export function VideoSessionProvider(props: {
...
+   */
+  const desiredOverlay = useMemo(
+    () => byFrame.get(frame) ?? [],
+    [byFrame, frame],
+  );
+
+  useEffect(() => {
+    if (!api?.setAnnotations) return;
+
+    const cur = api.getAnnotations ? api.getAnnotations() : [];
```
jblake42: Can we not just accept that it exists - annotorious defines that it exists

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -115,16 +135,60 @@ export function useVideoSession(): VideoSessionCtx {
...
-  const { projectId, sampleId, frame, setFrame, children } = props;
+  const { projectId, sampleId, children } = props;
+
+  const api = useAnnotator<Annotator<ImageAnnotation, ImageAnnotation>>();
+
+  /**
+   * Guard against feedback loops: when we call `api.setAnnotations`, Annotorious
+   * emits update events. During that window we ignore change handlers.
+   */
+  const suppressRef = useRef(false);
```
jblake42: Needing this is a little bit of a code smell to be - I feel like it points to there being multiple sources of truth

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -115,16 +135,60 @@ export function useVideoSession(): VideoSessionCtx {
...
+  const frameFromBackend = useMemo(() => {
+    if (!props.data) return 0;
+    const maybe = props.data as unknown as { frame?: number };
+    return maybe?.frame ?? 0;
+  }, [props.data]);
+
+  useLayoutEffect(() => {
+    if (!props.data) return;
+
+    const dp = props.dataParams as unknown as {
```
jblake42: This sort of thing is caused by the weak typing employed elsewhere

jblake42 (commented):
Good improvements - we'll discuss in the meeting

toktagger/ui/src/app/contexts/SampleContext.tsx
samueljackson92: There is a `setDataParams` exposed by `useSample()`. But feel free to modify/adapt if it doesn't work for your use case.

toktagger/ui/src/app/components/tools/toolbar.tsx
jblake42: Ignore this

jblake42: Don't ignore this

samueljackson92: Are you ok @jblake42 ? 😆

jblake42: Who knows anymore 😭

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/f568e2af956f00c88d12f3ba3bb1e51c2367a349

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/b0adca9be79f0509109a1f30cb1b021ef1bf195f

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: yep your right, addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/00b24781ec8d43301c2ddcaecc9f4629ec405d84

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
abdullah-ukaea: agreed, addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/ca818b5cb18481d9b921e6ced73cfa105dd46830

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: yep, addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/b4649c24b688391fdb08836d79f78c74af550da2

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/162/changes/4f2425543b0907bdfe394f44f6cf3382a8203188

toktagger/ui/src/types.ts
abdullah-ukaea: @wk9874 feel free to correct me if I am wrong but from my understanding the file https://github.com/ukaea/toktagger/blob/abdullah/video_annotation_refactor/toktagger/api/schemas/data.py shows that we can actually pass a None?

`class ImageParams(DataParams):
    name: LoaderType = LoaderType.IMAGE
    frame: int | None`

Which means I actually should make the zod schema on the front end actually accept nullable so I reverted this back in commit https://github.com/ukaea/toktagger/pull/162/changes/faad17ed2ef2ace32aa01ab788e02de3c34c263d

samueljackson92:
@abdullah-ukaea @wk9874 I've created https://github.com/ukaea/toktagger/pull/168 to deal with much improved project/sample creation, including custom annotation labels. 

I haven't updated video annotation to use the custom labels yet, as I will wait for this one to go in, then merge.

For @wk9874  in particular I have fixed being able to add multiple local image folders (so like `data/30241/frame_01.png` in #168

samueljackson92:
@abdullah-ukaea @wk9874 I've created https://github.com/ukaea/toktagger/pull/168 to deal with much improved project/sample creation, including custom annotation labels. 

I haven't updated video annotation to use the custom labels yet, as I will wait for this one to go in, then merge.

For @wk9874  in particular I have fixed being able to add multiple local image folders (so like `data/30241/frame_01.png` in #168

abdullah-ukaea:
ready for rereview @jblake42

toktagger/ui/src/app/video/components/video-navigation-bar.tsx
abdullah-ukaea: I kept VideoNavigationBar in this PR because the generic NavigationBar saves from SampleContext.annotations, but video edits live in VideoSessionProvider (Annotorious/byFrame) and only get converted via session.collectAllVideoBBoxes(). Using the standard nav.tsx right now would save stale/empty annotations, which is also why you see suppressRef/programmatic sync guards in the video session. I’m planning a follow-up PR to make SampleContext the single source of truth for video (commit at boundaries) so we can delete VideoNavigationBar and use the standardized nav everywhere.

If you want I could that in this PR also maybe take a look and let me know your thoughts

jblake42:
There definitely still appears to be issues with tracking of changes as you move through the frames - hopefully the recording shows this. The recording also shows an issue with the annotation being hard to select

https://github.com/user-attachments/assets/30a19c63-bf59-41d8-8db8-3a8bde9b6707

abdullah-ukaea:
> There definitely still appears to be issues with tracking of changes as you move through the frames - hopefully the recording shows this. The recording also shows an issue with the annotation being hard to select
> 
>  2026-02-18.14-51-46.mp4

will address this later this afternoon, I agree this needs to be fixed before it gets merged.

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -278,6 +263,11 @@ export default function ToolBar() {
   }
 
   const refreshAnnotations = async () => {
+    if (project.task === TaskType.Video) {
```
jblake42: The fact this logic differs for the video worries me slightly? Surely we should be perfoming the same actions no matter the view - to me it seems like updates from the DB are not fetched with the video logic. I know you said you were doing things differently at the moment for the frame annotation, this might be one of them that will be adjusted in the next PR.

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -301,7 +291,15 @@ export default function ToolBar() {
             <Header height="size-300" marginBottom="size-100">
               <span style={{ fontSize: "1.2rem" }}>Controls</span>
             </Header>
-            <NavigationBar project_id={project_id} sample_id={sample_id} />
+            {project.task === TaskType.Video ? (
```
jblake42, line 324: As discussed, this should be harmonized with the other views in the following PR

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -325,23 +323,27 @@ export default function ToolBar() {
...
-                {tools.map((item, i) => (
-                  <Disclosure key={i}>
-                    <DisclosureTitle>
-                      <span style={{ fontSize: "0.8rem" }}>{item.name}</span>
-                    </DisclosureTitle>
-                    <DisclosurePanel>{item.component}</DisclosurePanel>
-                  </Disclosure>
-                ))}
-              </Accordion>
+              {isVideo ? (
```
jblake42: Likewise with the harmonization

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -36,6 +37,7 @@ interface SampleContextType {
   dataParams: DataParams;
   viewParams: ViewParams;
   plotProps: PlotProps;
+  annotationLabels: { id: number; name: string }[];
```
jblake42, line 39: I don't think hoisting this state up to the context is massively helpful - we are looking to pull the categories from the project object that is already supplied by the context. I am happy for this to stay but the PR to add this functionality properly should probably revert these changes to avoid our context becoming bloated

toktagger/ui/src/app/contexts/SampleContext.tsx
jblake42, line 1: I have created an issue for this file as I would like to remove as much view specific code as possible

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -82,8 +82,8 @@ export function upsertBody(
 
     // Preserve existing body id if present; otherwise assign one once.
     const existingId =
-      typeof (prev as unknown as { id?: unknown })?.id === "string"
-        ? (prev as unknown as { id: string }).id || null
+      typeof (prev as { id?: unknown })?.id === "string"
```
jblake42: `prev` type already contains id - we don't need the cast, it adds no safety. Just check the type

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -197,22 +197,20 @@ export function readRectGeometry(
  * - ensure track id exists (uses fallback or allocator)
  */
 export function normalizeOverlay(
-  list: ImageAnnotation[],
+  list: unknown[],
```
jblake42: You have changed this to unknown, however I have traced this back and the only time this is used is to validate data which comes from the annotorious api. I really think this is not necessary, but it can stay as it is only an issue with tidiness

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -150,10 +150,13 @@ export function VideoSessionProvider(props: {
   const api = useAnnotator<Annotator<ImageAnnotation, ImageAnnotation>>();
 
   /**
-   * Guard against feedback loops: when we call `api.setAnnotations`, Annotorious
-   * emits update events. During that window we ignore change handlers.
+   * Annotorious emits create/update/delete events even when annotations are set
```
jblake42, line 149: This makes sense - as discussed I think there is probably a way to avoid this using a single point of truth but that will have to be experimented with

jblake42 (commented):
We'll discuss tomorrow - the tempermental issues with editting annotations is swaying me away from merging as is

***edit - just seen your comment above

abdullah-ukaea:
@samueljackson92 @jblake42 Could you both test this branch functionally and check everything looks okay?

samueljackson92 (changes requested):
Hi @abdullah-ukaea 

Some quick feedback from me. I have not reviewed the code.
 
1. I don't think tools should be nested like this:

Image

Should be a flat list. This toolbar is also missing the `Shot Labels` Tool.

2. Changing the video annotations labels in the projects page does not update in the UI

Image

For example, I changed "Other Anomaly" to "Other". Please see how this was done for other interfaces.

3. This tool doesn't seem to do anything?

Image

4. When I try and navigate between images I should get a "No more samples available" message, as there is only 1 sample. Also the previous button is not greyed out as in the time series interface. Instead I get:

image


6. Exporting, then importing video annotation does not refresh the interface. I have to refresh the page to get the updated bounding box displayed. Contrast this with the time series interface.

abdullah-ukaea:
> Hi @abdullah-ukaea
> 
> Some quick feedback from me. I have not reviewed the code.
> 
> 1. I don't think tools should be nested like this:
> 
> Image
> Should be a flat list. This toolbar is also missing the `Shot Labels` Tool.
> 
> 2. Changing the video annotations labels in the projects page does not update in the UI
> 
> Image
> For example, I changed "Other Anomaly" to "Other". Please see how this was done for other interfaces.
> 
> 3. This tool doesn't seem to do anything?
> 
> Image
> 4. When I try and navigate between images I should get a "No more samples available" message, as there is only 1 sample. Also the previous button is not greyed out as in the time series interface. Instead I get:
> 
> image
> 6. Exporting, then importing video annotation does not refresh the interface. I have to refresh the page to get the updated bounding box displayed. Contrast this with the time series interface.

@samueljackson92 @jblake42  @wk9874  I have updated the branch to address Sams comments. Could you retest the branch again?

samueljackson92:
@abdullah-ukaea the issues highlighted in my review are fixed for me now.

jblake42:
I have had a quick look through and I am happy for this to merge providing that issues are created for all of the points raised in this PR that have not been addressed. Some of the most important ones in my opinion are:

- Cleaning up the state logic so there is one source of truth
- Remove view-specific logic from the sample context
- Make use of standard navigation component (if not already doing that)
- Improve how view-specific data is shared to the toolbar

There are likely more so please have a look through the PR and pull them out into issues.

Great work on this refactor, we've ended up with a much more robust solution.

PR #163, opened 2026-01-27
Fix bug with unvalidated annotations not showing in UI

samueljackson92:
Also fix an issue with changing the types in multi-annotation selection.


toktagger/ui/src/app/components/annotators/peaks.tsx
```diff
@@ -56,15 +56,6 @@ export function PeakDetectionTool({
   useEffect(() => {
     const fetchData = async () => {
       if (!validSignal || !isEnabled) {
-        // Remove previous annotations from this annotator
-        setAnnotations((previousAnnotations: Annotation[]) => {
```
wk9874, line 60: Removing this completely would be a shame, since I think it is nice to be able to add/remove all annotations made from each annotator with the click of a button!

Instead, maybe we could do something like:


```js
export function PeakDetectionTool({
  project_id,
  sample_id,
}: PeakDetectionType) {
  const { annotations, setAnnotations, dataParams, data } = useSample();
  const [isEnabled, setIsEnabled] = useState<boolean>(() => {
    return annotations.some(
      (ann) => ann.created_by === AnnotatorTypes.PEAK_DETECTION
    )
  });

  ...

  useEffect(() => {
    const fetchData = async () => {
      if (!isEnabled) {
        // Remove previous annotations from this annotator
        setAnnotations((previousAnnotations: Annotation[]) => {
          const otherAnnotations = previousAnnotations.filter(
            (annotation: Annotation) =>
              annotation.created_by !== AnnotatorTypes.PEAK_DETECTION ||
              annotation.validated,
          );
          return otherAnnotations;
        });
        return;
      }
      else if (!validSignal) {
        return
      }

      const response = await fetch(
      ...
```

That way, as far as I can figure out:
* If you have (somehow) got unvalidated annotations from an annotator in the db, they will enable the tool which generated them, and be displayed on load
* If you turn the annotator off, unvalidated annotations from this annotator are removed
* If they are validated annotations, they'll always be displayed irrespective of the annotator being on / off

toktagger/ui/src/app/components/annotators/peaks.tsx
```diff
@@ -92,7 +83,8 @@ export function PeakDetectionTool({
       setAnnotations((previousAnnotations: Annotation[]) => {
         const otherAnnotations = previousAnnotations.filter(
           (annotation: Annotation) =>
-            annotation.created_by !== AnnotatorTypes.PEAK_DETECTION,
```
wk9874, line 95: Also please make the same changes to `modelPredictTool` in `toktagger/ui/src/app/components/tools/modelPredictSample.tsx`

wk9874 (changes requested):

samueljackson92:
I have made the changes as requested.

toktagger/ui/src/app/components/tools/modelPredictSample.tsx
```diff
@@ -18,16 +18,27 @@ type ModelPredictInfo = {
 };
 
 export function ModelPredictTool({ project_id, sample_id }: ModelPredictInfo) {
-  const { project, setAnnotations } = useSample();
-  const [isEnabled, setIsEnabled] = useState<boolean>(false);
+  const { annotations, project, setAnnotations } = useSample();
   const [selectedModel, setSelectedModel] = useState<string | null>(null);
+  const [isEnabled, setIsEnabled] = useState<boolean>(() => {
+    return annotations.some((ann) => ann.created_by === selectedModel);
```
wk9874: Ah damn, this wont work will it... Since `selectedModel` will be `null`, and hence this will always return not enabled, and then will immediately remove the model annotations (this is the only one which *really* needs to work, since the model will definitely be returning non validated annotations which should be displayed on load)

Maybe this check could be:

```js
const [isEnabled, setIsEnabled] = useState<boolean>(() => {
    return annotations.some((ann) => project.model_types.has(ann.created_by));
```
ie, if there are annotations from any ML model valid for this project, enable the tool?

wk9874 (changes requested):

toktagger/ui/src/app/components/tools/modelPredictSample.tsx
samueljackson92: Sounds like a sensible solution.

wk9874 (approved):
lgtm!

Issue #143, opened 2025-12-23
Sequential query strategy broken


wk9874:
Since we moved away from a persistent DataPool in the `/samples/next` endpoint, the query strategies no longer know which sample they are up to. Which means for the sequential QS, the same list of samples is provided from the DB each time, and so the same sample is always popped from the top

Also, the `/samples/next` endpoint currently returns the samples sorted by date created, but should probably instead return them sorted by `shot_id` as I expect that is what the user would expect the next button to cycle through

Suspect this is at least partially fixed by #120

samueljackson92:
I think this is done.

PR #170, opened 2026-02-10
Add state to query strategies

### LINKED ISSUE #165: Uncertainty query strategy is broken
This line uses `np.unique()` to remove duplicate sample IDs caused by multiple annotations on one sample:
https://github.com/ukaea/toktagger/blob/58af27f588feebeb2f328d93cb9d49689c2afc10/toktagger/api/core/query_strategy.py#L118

But `np.unique` returns the entries sorted, and so does not preserve the order of the samples based on annotations.

Also this reverse=True will make it return them in order uncertainty low -> high:
(https://github.com/ukaea/toktagger/blob/58af27f588feebeb2f328d93cb9d49689c2afc10/toktagger/api/core/query_strategy.py#L124)

But if you remove that, then this line will add all unannotated samples first:
https://github.com/ukaea/toktagger/blob/58af27f588feebeb2f328d93cb9d49689c2afc10/toktagger/api/core/query_strategy.py#L123

And finally, I normally start annotating by clicking on a random sample (or really the first sample in the table), to then press Next. But this line means that if ive clicked a non annotated sample, the Next will give me another unannotated sample, instead of the most uncertain
https://github.com/ukaea/toktagger/blob/58af27f588feebeb2f328d93cb9d49689c2afc10/toktagger/api/core/query_strategy.py#L62

wk9874:
The current way it is designed, the Uncertainty Query Strategy is very unreliable, see https://github.com/ukaea/toktagger/issues/165

This PR adds:
* Fixes to the Uncertainty QS to return you samples in the correct order, which is:
    1. Samples with unvalidated annotations, sorted highest to lowest uncertainty
    2. Samples with no annotations, sorted randomly
    3. Samples with validated annotations

* Adds a 'seenSampleIds' list of strings to the samples context, which appends/pops a sample ID each time you press Next/Previous. This does not persist across sessions so will be reset on a refresh (which I think is fine, and potentially is what I'd prefer to happen)

* `/samples/next` endpoint is now passed a list of seen sample IDs, instead of the current sample ID. This allows it to return you the correct next most uncertain sample, even if the order of the list changes between calls to /next (before, if you saved annotations on a sample, that sample would move near the end of the list in the query strategy, and so `next` would not give you the next most uncertain sample, but in fact a validated sample instead!)

* `/samples/previous` and the `get_previous_sample` methods in query strategies have been removed, in favour of popping from the seenSampleIds list and directly navigating to it. 

This has the effect of the other query strategies no longer to press Previous when nothing else has been seen. Eg, if youre on sequential, and load shot 100, you can no longer press Previous to go to shot 99. For me, this is an acceptable trade off to make Uncertainty QS work properly since thats the one I think is most important... It also has negligible impact on Random QS (who would care about going to the -1th random sample?)

To address the above, we could potentially implement separate buttons for next/previous Shot, which always move you sequentially through the shots. We could then remove the Sequential QS (and have an option for no QS?), and from any QS you can always step backward/forward through sequential shots. This could be a follow up PR, open to discussion on that!

Closes #165, #143


toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -37,13 +37,15 @@ interface SampleContextType {
   viewParams: ViewParams;
   plotProps: PlotProps;
   isLoading: boolean;
+  seenSampleIds: string[];
```
samueljackson92: Maybe rename as `visitedSampleIds`?

samueljackson92 (changes requested):
Some quick feedback from me:

 - We should cache the state in local [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage). This should prevent us loosing state on reload, but will loose state when the window/tab is closed.

 - In the sequential strategy, sorting the table (e.g. by shot ID) should also be reflected in the order we navigate.

wk9874:
Have addressed the two comments above:

> We should cache the state in local [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage). This should prevent us loosing state on reload, but will loose state when the window/tab is closed.

This is now cached in sessionStorage, with the key being the project ID. This means that if you flick between different projects, your previous button correctly cycles through the samples for that project only

> In the sequential strategy, sorting the table (e.g. by shot ID) should also be reflected in the order we navigate.

This was a bit more complicated since the sorting settings are on the samples table page, and the next button is on the individual sample page. Have added it so that the sort params can (optionally) be added to the URL of the page which you navigate to, and these are then passed to the endpoint when you press Next. You can still access sample pages without these params, and itll default to sorted by ascending shot ID.

wk9874:
Ready for rereview @samueljackson92

samueljackson92 (changes requested):
- When I try and navigate through shots with sequential querying, the "Jump to next shot" button always takes me to the first shot? This doesn't seem right. We should probably jump to the most recent "unvalidated" shot.
- As discussed, navigation in random ordering seems broken. 
  - Sometimes it only shows me 3, not 4 shots.
  - If I go out and back into another shot, it tells me I cannot navigate forward.
  - I think we should show the user random "unvisited" shots, then random "visited" shots from whereever they start.

wk9874:
To confirm expected functionality for each QS:

1. Uncertainty QS
    i. Shows you samples with unvalidated annotations in uncertainty high -> low order
    ii. Shows you all remaining unannotated samples, random order
    iii. Shows you samples with validated annotations, random order
    iv. Jump to next should jump to the next non-validated sample.

2. Sequential QS:
    i. Shows you samples in order based on how youve sorted them in the table, regardless of annotations / validated status
    ii. Jump to Next should from samples table should take you to the next non-validated sample within this order

3. Random QS:
    i. Shows you non validated samples, random order, regardless of if they have annotations or not (?)
    ii. Shows you validated samples, random order
    iii. Jump to next should jump to the next non-validated shot

Is this what you have in mind @samueljackson92 ?

samueljackson92:
Yes I think so.

tests/api/core/test_query_strategy.py
```diff
@@ -54,102 +53,48 @@ def test_sequential_strategy(samples, annotations):
...
-        next_sample = strategy.get_next_sample(next_sample_id)
+        next_sample = strategy.get_next_sample(visited_sample_ids)
         next_sample_id = next_sample.id
         assert next_sample == samples[i]
+        visited_sample_ids.append(next_sample_id)
 
-    # Should cycle back to start
-    next_sample = strategy.get_next_sample(next_sample_id)
+    # Should loop back to beginning
+    next_sample = strategy.get_next_sample(visited_sample_ids)
```
samueljackson92: No need to loop back to the beginning

samueljackson92 (changes requested):
- Sequential navigation breaks when I select a shot then navigate backwards. It seems to take me to the last shot?
 - I think sequential should work like random: it shouldn't be cyclical anymore. It should grey out the previous button when you're on the first shot. It should not let you navigate past the final shot.
 - I think the validated annotations indicator should be moved to the top above the controls:

Image

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -558,6 +561,25 @@ export default function ToolBar() {
...
+                }}
+              >
+                <Heading>
+                  {isValidated
+                    ? "Annotations Validated"
+                    : "Annotations Not Validated"}
+                </Heading>
+              </InlineAlert>
+            </Flex>
+          )}
```
samueljackson92, line 586: Can we move this to be it's own component?

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -139,6 +158,28 @@ export function SampleProvider({
   });
 
   const [isLoading, setIsLoading] = useState<boolean>(true);
+
+  const [isValidated, setIsValidated] = useState<boolean | null>(null);
+
+  const [visitedSampleIds, setvisitedSampleIds] = useState<string[]>(() => {
```
samueljackson92: I am a bit worried that `useSample` is becoming way too large and trying to do too many things. 

Consider refactoring this to either 1) be handled in the Page itself or 2) create a new context for the locally stored stuff.

samueljackson92: It looks like a lot of the new functionality is not related to getting the samples or annotations. I would try going with moving it to a new context.

samueljackson92 (changes requested):
Looking good, but a few refactoring bits on closer inspection...

toktagger/ui/src/app/contexts/SampleContext.tsx
wk9874: idk about this, it feels to me like they are quite related? Conceptually to me at least:

`isValidated` `visitedSampleIds` need to be updated every time a new sample is fetched, so makes sense that its value is defined in the context and updated each time a new sample's data is fetched in the refreshData function. I see this being somewhat similar to `isLoading` in that regard, which does live inside SampleContext

Agree on sortDescriptor tho, moving that into the nav bar

samueljackson92: What about using a context that depends on the `useSample` context?

Something roughly like:

```typescript
function SampleHistoryContext() { 
    const { sample } = useSample();
    const [vistedSampleIds, setVisitedSampleIds] = useState<blah>();
    useEffect(() => {
        ....
        setVisitedSampleIds(...);
    }, [sample]);
}
```

samueljackson92 (approved):

PR #172, opened 2026-02-16
Video Annotation:  Improve navigation to UFOs from Instance list 

### LINKED ISSUE #138: Image Annotation: Improve navigation to UFOs from Instance list
It's sometimes quite hard to click on the boxes you've drawn to move / resize / see the details about them, as you need to click the edge of the box. Could we add it so that:

1) If you click on an instance in the list on the left hand side, and that instance has an annotation on the frame you are looking at, it will highlight that frame so that it is selected for the user (who can then easily move / resize / delete it)

2) If you double click on an instance in the list, it automatically takes you to the first frame where there is an annotation from that instance, and again highlights it to the user as above

abdullah-ukaea:
NOT to be merged before the PR https://github.com/ukaea/toktagger/pull/162


abdullah-ukaea:
note: doesn't currently work post merge conflict currently fixing

samueljackson92:
@abdullah-ukaea can this be tackled now the other PR is in?

abdullah-ukaea:
> @abdullah-ukaea can this be tackled now the other PR is in?

Working on it now !

PR #192, opened 2026-03-02
Video Annotation: Improve navigation to UFOs from Instance list 

abdullah-ukaea:
This PR will address issue https://github.com/ukaea/toktagger/issues/138


abdullah-ukaea:
@wk9874  let me know if you have functional thoughts on this! Any feedback is welcome

wk9874:
Looks good! Like how clicking the 'jump to'  button also opens the dialogue on the image so I know which box I'm looking at :)

Few things:
1. Would be nice to sort the instances list in order of the frame in which they first appear, instead of (what I think it currently is) alphabetical. 
2. Might also be good to include the frame number(s) underneath the class in the instance list, ie here:
image

3. The query strategies are broken, I think because you are still using your own version of the nav bar instead of the main one? Is it possible to migrate to the main one? Otherwise if easier, you can just change your nav bar to act in the same way as the main nav bar when it comes to moving to previous/next shot - lmk if you need help with this

toktagger/ui/src/app/video/components/ui_elements.tsx
```diff
@@ -198,10 +202,14 @@ export function InstancePanel({
 
         {profiles.map((p) => {
           const count = profileCounts?.[p.key] ?? 0;
+          const canActivate = Boolean(
+            onActivate && typeof p.first_frame === "number",
```
jblake42: Can we replace the type check with a more obvious: `p.first_frame != null` - this checks for `null` and `undefined` but not `0`

toktagger/ui/src/app/video/components/ui_elements.tsx
```diff
@@ -64,6 +66,7 @@ export function InstancePanel({
   profiles,
   selectedKey,
   onSelect,
+  onActivate,
```
jblake42: This is not the most obvious name for this callback imo - I makes no reference to the fact it is used to jump to the first instance

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -497,23 +518,150 @@ export function VideoSessionProvider(props: {
...
+      });
+    },
+    [],
+  );
+
+  const selectAnnotationById = useCallback(
+    (id: string) => {
+      if (!id) return false;
+
+      const maybeApi = api as AnnotatorWithSetSelected | undefined;
```
jblake42: You have created a new type to explicitly include `setSelected` (which I think is unnecessary as it already includes that) - you then cast a plain annotator directly into this type with no further checks, proving it is likely that this additional type is not required. I cannot see how this mutation of the type has any benefit?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -497,23 +518,150 @@ export function VideoSessionProvider(props: {
...
+      const setSelected = maybeApi?.setSelected;
+      if (!setSelected) return false;
+
+      try {
```
jblake42: What errors have you seen that require this try/catch, does annotorious not abide by its type definition here?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -497,23 +518,150 @@ export function VideoSessionProvider(props: {
     [byFrame, frame],
   );
 
+  const overlayHasInstance = useCallback(
+    (list: ImageAnnotation[], req: FocusRequest) => {
+      const wantClass = (req.className || "").trim();
```
jblake42, line 516: Can there exist two instances with the same track ID but different classes? I though track IDs were unique. If they are unique, checking both seems redundent

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -497,23 +518,150 @@ export function VideoSessionProvider(props: {
...
+          return false;
+        }
+      }
+    },
+    [api],
+  );
+
+  const tryFocusPending = useCallback(
+    (rawOverride?: ImageAnnotation[]) => {
+      if (!api?.getAnnotations) return false;
```
jblake42: I know I keep going on about this, but I am interested whether you are adding these because you see an issue or just to be safe. `getAnnotations` should always exist. If we are seeing that it doesn't exist it could point to us being unsafe somewhere

abdullah-ukaea:
> Looks good! Like how clicking the 'jump to' button also opens the dialogue on the image so I know which box I'm looking at :)
> 
> Few things:
> 
> 1. Would be nice to sort the instances list in order of the frame in which they first appear, instead of (what I think it currently is) alphabetical.
> 2. Might also be good to include the frame number(s) underneath the class in the instance list, ie here:
> 
> image
> 3. The query strategies are broken, I think because you are still using your own version of the nav bar instead of the main one? Is it possible to migrate to the main one? Otherwise if easier, you can just change your nav bar to act in the same way as the main nav bar when it comes to moving to previous/next shot - lmk if you need help with this

1) fixed the instance list ordering in commit https://github.com/ukaea/toktagger/commit/448ee8532b20f6ee09f6860779cb3b9b12c554a8 

2) Had a quick discussion together and agreeed it didnt look great in the instance list but we have ideas for how to visualise the frames for the instances in a popup potentially, will create an issue and work on this later

3) query strategies are fixed in next PR https://github.com/ukaea/toktagger/pull/194

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -497,23 +518,150 @@ export function VideoSessionProvider(props: {
...
+        }
+        return false;
+      }
+
+      if (!selectAnnotationById(hit.id)) return false;
+
+      pendingFocusRef.current = null;
+      return true;
+    },
+    [api, desiredOverlay, overlayHasInstance, selectAnnotationById],
```
jblake42: There is a missing dependency here

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -497,23 +518,150 @@ export function VideoSessionProvider(props: {
...
+            : null,
+      };
+      setFocusNonce((n) => n + 1);
+    },
+    [],
+  );
+
   useEffect(() => {
-    if (!api?.setAnnotations) return;
+    if (!api?.setAnnotations || !api?.getAnnotations) return;
```
jblake42: Do we need these checks?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -614,6 +767,24 @@ export function VideoSessionProvider(props: {
     [api],
   );
 
+  useEffect(() => {
```
jblake42, line 753: I don't love that this sort of thing has to creep in. Could I just check what the reasoning for this is?

jblake42 (changes requested):
Functions well and the code looks pretty good. Only some small tweaks and then good to go. I am a little concerned about the need for handling animation frames again as I know that caused issues last time, but understand they may be needed

toktagger/ui/src/app/video/components/video-session.tsx
jblake42: Looks like you have already solved this

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: fixed! renamed to onJumpToFirstFrame

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: yep your right, removed this type

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: removed as it wasnt needed

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: trackId is class-scoped, not globally unique.
So the same track_id can exist in the same annotation session across different classes. 

So while in the same class e.g. minor ufo it is not possible to get the same track id, but in different classes it is possible. so I think I should keep this

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: no thanks for spotting this cleaned it up now .

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: No they are redundant, fixed now

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: So I think we need it, maybe I can remove it when I clean up my state in the next branch? not sure though.

this use effect is like the React-to-Annotorious sync boundary.

  - Annotorious keeps its own internal annotation store.
  - React state (byFrame/desiredOverlay, frame changes) does not auto-update that internal store.
  - This effect pushes desiredOverlay into Annotorious when those inputs change, then runs pending focus on the next frame tick.

abdullah-ukaea:
Think I addressed most of your concerns @jblake42 could you give it another look when you get a chance and let me know if its ready to merge? @wk9874 let me know if the instance list meets your request also.

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: yep done

toktagger/ui/src/app/video/components/video-session.tsx
jblake42: Yeah if that is the case we should keep it - in the future I think it could be worth trying to make this globally scoped as I could see it being confusing for users if it ever happened that the names overlapped as they would have to look at class and id to differentiate (as rare as that is). It could also possibly ease the logic to select elements?

I think this would be worth creating as issue for

toktagger/ui/src/app/video/components/video-session.tsx
jblake42: No problem, just checking. I think we should keep aiming to remove this sort of thing as we go where possible. Race conditions could end up being an issue otherwise

toktagger/ui/src/app/video/components/ui_elements.tsx
```diff
@@ -198,10 +203,14 @@ export function InstancePanel({
 
         {profiles.map((p) => {
           const count = profileCounts?.[p.key] ?? 0;
+          const canActivate = Boolean(
```
jblake42: Can you align this to the name change for clarity please

jblake42 (changes requested):
Great, all looking good. Please make the small change requested and have a look at the comment I have left unresolved regarding the unique IDs.

Following that we will merge

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/pull/192/commits/078f72daba4348386bbf5e57473e28b94e7c7d2d

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: Captured in issue https://github.com/ukaea/toktagger/issues/201

abdullah-ukaea:
I believe I have made all the relevant changes could @jblake42  or @samueljackson92  merge when you get a chance

jblake42:
LGTM

Issue #188, opened 2026-03-02
Video Annotation: Improve nextTrackIdForClass


abdullah-ukaea:
Would it be simpler just to keep a track of the last assigned track ID in the context and increment that rather than scan every time? When we first load exisiting annotations this value could be initialised. The only issue would be they would not be consecutive but I don't really see why that is an issue.

abdullah-ukaea:
Can we close this issue as PRs (https://github.com/ukaea/toktagger/pull/210 & https://github.com/ukaea/toktagger/pull/194 ) should have addressed this. @jblake42

jblake42:
Yes this can be closed

Issue #187, opened 2026-03-02
Video Annotation Toolbar Styling


jblake42:
The video annotation toolbar tools do not look very good in light mode

abdullah-ukaea:
Can we close this issue as PRs (https://github.com/ukaea/toktagger/pull/210 & https://github.com/ukaea/toktagger/pull/194 ) should have addressed this. @jblake42

jblake42:
Move to spectrum has improved this - closing

Issue #186, opened 2026-03-02
video annotation: Standardise video sample context rendering


abdullah-ukaea:
Standardise video sample context rendering by removing the isVideo conditional wrapper and aligning video tooling/context ordering with non-video sample views.

`
        <Flex>
          {isVideo ? (
`

abdullah-ukaea:
Can we close this issue as PRs (https://github.com/ukaea/toktagger/pull/210 & https://github.com/ukaea/toktagger/pull/194 ) should have addressed this. @jblake42

jblake42:
This has been addressed

Issue #185, opened 2026-03-02
Video Annotation: improve state logic and make use of standard nav component


abdullah-ukaea:
There needs to be some more refactoring done to 
- Cleaning up the state logic so there is one source of truth
- Make use of standard navigation component  instead of video-navigation-bar.tsx

abdullah-ukaea:
Can we close this issue as PRs (https://github.com/ukaea/toktagger/pull/210 & https://github.com/ukaea/toktagger/pull/194 ) should have addressed this. @jblake42

jblake42:
We are now using the standard nav functionality so this can be closed - there may be future work but that is outside the scope of this

PR #194, opened 2026-03-03
Video Annotation: More Refactoring

abdullah-ukaea:
**Merge PR https://github.com/ukaea/toktagger/pull/192 first!**

- Cleaning up the state logic so there is one source of truth addressing issue  https://github.com/ukaea/toktagger/issues/185
- Remove view-specific logic from the sample context addressing issue https://github.com/ukaea/toktagger/issues/186
- Make use of standard navigation component addressing issue https://github.com/ukaea/toktagger/issues/185
- Makes the toolbar light mode friendly addressing issue https://github.com/ukaea/toktagger/issues/187
- Replaces nextTrackIdForClass with buildNextTrackIdState and allocateNextTrackId addressing issue https://github.com/ukaea/toktagger/issues/188


abdullah-ukaea:
This will fix the broken query strategy that @wk9874 pointed out for PR https://github.com/ukaea/toktagger/pull/192 due to using the generic nav react component

abdullah-ukaea:
not ready for review yet, need to do some rebasing

abdullah-ukaea:
ready for review

abdullah-ukaea:
@samueljackson92 I reran the end to end test and it seemed to build this time.

samueljackson92:
Seems like @wk9874's E2E tests are not 100% stable.

wk9874:
Works well I think! Only *tiny* issues I've spotted: 

In light mode, the 'Jump to Frame' text is really hard to read
image

If you use 'Jump to Shot' searchbar on the left hand side, it will keep the current frame number you are looking at. This means that if you were to jump to shot which doesnt have that frame available you might get an error back. Should probably request the first available frame, like it does when you use the previous / next buttons?

wk9874:
> Seems like @wk9874's E2E tests are not 100% stable.

image

samueljackson92:
> If you use 'Jump to Shot' searchbar on the left hand side

@abdullah-ukaea just missing this second point from Matt then

abdullah-ukaea:
> > If you use 'Jump to Shot' searchbar on the left hand side
> 
> @abdullah-ukaea just missing this second point from Matt then

working on it right now, should be done soon was having some environment issues last friday which delayed me

abdullah-ukaea:
> Works well I think! Only _tiny_ issues I've spotted:
> 
> In light mode, the 'Jump to Frame' text is really hard to read image
> 
> If you use 'Jump to Shot' searchbar on the left hand side, it will keep the current frame number you are looking at. This means that if you were to jump to shot which doesnt have that frame available you might get an error back. Should probably request the first available frame, like it does when you use the previous / next buttons?

I found that commit https://github.com/ukaea/toktagger/pull/194/changes/b188613c4f6143f7a0dff7cdcbf51a4895d5eb09 fixes the issue described by Matt in his 2nd comment. could you take a look and test it functionally and let me know if everything works as intended. i only changed a few lines in sample context and video view everything else is the same. @samueljackson92  I still need to do some linting now though, before merging i wont pass prettier or eslint I think right now

abdullah-ukaea:
@samueljackson92 ready for review

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -103,16 +108,23 @@ function NextButton({
...
     await saveSampleAnnotations(
       project_id,
       sample_id,
-      annotations,
+      annotationsToSave,
       saveOnNavigate,
     );
+    if (saveOnNavigate) {
+      navAdapter?.afterSave?.();
+    }
```
samueljackson92: Why do we need to separate if statements with the same condition here? Can we combine with the below?

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -192,16 +205,23 @@ function PreviousButton({
...
```
samueljackson92: As above.

samueljackson92 (changes requested):
Sorry @abdullah-ukaea, two more very minor comments.

toktagger/ui/src/app/components/tools/nav.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/194/changes/46de3d2b9019e54c3151799ff297eae44f833757

toktagger/ui/src/app/components/tools/nav.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/commit/46de3d2b9019e54c3151799ff297eae44f833757

samueljackson92 (approved):

PR #195, opened 2026-03-04
Add support for 2D profiles

samueljackson92:
This pull request introduces the ability to annotate different types of 2D profiles with bounding box and polygon annotations. 

 - Added the ability create bounding box and polygon annotations
 - Added the ability to manipulate (delete, change type of) annotations on spectrogram view.
 - Refactored  threshold annotator to output lists of PolygonAnnotation objects instead of masks


abdullah-ukaea:
Would it be possible to disable the right click context menu ? as currently when a user is in spectogram view, and right clicks it gives the ability to use time series annotations which will confuse the user. @samueljackson92 


https://github.com/user-attachments/assets/8780b9c6-06a6-4ff3-8495-eee583c81d7b

abdullah-ukaea:
Also the box select tool available to the user in spectogram view seems to crash the UI completely when used. @samueljackson92 


https://github.com/user-attachments/assets/2f638ad9-d896-4bb2-8529-54046e421b56

jblake42:
> Also the box select tool available to the user in spectogram view seems to crash the UI completely when used. @samueljackson92
> box_select_crash_short.mp4

I have the same issue

jblake42:
Here are my thoughts on this. I think the functionality added by this PR is great and very important for this kind of data set. I do have some reservations:

- This functionality is a massive shift from how other time series data annotation works as it uses plotly's tooling directly. I fear that this adds a significant amount of confusion and I don't think the unification of this tooling with the others will be trivial as it relies on plotly's built in functionality (though I could be wrong)
- I think that all the functionality added by this could be done using the more flexible D3 approach that was decided upon for the time series plots - bounding boxes should be a trivial fix and polygon annotations in a way similar to the video annotation should also not be too difficult. The freedraw polygons will be slightly more challenging but possible. Making use of the D3 tooling especially after the refactor is pretty simple and will give us a consistent look and feel out of teh box
- There seems to be a number of stability issues relating to the tooling including visual artifacts when drawing and crashing when selecting
- My other concern is that this is a 50,000 change PR - I really think the changes needed to add this to the existing D3 ecosystem would be far easier and more maintainable as a number of us have a reasonable understanding of this already. I think this could run into a similar issue to the initial implementation of the video annotation tooling in that it is a vastly different way of doing things that has to be learnt by the team to maintain.
- One of my other biggest comments is that this tooling (bounding box and polygon) would most likely be useful for the time-series tooling so it is not as if these are even separate 'views'. This means that it is feasible for a user to need to use both styles of interaction in one plot which I think will be unintuitive.

My suggestion personally would be to park this PR and see if we can add these tools in one by one using the D3 approach and compare the ease of implementation. If it is impossible to add the functionality then we can relook at this and see how easy it is to unify, but if it is fairly easy I think we pivot that approach. I am happy to actually block some time out next week and at least get a bounding box demo up and running using D3 so we can make a judgement call.

jblake42:
Please see #306 for a proof of concept for the D3-based bounding boxes. Adding this tooling was pretty low-effort so I am confident we could support other tooling as well

scripts/full_setup.py
```diff
@@ -0,0 +1,78 @@
...
+    requests.put(f"http://localhost:8002/projects/{project_id}/samples", json=samples)
+
+
+def main():
+    shot_files = Path("./data/summary").glob("*.parquet")
+    shot_files = list(shot_files)
+    shot_ids = [int(path.stem) for path in shot_files]
+
+    project_id = create_project("Full ELM Project", "ELM", "parquet")
+    create_local_samples(
```
wk9874: Why is this in a new script rather than in `setup.py`?

tests/api/core/test_data_loaders.py
```diff
@@ -230,14 +230,10 @@ def test_fair_mast_dataloader():
...
     data_loader = data_loaders.FAIRMASTDataLoader()
-    # FAIR MAST server seems to be a bit flaky at the moment, causing intermittent CI failures
-    # Will catch TimeoutError here, and skip the test if encountered
     try:
         data = data_loader.get_sample(sample, params=DataParams(name="identity"))
-    except TimeoutError:
-        pytest.skip(
-            "Timeout error when connecting to FAIR MAST server - skipping test..."
-        )
+    except Exception:
```
wk9874: Why did this get loosened? This might catch errors thrown by get_sample which the test should be failing on

toktagger/api/core/annotators.py
```diff
@@ -141,18 +149,101 @@ def downsample_time_series(
     return time_coarse, signal
 
 
-def compute_stft(data: TimeSeriesData) -> np.ndarray:
+def _coords_to_flat_list(coords) -> List[float]:
```
wk9874: Typing on coords

toktagger/api/core/annotators.py
```diff
@@ -548,50 +639,114 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
-        threshold_value = np.percentile(values, self.params.percentile)
-        threshold_mask = values > threshold_value
-        return SpectrogramMask(
-            label="SpectrogramMask",
-            values=threshold_mask.tolist(),
-            created_by=AnnotatorTypes.SPECTROGRAM_THRESHOLD,
-        )
+        if isinstance(signal, TimeSeriesData):
+            dim_1, time, values = compute_stft(signal)
+            values = values  # Transpose to have shape (time, dim_1)
```
wk9874: `values = values`? Judging by the comment this may need to be transposed?

toktagger/api/core/annotators.py
```diff
@@ -548,50 +639,114 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
+
+        labeled_regions = measure.label(mask, connectivity=2)
+
+        if self.params.min_size > 0:
+            labeled_regions = morphology.remove_small_objects(
+                labeled_regions, min_size=self.params.min_size
+            )
+
+        # Convert each labeled region to polygons
+        polygons = []
```
wk9874: Would be good to type this

toktagger/api/core/annotators.py
```diff
@@ -548,50 +639,114 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
+                # Contour coordinates are (row, col), convert to (x, y)
+                contour_coords = [
+                    (int(np.round(c[1])), int(np.round(c[0]))) for c in contour
+                ]
+                contour_coords = [(time[x], dim_1[y]) for x, y in contour_coords]
+                poly = Polygon(contour_coords)
+                # if poly.is_valid and poly.area > 0:
+                polygons.append(poly)
+
+        polygons = [shapely_to_coco_style_annotation(polygon) for polygon in polygons]
```
wk9874: Since `polygons` changes here from a list of Polygon objects to a list of dictionaries, I would assign this to a new variable and type it

toktagger/api/core/annotators.py
```diff
@@ -603,6 +758,6 @@ def predict(self, data: MultiVariateTimeSeriesData) -> SpectrogramMask:
         AnnotatorTypes.CHANGE_POINT_DETECTION,
         AnnotatorTypes.JUMP_DETECTION,
     ],
-    Task.SPECTROGRAM: [AnnotatorTypes.SPECTROGRAM_THRESHOLD],
+    Task.PROFILE_2D: [AnnotatorTypes.PROFILE2D_THRESHOLD],
```
wk9874: This is a very tiny nitpick - I wonder if itll be confusing if some things are referenced as `PROFILE_2D` and some as `PROFILE2D` ?

toktagger/api/core/views.py
```diff
@@ -19,30 +19,40 @@ def __call__(self, data: Data):
...
-            response = {}
-            for key, value in data.values.items():
-                response[key] = self.convert_timeseries_to_spectrogram(value)
+    def __call__(
+        self, data: MultiProfile2DData | MultiVariateTimeSeriesData
+    ) -> Profile2DData:
+        if self.params.signal_name not in data.values:
+            raise RuntimeError("Signal name not found in data")
+
+        profile_data = data.values[self.params.signal_name]
```
wk9874: Should this be `.get()` to be safe?

toktagger/api/core/views.py
```diff
@@ -19,30 +19,40 @@ def __call__(self, data: Data):
...
+
+        if profile_data is None:
+            raise RuntimeError(
+                f"Profile data for {self.params.signal_name} does not exist."
+            )
+
+        return self.convert_profile_to_view(profile_data)
+
+    def convert_profile_to_view(
```
wk9874, line 29: Would prefer this to be defined above where it is called

toktagger/api/core/views.py
```diff
@@ -19,30 +19,40 @@ def __call__(self, data: Data):
...
+        self, data: Profile2DData | TimeSeriesData
+    ) -> Profile2DData:
+        if isinstance(data, TimeSeriesData):
+            dim_1, time, values = compute_stft(data)
+            values = values.T  # Transpose to have shape (time, dim_1)
```
wk9874: I think you also transpose inside this function? Might want to make sure it isnt being done twice

toktagger/api/core/views.py
```diff
@@ -52,41 +62,39 @@ def convert_timeseries_to_spectrogram(
             self.params.time_max if self.params.time_max is not None else time.max()
```
wk9874, line 46: time_min = time_max?

toktagger/api/core/views.py
```diff
@@ -52,41 +62,39 @@ def convert_timeseries_to_spectrogram(
...
-            else values.max()
+        values_max = (
+            self.params.values_max
+            if self.params.values_max is not None
+            else np.nanmax(values)
         )
 
-        ds = xr.DataArray(values, coords=dict(frequency=freq, time=time))
+        ds = xr.DataArray(
+            values.T, coords=dict(dim_1=dim_1, time=time), dims=["dim_1", "time"]
```
wk9874: Suspicious of how many transposes are happening...

pyproject.toml
```diff
@@ -34,6 +34,11 @@ dependencies = [
   "mongita>=1.2.0",
   "filelock>=3.19.1",
   "hmmlearn>=0.3.3",
+  "shapely>=2.0.7",
+  "scikit-image>=0.24.0",
+  "torch>=2.8.0",
+  "ray>=2.51.1",
```
wk9874: Shouldn't be here

pyproject.toml
```diff
@@ -34,6 +34,11 @@ dependencies = [
   "mongita>=1.2.0",
   "filelock>=3.19.1",
   "hmmlearn>=0.3.3",
+  "shapely>=2.0.7",
+  "scikit-image>=0.24.0",
+  "torch>=2.8.0",
```
wk9874: Shouldn't be here

pyproject.toml
```diff
@@ -34,6 +34,11 @@ dependencies = [
...
+  "ray>=2.51.1",
+  "appdirs>=1.4.4",
```
wk9874: Why have we introduced this dependency?

toktagger/api/crud/db.py
```diff
@@ -1,10 +1,10 @@
-from pathlib import Path
 import pymongo
 import pydantic
 import typing
 from bson.objectid import ObjectId
+from pathlib import Path
+from appdirs import user_cache_dir
```
wk9874: Why change from platformdirs?

Either need to change this everywhere and remove the platformdirs dependency, or change this back to platformdirs and remove appdirs dependency

toktagger/api/schemas/annotations.py
```diff
@@ -72,6 +75,18 @@ class SpectrogramMask(AnnotationBase):
     values: list[list[float]]
 
 
+class PolygonAnnotation(AnnotationBase):
+    type: Literal["polygon"] = "polygon"
+    segmentation: list[list[float]]
+    area: float
+    bbox: list[float]  # [x, y, width, height]
```
wk9874: Confused by why this has segmentation and bbox?

Should try to unify this and VideoPolygon, in the same way BoundingBox and VideoBoundingBox are done

toktagger/api/schemas/views.py
```diff
@@ -1,26 +1,27 @@
...
 
 class ViewType(str, Enum):
     IDENTITY = "identity"
-    SPECTROGRAM = "spectrogram"
+    PROFILE_2D = "profile_2d"
 
 
-class ViewParams(ConfiguredModel):
-    name: ViewType = ViewType.IDENTITY
+class ViewParams(BaseModel):
```
wk9874, line 11: I think this is correct in inheriting from BaseModel instead of ConfiguredModel - could I be cheeky and ask you to also change DataParams to do the same? :)

wk9874 (changes requested):
Reviewed backend changes

samueljackson92:
I've merged `dev` with this branch and updated the implementation to use @jblake42's bounding box and polygon implementations. I think this is ready for another round.

abdullah-ukaea:
I have spent a bit of time functionally testing the new spectrogram view on this branch and I do not think this is ready for a code review yet, it is close though.

I have found some bugs / inconsistencies with other views I would like you to address. Please take a look at the below comments and we can discuss in the next meeting @samueljackson92

abdullah-ukaea:
**Problem 1 - View mode experience: enable scroll-wheel zoom**

In View mode, I expected spectrogram navigation to be consistent with the time-series and video views:
* Left-click and drag to pan.
* Use the mouse wheel to zoom in and out.

Panning currently works, but the mouse wheel does not zoom. To zoom, the user must switch from Pan to Zoom in the Plotly toolbar and then select a region, which makes navigation unnecessarily cumbersome.

Please enable both pan and mouse wheel zoom simultaneously in spectrogram view mode, consistent with the time-series and video views. If there is a technical or UX reason not to support this, could you explain it?

## Current Spectrogram view

https://github.com/user-attachments/assets/ec19e0b0-1cf5-4310-ac83-804f5a1411f8


## What time series does


https://github.com/user-attachments/assets/3f0f219d-98c6-4088-bf66-fc4ceaa1ddc7

abdullah-ukaea:
**Problem 2: Creating an annotation resets the zoom level**

In Edit mode, creating a bounding box or polygon unexpectedly resets the spectrogram’s zoom level.

Steps to reproduce:
1. Enter Edit mode and select the bounding-box or polygon tool.
2. Zoom into a specific region of the spectrogram.
3. Hold Ctrl and draw an annotation.
4. Complete the annotation.

As soon as the annotation is completed, the plot zooms back out. This is disruptive because users may need to remain closely zoomed in while annotating small or detailed events.

The viewport and zoom level should remain unchanged after an annotation is created. Please fix this behaviour, or explain if the zoom reset is intentional.

## Example with bounding box

https://github.com/user-attachments/assets/ff719434-26a3-4ee7-b266-306abc3f8a11

## Example with polygon and time region

https://github.com/user-attachments/assets/1466d652-e468-41a6-9a2b-eb4a0b91cfa0

abdullah-ukaea:
~~**Problem 3: Holding Ctrl unexpectedly changes the Plotly toolbar state**~~

~~When I hold Ctrl to create an annotation, the Pan button in the Plotly toolbar appears to toggle off and back on.~~

~~This suggests that Ctrl may also be triggering a Plotly interaction or keyboard shortcut. It may be related to the zoom-reset issue described above and could cause confusing or unintended behaviour while annotating.~~

## Zoomed in plotly toolbar

https://github.com/user-attachments/assets/2a3607b6-f54f-4057-831c-8c4ae12da763

## Whole spectogram view 

https://github.com/user-attachments/assets/284d9ff8-3142-44fc-9105-546d9e903817

**Ignore this comment for now as Josh mentioned**

abdullah-ukaea:
**Problem 4: Threshold controls need consistent alignment and spacing**

The expanded Threshold panel needs some UI styling. The controls currently have inconsistent alignment, widths, and spacing.

This makes the panel look unfinished and harder to scan. Does not look very aesthetic in general.

threshold_ui

whole_view_threshold_ui

abdullah-ukaea:
**Problem 5: Zoom controls do not support gradual zooming out**

This overlaps with some of the other issues such as problem 1 but I really do not like how zooming works here. It does not feel very natural and is hard to use.

From testing, the spectrogram zoom interactions appear to be:
* Click and drag left mouse button to select an area and zoom in.
* Double left click to autoscale.
* Right-click to reset to the fully zoomed-out view.

 What seems to be missing is a way to zoom out gradually. Users must either remain at the current zoom level or reset the entire view, making it difficult to adjust the viewport precisely.

Could we align this interaction with the time-series and video views, by supporting smooth mouse-wheel zooming in both directions while keeping the current viewport centred appropriately? 

## Spectrogram zoom

https://github.com/user-attachments/assets/74473936-0da5-4105-9824-9cd2ab14312f

## Time series zoom

https://github.com/user-attachments/assets/84a9e672-a0f9-48c5-9830-7653ea4e2238

## video zoom

https://github.com/user-attachments/assets/4b017e90-3a88-440a-b69c-a6ea008492cf

jblake42:
> **Problem 3: Holding Ctrl unexpectedly changes the Plotly toolbar state**
> 
> When I hold Ctrl to create an annotation, the Pan button in the Plotly toolbar appears to toggle off and back on.
> 
> This suggests that Ctrl may also be triggering a Plotly interaction or keyboard shortcut. It may be related to the zoom-reset issue described above and could cause confusing or unintended behaviour while annotating.
> ## Zoomed in plotly toolbar
> plotly_toolbar_when_I_press_Ctrl.mp4
> ## Whole spectogram view
> whole_spectogram_view_plotly_toolbar_when_I_press_Ctrl.mp4

This is currently expected behaviour required to prevent plotly handling pan events when drawing - time series does this too. This should not cause other issues.

There may be other ways of handling this that can be investigated but I do not think this needs to be looked at here unless it actually is causing issues

abdullah-ukaea:
Is this ready to be reviewed? @samueljackson92

samueljackson92:
Yes it is now the tests have passed. :)

toktagger/ui/src/app/core.tsx
```diff
@@ -189,10 +189,12 @@ export async function saveSampleAnnotations(
   if (!saveOnNavigate) {
     return;
   }
-  // Saving validates annotations without changing their creator metadata.
+  // Saving validates annotations and attributes them to the user, so annotator
+  // generated annotations become user-created once accepted.
   const updatedAnnotations = annotations.map((annotation: Annotation) => ({
     ...annotation,
     validated: true,
+    created_by: "manual",
```
abdullah-ukaea, line 197: Why have we added this to the 2D profile PR? After speaking with @wk9874 I deliberately reverted this in commit https://github.com/ukaea/toktagger/commit/9d6a5a13915e6b77dcb85a4de4844398721b6af8 

Saving should validate annotations without changing what created them. Setting created_by: "manual" here overwrites the provenance of model and threshold-generated annotations, so we can no longer distinguish them from manually created annotations. I think it is important to understand exactly what created an annotation a human or model.

Please retain the existing created_by value and only set validated: true in the save and afterSave paths.

toktagger/ui/src/app/core.tsx
```diff
@@ -204,6 +206,7 @@ export async function saveSampleAnnotations(
     body: JSON.stringify(updatedAnnotations),
   });
   if (!response.ok) {
+    console.error("Failed to save annotations:", await response.json());
     throw new Error(`Failed to save annotations: ${response.statusText}`);
```
abdullah-ukaea, line 210: Can you check if this is needed? 

This error is already caught by the save handler and displayed to the user in a toast, So the additional console.error seems redundant. It doesn't seem useful for a developer or a user of toktagger. If you disagree feel free to push back.

toktagger/ui/tsconfig.tsbuildinfo
abdullah-ukaea, line 1: As mentioned in other PRs we should not be committing this file and pushing to dev. Delete this file and please add a rule in your CLAUDE.md to not commit this file to the remote branch if it is running typescript build commands.

toktagger/ui/src/app/video/components/types.ts
```diff
@@ -67,7 +67,9 @@ export type VideoPolygon = {
   frame: number;
   track_id: string;
   label: string;
-  segmentation: number[];
+  segmentation: number[][];
+  area: number;
+  bbox: number[];
```
abdullah-ukaea, line 72: I am very confused why we are changing the video polygon schema in the UI side when we are not using this for the spectrogram view ..... was this a mistake? As we are using the time series d3 based polygon tooling not video.

This actually breaks the schema contract with the backend for the video polygons so saving no longer works. 

Please revert the video polygon changes in types.ts, anno-utils.ts, and overlay-sync-utils.ts to match dev. This is out of scope of this PR.

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
abdullah-ukaea, line 1: Please revert all changes to match dev, this PR should not be refactoring video code for the spectrogram view.

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea, line 1: Please revert all changes to match dev, this PR should not be refactoring video code for the spectrogram view.

toktagger/ui/src/app/video/components/video-view.tsx
```diff
@@ -112,7 +112,7 @@ function VideoProvidersInner({ children }: { children: React.ReactNode }) {
   const { videoPropagate, setVideoPropagate } = useVideoUiState();
 
   if (!project_id || !sample_id || !data) {
-    return <>{children}</>;
+    return null;
```
abdullah-ukaea, line 115: Bit surprised we are changing video code for spectrogram view.

This changes the video provider fallback to render nothing while data is unavailable, which can hide the loading UI and leave a blank view. This does not appear related to profile2d, so could we retain the existing dev behaviour and continue rendering children here?

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -220,113 +292,81 @@ export function SampleProvider({
...
 
-        setData(viewData);
+        setData(fetchedData);
 
-        // Video: remember last good frame so we can roll back on missing-frame errors.
-        if (projectData.task === TaskType.Video) {
-          const frame = (viewData as { frame?: unknown }).frame;
+        // video: remember last good frame so we can roll back on missing-frame errors
+        if (project.task === TaskType.Video) {
+          const frame = (fetchedData as unknown as { frame?: unknown }).frame;
```
abdullah-ukaea, line 369: Again very surprised we are changing video code in a spectrogram PR. This is out of scope and not justified please only change code related to spectrogram here and for video keep what is in dev as is.

Please preserve the existing video behaviour and add the profile2d handling without changing it. 

From what I can tell you correctly did the following:
- Add the profile2d data and view parameter types.
- Initialise Profile2DViewParams using the sample’s signal name.
- Parse profile2d responses.

To support that initialisation it looks like you split the existing combined loading effect into separate metadata and data effects, that refactor was not strictly necessary.... and I would argue unless you have a good reason perhaps revert it?

toktagger/ui/src/app/components/annotators/thresholding.tsx
```diff
@@ -34,110 +86,151 @@ export default function SpectrogramThresholdTool({
...
           (annotation: Annotation) =>
-            annotation.created_by !== AnnotatorTypes.SPECTROGRAM_THRESHOLD ||
-            annotation.validated,
+            !(
+              annotation.created_by === AnnotatorTypes.PROFILE2D_THRESHOLD &&
+              !annotation.validated &&
+              annotation.signal_name === signalName
+            ),
         );
         return otherAnnotations.concat(payload);
```
abdullah-ukaea, line 142: The current implementation of this use effect has quite a big edge case related to stale requests.

Each parameter change starts a new request without cancelling the previous one. A stale response can therefore replace the latest result or re-add threshold annotations after the tool has been disabled. The initial request can also run before the data-derived range is ready.

Please cancel or invalidate stale requests and only start once the range has been initialised. Also check response.ok and validate the payload with AnnotationsSchema before updating annotation state.

toktagger/ui/src/app/core.tsx
samueljackson92: This was how we were doing it before in time series interface. Annotators overwrite/remove annotations created by themselves. Probably the correct behaviour is as you suggest: use the validated annotation.

toktagger/ui/tsconfig.tsbuildinfo
samueljackson92: Should be in .gitignore then

toktagger/ui/src/app/components/plots/useAnnotationTooling.ts
```diff
@@ -0,0 +1,299 @@
...
+      return;
+    }
+
+    const suppressPlotlyDrag = (event: MouseEvent) => {
+      if (isDrawingRef.current) {
+        event.stopPropagation();
+      }
+    };
+
+    plot.addEventListener("mousedown", suppressPlotlyDrag, true);
```
abdullah-ukaea, line 96: Apologies, this is partly due to me originally raising Problem 3, but after Josh’s comment (https://github.com/ukaea/toktagger/pull/195#issuecomment-5091795933) we agreed to defer it. This capture phase handler suppresses Plotly’s mousedown interaction while Ctrl sets isDrawing, changing the shared behaviour used by both profile2d and time-series plots. We want to keep the behaviour as same as possible between all the views.

Please restore the previous drag-mode handling and remove test_profile2d_dragmode_stays_pan_while_drawing in the file https://github.com/ukaea/toktagger/blob/slj/profile2d/tests/end_to_end/test_profile2d_page.py

tests/end_to_end/test_profile2d_page.py
```diff
@@ -0,0 +1,354 @@
...
+    assert get_dragmode(page) == "pan"
+
+    # Holding Ctrl marks the app as "drawing"; the drag mode (and therefore the
+    # Plotly toolbar's active button) must stay put rather than flipping to false.
+    page.keyboard.down("Control")
+    page.wait_for_timeout(300)
+    dragmode_while_drawing = get_dragmode(page)
+    page.keyboard.up("Control")
+
+    assert dragmode_while_drawing == "pan"
```
abdullah-ukaea, line 354: we can remove this as mentioned in the comment in toktagger/ui/src/app/components/plots/useAnnotationTooling.ts but we have decided to not take action for problem 3 in this PR.

toktagger/ui/src/app/video/components/types.ts
samueljackson92: Actually I don't think it is out of scope. On the server side `Polygon` is a parent class to `VideoPolygon`, how we are storing segmentations should be consistent. 

Which way we go is another argument. Probably we should follow COCO https://cocodataset.org/#home

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -179,28 +167,19 @@ export default function ToolBar() {
...
-    const labels = project.shot_labels || ["Valid Shot", "Invalid Shot"];
-    tools.push({
-      name: "Shot Labels",
-      component: <ShotLabels labels={labels}></ShotLabels>,
-    });
+  const labels = project.shot_labels || ["Valid Shot", "Invalid Shot"];
+  tools.push({
+    name: "Shot Labels",
+    component: <ShotLabels labels={labels}></ShotLabels>,
+  });
```
abdullah-ukaea, line 174: There are now 2 shot labels for video view?

Shot Labels is now added for every task and then added again in the video branch. Please add this tool in one place only. If Shot Labels should be available for every task, keep the shared entry and remove the video-specific copy.

But perhaps before making this change read the other toolbar comments I will make as I noticed you changed a lot.

Image

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -35,93 +28,98 @@ import { JumpDetectionTool } from "../annotators/jump";
...
+          const NEXT_SAMPLE_URL = `/ui/projects/${project_id}/samples/${sample._id}`;
+          navigate(NEXT_SAMPLE_URL);
+        } else {
+          setErrorMessage("Shot not found!");
+        }
+      } catch (err) {
+        console.error("Failed to fetch data:", err);
+      }
+    } else {
+      setErrorMessage("Please enter a number.");
```
abdullah-ukaea, line 112: Why has this been duplicated? does this function not live already inside https://github.com/ukaea/toktagger/blob/dev/toktagger/ui/src/app/components/tools/nav.tsx

https://github.com/ukaea/toktagger/blob/900a566315ad32730347bdc67ba3d51f730a0c1e/toktagger/ui/src/app/components/tools/nav.tsx#L336

please delete.

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -35,93 +28,98 @@ import { JumpDetectionTool } from "../annotators/jump";
...
-  );
+async function getShotSample(project_id: string, shot_id: string) {
+  const NEXT_URL = `${BACKEND_API_URL}/projects/${project_id}/samples?shot_id=${shot_id}`;
+  const sampleResult = await fetch(NEXT_URL);
+  const sampleArray = await sampleResult.json();
+  let sample = null;
+  if (sampleArray.length > 0) {
+    sample = sampleArray[0];
+  }
+  return sample;
```
abdullah-ukaea, line 76: Why has this been duplicated? does this not already live inside https://github.com/ukaea/toktagger/blob/dev/toktagger/ui/src/app/core.tsx

https://github.com/ukaea/toktagger/blob/900a566315ad32730347bdc67ba3d51f730a0c1e/toktagger/ui/src/app/core.tsx#L147

please delete.

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -35,93 +28,98 @@ import { JumpDetectionTool } from "../annotators/jump";
...
+  annotations: Annotation[],
+) {
+  const ANNOTATIONS_URL = `${BACKEND_API_URL}/projects/${project_id}/samples/${sample_id}/annotations`;
+
+  const validatedAnnotations: Annotation[] = annotations.map(
+    (annotation: Annotation) => ({
+      ...annotation,
+      validated: true,
+      created_by: "manual",
+    }),
```
abdullah-ukaea, line 55: Very confused why this has been added, there is a different named existing function which is called updatedAnnotations which already does this. https://github.com/ukaea/toktagger/blob/900a566315ad32730347bdc67ba3d51f730a0c1e/toktagger/ui/src/app/core.tsx#L193

saveAnnotationsValidated does not exist on dev for this toolbar file .... saving and Shot Search already use the shared implementation in core.tsx and nav.tsx. This duplicate helper is unused ? and also changes created_by during validation  which as mentioned before valdiation and created by have different purposes, so please remove the toolbar local save.

toktagger/ui/src/app/components/tools/toolbar.tsx
```diff
@@ -179,28 +167,19 @@ export default function ToolBar() {
     defaultExpanded?: boolean;
   }[] = [];
 
-  if (data && project.task == TaskType.TimeSeries) {
-    const result = MultiVariateTimeSeriesDataSchema.safeParse(data);
-
-    if (!result.success) {
-      console.warn("Time series data is not available");
-      return;
-    }
```
abdullah-ukaea, line 188: why are we refactoring time series code in toolbar ?

It seems you are moving the existing time-series data validation out of the toolbar and changing the props of the time-series annotators. Is this refactor required for profile2d? 

What is the purpose of this? It seems out of scope no?

toktagger/ui/src/app/components/tools/toolbar.tsx
abdullah-ukaea, line 1: Given the amount of things that need to be changed for toolbar.

I honestly think that it might be easier for you to start with the toolbar.tsx script from dev branch, and then add in your specific profile 2d related changes again in. There is a lot of redundant functions, duplicates and then time series and video toolbar tools are being moved around also which seems weird.

Claude should be able to easily do this, just please tell it not do anything out of scope.

To recap:

This PR moves the existing time-series data validation out of the toolbar and changes the props of the time-series annotators. Is this refactor required for profile2d?
 
Please keep the existing time-series and video toolbar behaviour from dev and limit this change to adding the profile2d tools. Any broader toolbar refactor would be easier to review separately.

Then the functions  ShotSearch, getShotSample, and saveAnnotationsValidated need to be addressed. 

Hence path of least resistance is to copy the dev branch toolbar in, add in your profile2d related imports and tools then you are done.

toktagger/ui/src/app/profile2d/components/profile2d.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+  const heatmapData: Partial<Plotly.PlotData> = {
+    type: "heatmap",
+    x: viewData.time,
+    y: viewData.dim_1,
+    z: values,
+    // customdata: viewData.values,
+    hovertemplate:
+      "time: %{x:.2f}s<br>dim_1: %{y:.2f}Hz<br>values: %{customdata:.2e}<extra></extra>",
+    // @ts-expect-error Plotly.React types do not define shared color axis, but Plotly supports it.
```
abdullah-ukaea, line 215: Is this correct? 

The hover template references %{customdata}, but no customdata is supplied to the heatmap, so the measurement cannot be displayed correctly.

Please restore customdata: viewData.values to show the original measurement, particularly in log mode. Alternatively, use %{z} if displaying the transformed value is intentional.

toktagger/ui/src/app/profile2d/components/profile2d.tsx
```diff
@@ -0,0 +1,333 @@
...
+    ],
+    colorbar: {
+      ticks: "outside",
+      tickmode: "array",
+      tickfont: {
+        size: 10,
+      },
+    },
+  };
+  return { values: data.values, colorAxis };
```
abdullah-ukaea, line 63: This is bad typing.

These helpers use implicit any parameters and declare a Plotly trace return type even though they return { values, colorAxis }. The component also casts context data instead of validating it.

Please type the helper inputs and result correctly, and use the existing Zod schemas to validate data and viewParams rather than casting them.

toktagger/ui/src/app/contexts/SampleContext.tsx
abdullah-ukaea, line 1: I have a bit of a problem with this file and the code it is changing.

This PR needs to add profile2d data, view parameters, and parsing to the shared context, but it also restructures the existing loading lifecycle and changes video behaviour for some reason.
 
 Please keep SampleContext generic and preserve the existing time-series and video flows from dev. Any profile2d-specific UI state or broader context refactor should be kept in a profile2d provider or handled separately.

toktagger/ui/src/app/contexts/TimeSeriesContext.tsx
abdullah-ukaea, line 1: TimeSeriesProvider now stamps new annotations with signalName, but it still loads ever supported annotation from the sample without checking that signal. Switching profile2d signals can therefore display or edit annotations belonging to another signal.

Please manage only annotations matching the active signalName and preserve annotations for other signals unchanged. If this context is intended to support multiple plot types, its signal-scoping behaviour should be explicit.

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -42,7 +44,7 @@ interface SampleContextType {
     updater: (annotations: Annotation[]) => Annotation[] | Annotation[],
   ) => void;
   setDataParams: (params: DataParams) => void;
-  setViewParams: (params: ViewParams) => void;
+  setViewParams: (params: ViewParams | Profile2DViewParams) => void;
```
abdullah-ukaea, line 47: These context setters are typed as accepting values, but they are also used with React functional updates. viewParams can additionally be undefined because its state has no initial value, despite the context only declaring null.

Please type the setters as React.Dispatch<React.SetStateAction<T>> and initialise viewParams to null. Use the same setter pattern consistently for annotations and plot properties.

Also related  to the use state below

```tsx
const [viewParams, setViewParams] = useState<
  ViewParams | Profile2DViewParams | null
>(null);
```

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -220,113 +292,81 @@ export function SampleProvider({
...
-              });
-
-              // Roll back params; do NOT set error and do NOT clear data.
-              setDataParams((prev) => ({
-                ...prev,
-                name: "image",
-                frame: lastGood,
-              }));
+    refreshData();
+  }, [projectId, sampleId, dataParams]);
```
abdullah-ukaea, line 323: This effect only loads project, sample, and annotation metadata, but it still depends on dataParams. Every video frame or data-parameter change therefore refetches all three resources and can trigger another data request when the project and sample objects are replaced.
 
Please make this effect depend only on projectId and sampleId; the data effect should handle dataParams and viewParams. If the split effects are retained, stale requests should also be cancelled or ignored.

Before making changes make sure to read all the other sample context comments as you are changing video/timeseries context also here. You might decide to use what is in dev and just add in your small profile 2d changes only so this could be reverted.

toktagger/ui/src/app/components/tools/amplitudeSlider.tsx
abdullah-ukaea, line 1: This file is not used or imported anywhere? 

can it be deleted?

abdullah-ukaea:
can this file be deleted https://github.com/ukaea/toktagger/blob/slj/profile2d/toktagger/ui/src/app/spectrogram/components/spectrogram.tsx ?

now that https://github.com/ukaea/toktagger/blob/slj/profile2d/toktagger/ui/src/app/profile2d/components/profile2d.tsx exists?

If so delete this folder toktagger/ui/src/app/spectrogram/ probably with a git rm command

abdullah-ukaea:
Finished my first code review @samueljackson92 feel free to address my comments and push back on anything you disagree with

samueljackson92:
Thanks @abdullah-ukaea. I hopefully should have time to take a look later in the week.

abdullah-ukaea:
**Summary of code review 1 as I added a lot of comments**

Overall, this branch performs well in functional testing, but the first code review found a lot of things which need fixing.

My main concern is maintainability and scope.

Possibly because this is a old long-running branch with several merges from dev, it contains stale or duplicated code and changes to shared time-series, video, navigation, and save behaviour that are not required for profile2d.

Remember to double check the video view and time series view work / look as expected after making your changes before you request the second code review!

If the shared code is brought back in line with dev and the typing, schema, and request-handling issues are addressed, I think this will be ready to merge.

toktagger/ui/tsconfig.tsbuildinfo
abdullah-ukaea: I think it is https://github.com/ukaea/toktagger/blob/5df0c3dcca420728e0c49fe0d0b6529fd03a0d7b/.gitignore#L180

https://github.com/ukaea/toktagger/blob/654c06d957c2eef3977c2aba81f19e75785d961d/.gitignore#L181

services/ui/vite.config.js
abdullah-ukaea, line 1: delete this file, this folder does not exist on dev anymore.

abdullah-ukaea: please delete the entire services/ path with a git rm command.

toktagger/ui/src/app/video/components/types.ts
abdullah-ukaea: You are right that Polygon and VideoPolygon should consistently use COCO-style nested segmentation. 

I have implemented the complete frontend/backend change in the focused PR https://github.com/ukaea/toktagger/pull/341, so please revert all the video files in this PR to dev. Since that PR is tiny you can review it first, merge it into dev and then rebase dev into this branch. 

I strongly believe that area and bbox should not live in the video drawing code because they can be derived cheaply from segmentation in a backend/export layer, which I am happy to address separately if needed.

abdullah-ukaea:
Could you add a CLAUDE.md to the project root in this branch if possible @samueljackson92,  we need to add a project level CLAUDE.md for toktagger given the usage of claude

Reference: https://code.claude.com/docs/en/best-practices

toktagger/ui/src/app/video/components/types.ts
samueljackson92: As discussed, will fold in PR #341.

abdullah-ukaea:
This can be closed when PR https://github.com/ukaea/toktagger/pull/349 is merged

PR #199, opened 2026-03-04
Fix eslint checking

samueljackson92:
Fix checking of eslint errors in code base.


jblake42:
This still does not work - we do not have any eslint react modules by the look of it so we will not get dependency checking etc.

samueljackson92:
@jblake42 I clearly don't understand how eslint works. I have added plugins for react and it now spots missing hook deps.

jblake42 (approved):
This appears to have fixed the issues so happy for the merge

PR #206, opened 2026-03-06
Make ML Models Optional

wk9874:
Makes the ML model functionality with Ray, and the dependencies (`ray` and `torch`) optional, so that for users who only want to do labelling (or who are installing on a cluster with limited space), they dont need those huge packages


toktagger/api/models/base.py
```diff
@@ -7,13 +7,24 @@
...
+import uuid
 import logging
 
 logger = logging.getLogger("ray")
 
 
+def models_dependencies_installed() -> bool:
+    return (
+        importlib.util.find_spec("torch") is not None
+        and importlib.util.find_spec("ray") is not None
```
samueljackson92: I think we should think about this a bit more. I don't like the idea of us having to do this in every file for every dependancy.

tests/conftest.py
```diff
@@ -87,6 +87,7 @@ async def api_client(mongo_container):
     app = server.app
     lifespan_ctx = app.router.lifespan_context(app)
     await lifespan_ctx.__aenter__()
+    app.state.task_registry = TaskRegistry(max_actors=5)
```
samueljackson92: Why fixed to 5? Perhaps this should be an environment variable?

toktagger/api/models/base.py
```diff
@@ -190,3 +201,72 @@ def get(self, name):
         if not model_class:
             raise ValueError(f"No Model class called '{name}' found in registry!")
         return ray.remote(model_class)
+
+
+class TaskRegistry:
```
samueljackson92: This could be confusing with what we call tasks? Maybe `ActorRegistry`?

toktagger/api/models/base.py
samueljackson92: Perhaps we just check if `ray` is available, and assume that the other dependancies are install. We don't have control over custom models which could use whatever they like.

samueljackson92 (changes requested):

samueljackson92 (changes requested):
Testing locally, doing a `uv pip install -e ".[models]"` and running `uv run toktagger` gives me the message:

```
(raylet) Traceback (most recent call last):
(raylet)   File "/Users/rt2549/projects/viz-annotation/.venv/lib/python3.12/site-packages/ray/_private/workers/default_worker.py", line 8, in <module>
(raylet)     import ray
(raylet) ModuleNotFoundError: No module named 'ray'
```

After a doing a `uv sync`

wk9874:
> Testing locally, doing a `uv pip install -e ".[models]"` and running `uv run toktagger` gives me the message:
> 
> ```
> (raylet) Traceback (most recent call last):
> (raylet)   File "/Users/rt2549/projects/viz-annotation/.venv/lib/python3.12/site-packages/ray/_private/workers/default_worker.py", line 8, in <module>
> (raylet)     import ray
> (raylet) ModuleNotFoundError: No module named 'ray'
> ```
> 
> After a doing a `uv sync`

I actually hate `uv` so much, makes things way more complex than it needs to be

I suspect this will be because you're trying to install with uv in two ways? Your `uv pip install .[models]` installs ray and pytorch, but then `uv sync` uninstalls them again, and so `uv run toktagger` doesnt work. You either need to do:

1. `uv pip install .[models]`, then activate the env with `source .venv/bin/activate` and run `toktagger` from there
2. Do a `uv sync --all-extras`, then `uv run toktagger`
3. Do `uv run --all-extras toktagger`

Each of those commands work in slightly different ways, which is super annoying. But should**** all do equivalent stuff

samueljackson92:
Ah ok, gotcha. That works for me.

samueljackson92 (approved):

PR #208, opened 2026-03-09
Fix loading view for video frames

samueljackson92:


samueljackson92:
This fix requires a bit more thought to not upset video annotation.

Issue #174, opened 2026-02-18
Minimise view-specific code within the sample context


jblake42:
A not-inconsiderable amount of view-specific code has started to make its way into the sample context. I think this context should remain as view-agnostic as possible to make the addition of tooling easier.

We should look to pull out specifics into the view's code - we can do this by passing in the required params for fetch calls for example when the view initialises

Issue #155, opened 2026-01-20
Refactor Toolbar


jblake42:
The toolbar currently uses logic to populate the DOM based on the selected view which has led to a huge file - we should try and split this up more to make this clearer.

Additionally, tooling within the toolbar often requires access to contexts which are included lower in the DOM which causes challenges. We should look into whether we move the toolbar in the DOM or use tools such as portals to populate the tool bar from lower in the DOM

PR #210, opened 2026-03-10
Abdullah/cleanup state

wk9874:
This is a PR which mimics the PR which was merged so that Abdullah's fixes could make it into v0.1.0 while Josh was away on AL. We did some functionality testing to make sure it was all ok, with the plan being to review this code properly at a later date.

I've made this so that it is easier for Josh to see the changes as they were before merging and so can add his PR comments - Abdullah can then make the changes on this branch (or a separate branch, if easier), and merge back in to dev when happy

(PS: Make sure to change the base branch before merging in!)

### DIFF
--- a/toktagger/ui/src/app/components/tools/nav.tsx
+++ b/toktagger/ui/src/app/components/tools/nav.tsx
@@ -28,7 +28,7 @@ import { useSample } from "@/app/contexts/SampleContext";
 import { useSampleHistory } from "@/app/contexts/SampleHistoryContext";
 import { getNextSample } from "@/app/core";
 import type { SortDescriptor, SortDirection, Key } from "@react-types/shared";
-import { useNavAdapter } from "@/app/video/components/video-nav-adapter";
+import { useNavAdapter } from "@/app/contexts/NavAdapterContext";
 
 const TOAST_TIMEOUT = 5000;
 
--- a/toktagger/ui/src/app/components/tools/nav.tsx
+++ b/toktagger/ui/src/app/components/tools/nav.tsx
@@ -28,9 +28,7 @@ import { useSample } from "@/app/contexts/SampleContext";
 import { useSampleHistory } from "@/app/contexts/SampleHistoryContext";
 import { getNextSample } from "@/app/core";
 import type { SortDescriptor, SortDirection, Key } from "@react-types/shared";
-import {
-  useNavAdapterOptional,
-} from "@/app/video/components/video-nav-adapter";
+import { useNavAdapterOptional } from "@/app/video/components/video-nav-adapter";
 
 const TOAST_TIMEOUT = 5000;
 
--- a/toktagger/ui/src/app/components/tools/nav.tsx
+++ b/toktagger/ui/src/app/components/tools/nav.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { Annotation, Project } from "@/types";
+import { Annotation, Project, type NavAdapter } from "@/types";
 import {
   Flex,
   Button,
@@ -30,7 +30,6 @@ import { getNextSample } from "@/app/core";
 import type { SortDescriptor, SortDirection, Key } from "@react-types/shared";
 import {
   useNavAdapterOptional,
-  type NavAdapter,
 } from "@/app/video/components/video-nav-adapter";
 
 const TOAST_TIMEOUT = 5000;
--- a/toktagger/ui/src/app/components/tools/nav.tsx
+++ b/toktagger/ui/src/app/components/tools/nav.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { Annotation, Project, type NavAdapter } from "@/types";
+import { Project, type NavAdapter } from "@/types";
 import {
   Flex,
   Button,
@@ -28,7 +28,7 @@ import { useSample } from "@/app/contexts/SampleContext";
 import { useSampleHistory } from "@/app/contexts/SampleHistoryContext";
 import { getNextSample } from "@/app/core";
 import type { SortDescriptor, SortDirection, Key } from "@react-types/shared";
-import { useNavAdapterOptional } from "@/app/video/components/video-nav-adapter";
+import { useNavAdapter } from "@/app/video/components/video-nav-adapter";
 
 const TOAST_TIMEOUT = 5000;
 
@@ -80,10 +80,9 @@ async function navigateToNextSample(
 type ButtonInfo = {
   project_id: string;
   sample_id: string;
-  annotations: Annotation[];
   setIsValidated: (validated: boolean) => void;
   saveOnNavigate?: boolean;
-  navAdapter?: NavAdapter | null;
+  navAdapter: NavAdapter;
 };
 
 type NextButtonInfo = ButtonInfo & {
@@ -100,7 +99,6 @@ type PreviousButtonInfo = ButtonInfo & {
 function NextButton({
   project_id,
   sample_id,
-  annotations,
   setIsValidated,
   visitedSampleIds,
   sortDescriptor,
@@ -110,9 +108,7 @@ function NextButton({
   const navigate = useNavigate();
 
   const moveNextShot = useCallback(async () => {
-    const annotationsToSave = navAdapter
-      ? navAdapter.getAnnotations()
-      : annotations;
+    const annotationsToSave = navAdapter.getAnnotations();
     await saveSampleAnnotations(
       project_id,
       sample_id,
@@ -120,7 +116,7 @@ function NextButton({
       saveOnNavigate,
     );
     if (saveOnNavigate) {
-      navAdapter?.afterSave?.();
+      navAdapter.afterSave?.();
       setIsValidated(true);
     }
     await navigateToNextSample(
@@ -132,7 +128,6 @@ function NextButton({
   }, [
     project_id,
     sample_id,
-    annotations,
     navigate,
     saveOnNavigate,
     setIsValidated,
@@ -152,7 +147,7 @@ function NextButton({
 
     window.addEventListener("keydown", handleKeyDown);
     return () => window.removeEventListener("keydown", handleKeyDown);
-  }, [annotations, sample_id, project_id, navigate, moveNextShot]);
+  }, [sample_id, project_id, navigate, moveNextShot]);
 
   return (
     <View marginStart="size-100">
@@ -194,7 +189,6 @@ export function JumpToNextButton({
 function PreviousButton({
   project_id,
   sample_id,
-  annotations,
   setIsValidated,
   isDisabled,
   popVisitedSampleId,
@@ -205,9 +199,7 @@ function PreviousButton({
   const navigate = useNavigate();
 
   const movePreviousShot = useCallback(async () => {
-    const annotationsToSave = navAdapter
-      ? navAdapter.getAnnotations()
-      : annotations;
+    const annotationsToSave = navAdapter.getAnnotations();
     await saveSampleAnnotations(
       project_id,
       sample_id,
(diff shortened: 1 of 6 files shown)


toktagger/ui/src/app/contexts/SampleContext.tsx
jblake42, line 1: I have created an issue (#174) as I would to remove as much of this view-specific logic as possible. The rationale behind this is that it is hard to track where certain things are handled - this context should be view-agnostic. You don't need to do anything here as other views are also doing the same thing

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -50,29 +48,28 @@ const SampleDataBreadCrumbs = ({
...
-    setDataParams,
-    isLoading,
-    error,
-  } = useSample();
+function SampleTaskProviders({ children }: { children: React.ReactNode }) {
+  const { project } = useSample();
 
-  const isVideo = project?.task === TaskType.Video;
+  if (project?.task === TaskType.Video) {
+    return <VideoProviders>{children}</VideoProviders>;
```
jblake42, line 65: This is a good solution to the toolbar issue - I think I still would prefer the toolbar to be refactored at somepoint to remove the need for the provider to be so high in the DOM, this way the provider instantiation can be contained within the view-specific components (#155)

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -111,34 +108,10 @@ function SamplePageContent(props: { projectId: string; sampleId: string }) {
         <ModelTrainModal project={project} />
         <ModelPredictModal project={project} />
         <Flex>
-          {isVideo ? (
```
jblake42: Good!

toktagger/ui/src/app/video/components/ui_elements.tsx
```diff
@@ -26,8 +26,8 @@ export function ClassPanel({
   setSelectedClassName: (v: string | null) => void;
 }) {
   return (
-    <div className="rounded-xl border border-gray-700 bg-black shadow-sm p-3 w-48 mx-auto">
-      <Text UNSAFE_className="text-sm font-medium mb-2 block text-white">
+    <div className="mx-auto w-48 rounded-xl border border-gray-200 bg-white p-3 text-gray-900 shadow-sm dark:border-gray-700 dark:bg-gray-950 dark:text-gray-100">
```
jblake42: I think we should be trying to use spectrum components here as they will keep styling consistent especially when changing view-modes. A `View` would be suitable here I think

toktagger/ui/src/app/video/components/ui_elements.tsx
```diff
@@ -26,8 +26,8 @@ export function ClassPanel({
...
+      <Text UNSAFE_className="mb-2 block text-sm font-medium text-gray-700 dark:text-gray-200">
```
jblake42: Do we 100% need to use unsafe styling - spectrum recommends against this

toktagger/ui/src/app/video/components/ui_elements.tsx
```diff
@@ -111,7 +111,7 @@ export function InstancePanel({
                 return next;
               });
             }}
-            className="w-full rounded-lg border shadow-sm px-2.5 py-1.5 text-left bg-black text-white border-gray-600 hover:bg-gray-800 flex items-center justify-between"
+            className="flex w-full items-center justify-between rounded-lg border border-gray-200 bg-white px-2.5 py-1.5 text-left text-gray-900 shadow-sm hover:bg-gray-50 disabled:cursor-not-allowed disabled:opacity-60 dark:border-gray-700 dark:bg-gray-950 dark:text-gray-100 dark:hover:bg-gray-900"
```
jblake42: Again, I think this component could make use of Spectrum a little more

toktagger/ui/src/app/video/components/ui_elements.tsx
jblake42, line 1: I have made comments in here about using spectrum - as this is styling I am happy that it is not blocking and an issue can be created. Since we have made the decision to use spectrum we should aim to stick with it as far a possible to maintain a consistent look and feel throughout and to handle light/dark mode effectively. Spectrum also enforces a number of accessability features for free

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
```diff
@@ -0,0 +1,61 @@
+"use client";
+
+import React, { createContext, useContext } from "react";
+import { useSample } from "@/app/contexts/SampleContext";
+import type { Annotation } from "@/types";
+import { useVideoSession } from "./video-session";
+
+export type NavAdapter = {
```
jblake42: I don't think this is the place for this type definition

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
jblake42, line 1: I personally do not see the benefit of this being a provider - all you do is instantiate it and then pass down functions. I think this should just be treated as a normal adapter. Please let me know if there is a specific reason?

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const session = useVideoSession();
+  const { annotations } = useSample();
+
+  const adapter: NavAdapter = {
+    getAnnotations: () => {
+      const shotLabels = (annotations ?? []).filter(
+        (annotation): annotation is Annotation =>
+          annotation.type === "class_label",
+      );
+      const videoBoxes = session.collectAllVideoBBoxes() as Annotation[];
```
jblake42: This does not seem right to me - you are performing the cast but the types do not seem to be the correct type overlap

toktagger/ui/src/app/video/components/video-toolbox.tsx
```diff
@@ -236,18 +236,20 @@ export function VideoToolbox() {
...
+          </div>
           <VideoClassPanel
             items={classItems}
             selectedClassName={session.selection.className}
             setSelectedClassName={onSelectClassName}
           />
         </div>
 
-        <div className="border-t border-gray-800 mx-4" />
+        <div className="mx-4 border-t border-gray-200 dark:border-gray-800" />
```
jblake42: Can we use spectrum dividers?

toktagger/ui/src/app/components/tools/nav.tsx
jblake42, line 1: I am tempted to say that if a nav adapter is not passed we should create a default one - that way we only need to pass the adapter and don't have to constantly check if the adapter exists. It also means we are not passing the annotation object around when it is not needed

jblake42 (changes requested):
This looks good - nothing major is my comments.

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: Agreed, done in commit https://github.com/ukaea/toktagger/pull/210/changes/a57b797dad5e77a24cdff77b454696f99e825909

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: Agreed unsafe should be used as a last resort, also fixed in commit https://github.com/ukaea/toktagger/commit/a57b797dad5e77a24cdff77b454696f99e825909

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
abdullah-ukaea: agreed, move to ui types.ts file in commit https://github.com/ukaea/toktagger/pull/210/changes/6fd97a1e0503c8c74a2b937787b5b3618d708fc9

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: Agreed, done in commit https://github.com/ukaea/toktagger/pull/210/changes/c3ef1a27b15b806468012da48f25f83ff41e278e

toktagger/ui/src/app/video/components/video-toolbox.tsx
abdullah-ukaea: sure, done in commit https://github.com/ukaea/toktagger/pull/210/commits/345caf2a7a231d44e0dcf63f818bd3fba60ff293

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
abdullah-ukaea: As discussed we will leave this current approach with the provider and after the time series refactor we may revisit this and clean this up.

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
abdullah-ukaea: Does this commit perform the correct cast? https://github.com/ukaea/toktagger/pull/210/commits/e47707294ddb18339ecf0ee5a5f3f61bc45255b9 but yep I agree that the line
`const videoBoxes = session.collectAllVideoBBoxes() as Annotation[];`

seems to be doing a unsafe cast so tried to fix it now.

toktagger/ui/src/app/components/tools/nav.tsx
abdullah-ukaea: what are your thoughts on the commit https://github.com/ukaea/toktagger/pull/210/commits/526c7d3a5135816fa6983a29e0dc34a63c583e39 I can revert it if this is not what you wanted.

toktagger/ui/src/app/components/tools/nav.tsx
```diff
@@ -28,10 +28,7 @@ import { useSample } from "@/app/contexts/SampleContext";
 import { useSampleHistory } from "@/app/contexts/SampleHistoryContext";
 import { getNextSample } from "@/app/core";
 import type { SortDescriptor, SortDirection, Key } from "@react-types/shared";
-import {
-  useNavAdapterOptional,
-  type NavAdapter,
-} from "@/app/video/components/video-nav-adapter";
+import { useNavAdapter } from "@/app/video/components/video-nav-adapter";
```
jblake42: This hook should not be coming from the video component code - this should be generic as the functionality it provides is generic

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
```diff
@@ -31,6 +29,22 @@ export function useNavAdapterOptional(): NavAdapter | null {
   return useContext(NavAdapterContext);
 }
 
+export function useNavAdapter(): NavAdapter {
```
jblake42: See previous comment regarding location of this code

jblake42 (changes requested):
Yeah I like this approve a lot more than the optional adapters in the code

toktagger/ui/src/app/video/components/video-nav-adapter.tsx
abdullah-ukaea: does this commit https://github.com/ukaea/toktagger/pull/210/changes/80f44626db818296e00a669c5cb2a076be2ead41 do the change you requested?

toktagger/ui/src/app/components/tools/nav.tsx
abdullah-ukaea: does this commit https://github.com/ukaea/toktagger/pull/210/changes/80f44626db818296e00a669c5cb2a076be2ead41 do the change you requested?

abdullah-ukaea:
Ready to merge when you are happy with this @jblake42

jblake42:
LGTM

PR #211, opened 2026-03-20
Add dataloader worker registry and refactor dataloaders

wk9874:
* Adds a dataloader worker registry, so that models have access to user extended data loaders
* Refactors dataloaders to accept params inside `get_sample`, instead of in `__init__`
* Update docs
* Fix project sorting test (hangover from previous merge)

### DIFF
--- a/docs/custom_dataloaders.md
+++ b/docs/custom_dataloaders.md
@@ -206,7 +206,7 @@ class CSVTimeSeriesLoader(DataLoader):
     def get_sample(
         self, 
         sample: Sample,
-        params: DataParams,
+        params: DataParams = DataParams(),
     ) -> MultiVariateTimeSeriesData:
         """
         Load time series data from a CSV file.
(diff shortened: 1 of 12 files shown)


docs/custom_dataloaders.md
```diff
@@ -205,8 +205,8 @@ class CSVTimeSeriesLoader(DataLoader):
     @pydantic.validate_call
     def get_sample(
         self, 
-        shot_id: int, 
-        sample_data: TimeSeriesFileData
+        sample: Sample,
+        params: DataParams,
```
samueljackson92: Maybe we should have the "identity" params be the default args, to prevent us copying that around everywhere.

samueljackson92 (commented):
Looks good to me, but one small comment.

samueljackson92 (changes requested):

docs/custom_dataloaders.md
wk9874: Made it a default in the base class, and its already a default in the /data endpoint

Issue #140, opened 2025-12-18
Image Annotation: Add a `propagate annotations` toggle


wk9874:
If I dont care about tracking UFOs across multiple frames, or there are very few frames in my sample where there are consecutive UFOs across multiple frames, it is annoying to have to click each box from the previous frame and delete them on the next frame...

It'd be good to have a 'propagate boxes' toggle - If you had a toggle that just meant the boxes don't propagate over frames, then users could still do 'tracking' if they wanted by clicking on the instance in the list and drawing a new box on the next frame... it would just be quicker to use if you dont have many consecutive UFO events rather than having to right click -> delete each box.

Keep it in 'tracking' mode, ie keep it storing instance labels and stuff, just make the toggle enable / disable the boxes moving across frames automatically

PR #216, opened 2026-03-31
Video Annotation: Polygon tooling & Propagate annotations toggle

abdullah-ukaea:
This PR addresses issues:
- https://github.com/ukaea/toktagger/issues/140
- Adds polygon tooling as a frame tool option

image

### DIFF
--- a/toktagger/ui/src/app/video/components/anno-utils.ts
+++ b/toktagger/ui/src/app/video/components/anno-utils.ts
@@ -8,6 +8,8 @@ import {
   type PolygonGeometry,
   type Rectangle,
   type RectangleGeometry,
+  type ImageAnnotationTarget,
+  type Shape,
 } from "@annotorious/react";
 import type {
   VideoAnnotationShape,
@@ -30,22 +32,22 @@ type UnknownRecord = Record<string, unknown>;
 
 // Our app stores a frame key on target.source (not present in upstream Annotorious types).
 export type VideoImageAnnotation = ImageAnnotation & {
-  target: ImageAnnotation["target"] & {
+  target: ImageAnnotationTarget & {
     source?: string;
   };
 };
 
-export type RectangleAnnotation = ImageAnnotation & {
-  target: ImageAnnotation["target"] & {
-    selector: Rectangle;
+export type ImageAnnotationWithSelector<TSelector extends Shape> = Omit<
+  ImageAnnotation,
+  "target"
+> & {
+  target: Omit<ImageAnnotationTarget, "selector"> & {
+    selector: TSelector;
   };
 };
 
-export type PolygonAnnotation = ImageAnnotation & {
-  target: ImageAnnotation["target"] & {
-    selector: Polygon;
-  };
-};
+export type RectangleAnnotation = ImageAnnotationWithSelector<Rectangle>;
+export type PolygonAnnotation = ImageAnnotationWithSelector<Polygon>;
 
 function withTargetSource(
   a: ImageAnnotation,
--- a/toktagger/ui/src/app/video/components/anno-utils.ts
+++ b/toktagger/ui/src/app/video/components/anno-utils.ts
@@ -4,6 +4,10 @@ import {
   ShapeType,
   type AnnotationBody,
   type ImageAnnotation,
+  type Polygon,
+  type PolygonGeometry,
+  type Rectangle,
+  type RectangleGeometry,
 } from "@annotorious/react";
 import type {
   VideoAnnotationShape,
@@ -28,6 +32,18 @@ type UnknownRecord = Record<string, unknown>;
 export type VideoImageAnnotation = ImageAnnotation & {
   target: ImageAnnotation["target"] & {
     source?: string;
+  };
+};
+
+export type RectangleAnnotation = ImageAnnotation & {
+  target: ImageAnnotation["target"] & {
+    selector: Rectangle;
+  };
+};
+
+export type PolygonAnnotation = ImageAnnotation & {
+  target: ImageAnnotation["target"] & {
+    selector: Polygon;
   };
 };
 
@@ -141,14 +157,14 @@ function newBodyId(): string {
 }
 
 /** True if the annotation target is a rectangle selector. */
-export function isRectangleAnno(a: ImageAnnotation): boolean {
+export function isRectangleAnno(a: ImageAnnotation): a is RectangleAnnotation {
   // ShapeType.RECTANGLE is already the string literal "RECTANGLE",
   // so we don't need a separate check for the string form.
   return a.target.selector.type === ShapeType.RECTANGLE;
 }
 
 /** True if the annotation target is a polygon selector. */
-export function isPolygonAnno(a: ImageAnnotation): boolean {
+export function isPolygonAnno(a: ImageAnnotation): a is PolygonAnnotation {
   return a.target.selector.type === ShapeType.POLYGON;
 }
 
@@ -156,34 +172,12 @@ function isFiniteNumber(v: unknown): v is number {
   return Number.isFinite(v);
 }
 
-/** Rectangle geometry shape we rely on at runtime. */
-type RectGeometry = {
-  x: number;
-  y: number;
-  w: number;
-  h: number;
-};
-
-type PolygonGeometry = {
-  points: [number, number][];
-};
-
 /** Read rectangle geometry (returns null if missing/invalid). */
 export function readRectGeometry(
-  a: ImageAnnotation,
-): { x: number; y: number; w: number; h: number } | null {
-  // Make this function self-contained: only rectangles have x/y/w/h.
-  if (!isRectangleAnno(a)) return null;
-
+  a: RectangleAnnotation,
+): RectangleGeometry | null {
   const g = a.target.selector.geometry;
-  if (!g || typeof g !== "object") return null;
-
-  const rg = g as Partial<RectGeometry>;
-
-  const x = rg.x;
-  const y = rg.y;
-  const w = rg.w;
-  const h = rg.h;
+  const { x, y, w, h } = g;
 
   // TS now narrows x/y/w/h to number after these checks.
   if (
@@ -196,33 +190,25 @@ export function readRectGeometry(
 
   if (w <= 0 || h <= 0) return null;
 
-  return { x, y, w, h };
+  return g;
 }
 
 /** Read polygon geometry (returns null if missing/invalid). */
 export function readPolygonGeometry(
-  a: ImageAnnotation,
-): { points: [number, number][] } | null {
-  if (!isPolygonAnno(a)) return null;
-
+  a: PolygonAnnotation,
+): PolygonGeometry | null {
   const g = a.target.selector.geometry;
-  if (!g || typeof g !== "object") return null;
-
-  const pg = g as Partial<PolygonGeometry>;
-  const points = pg.points;
+  const { points } = g;
   if (!Array.isArray(points) || points.length < 3) return null;
 
-  const out: [number, number][] = [];
   for (const point of points) {
     if (!Array.isArray(point) || point.length !== 2) return null;
 
     const [x, y] = point;
     if (!isFiniteNumber(x) || !isFiniteNumber(y)) return null;
-
-    out.push([x, y]);
   }
 
-  return { points: out };
+  return g;
 }
 
 /**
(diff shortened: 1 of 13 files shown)


abdullah-ukaea:
ready for review @jblake42

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -143,6 +147,11 @@ export function isRectangleAnno(a: ImageAnnotation): boolean {
   return a.target.selector.type === ShapeType.RECTANGLE;
 }
 
+/** True if the annotation target is a polygon selector. */
+export function isPolygonAnno(a: ImageAnnotation): boolean {
```
jblake42: I think it would be nice for these functions to be proper type guards. That way you can set the type of the annotation to Rectangle/Polygon and then this can be passed around. This way you do not need to keep repeating type checks - see other comments in this file

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -186,9 +199,35 @@ export function readRectGeometry(
   return { x, y, w, h };
 }
 
+/** Read polygon geometry (returns null if missing/invalid). */
+export function readPolygonGeometry(
+  a: ImageAnnotation,
```
jblake42: This would not be an `ImageAnnotation` it would be a `PolygonAnnotation` and as such you would not need the `isPolygonAnno` check

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -155,6 +164,10 @@ type RectGeometry = {
   h: number;
 };
 
+type PolygonGeometry = {
```
jblake42: I don't think we need this new type - annotorious exposes a polygon geometry type

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -186,9 +199,35 @@ export function readRectGeometry(
...
+/** Read polygon geometry (returns null if missing/invalid). */
+export function readPolygonGeometry(
+  a: ImageAnnotation,
+): { points: [number, number][] } | null {
+  if (!isPolygonAnno(a)) return null;
+
+  const g = a.target.selector.geometry;
+  if (!g || typeof g !== "object") return null;
+
+  const pg = g as Partial<PolygonGeometry>;
```
jblake42: We also wouldn't need this type assertion

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -275,6 +314,46 @@ export function annoToVideoBBox(
   };
 }
 
+/** Convert a normalized ImageAnnotation -> backend VideoPolygon. */
+export function annoToVideoPolygon(
+  a: ImageAnnotation,
```
jblake42: Again, it would nice if this param had a polygon type

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -275,6 +314,46 @@ export function annoToVideoBBox(
...
+    created_by: "manual",
+  };
+}
+
+/** Convert a normalized ImageAnnotation -> backend video annotation shape. */
+export function annoToVideoAnnotation(
+  a: ImageAnnotation,
+  frame: number,
+): VideoAnnotationShape | null {
+  if (isRectangleAnno(a)) return annoToVideoBBox(a, frame);
```
jblake42, line 338: This is where the actual type guarding happens. If the function is defined as a type guard it means `a` can be passed into each of the functions with an explicit type

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -318,9 +397,62 @@ export function videoBBoxToAnno(
...
+    if (x < minX) minX = x;
+    if (y < minY) minY = y;
+    if (x > maxX) maxX = x;
+    if (y > maxY) maxY = y;
+  }
+
+  const geometry = {
+    points,
+    bounds: { minX, minY, maxX, maxY },
+  } as ImageAnnotation["target"]["selector"]["geometry"];
```
jblake42: We should just use the types annotorious exposes directly here and in the lines below

toktagger/ui/src/app/video/components/video-toolbox.tsx
jblake42, line 1: This is very non-spectrum - can we align more closely to spectrum at all? For example spectrum support grids rather than using tailwind in a `div`

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -110,14 +120,41 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
+  const formatDetails = (annotation: ImageAnnotation) => {
+    const rect = readRectGeometry(annotation);
+    if (rect) {
+      return `x=${Math.round(rect.x)}, y=${Math.round(rect.y)}, w=${Math.round(rect.w)}, h=${Math.round(rect.h)}`;
+    }
+
+    const polygon = readPolygonGeometry(annotation);
+    if (!polygon) return null;
+
+    let minX = Infinity;
```
jblake42: We've done this calculation elsewhere - is there no way to use the bounds values to store this once and just reference that?

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
```diff
@@ -101,6 +108,102 @@ function withRectGeometry(
...
+  if (polygonArea(deduped) <= 0) return null;
+
+  return deduped;
+}
+
+function withPolygonGeometry(
+  a: ImageAnnotation,
+  points: [number, number][],
+): ImageAnnotation {
+  let minX = Infinity;
```
jblake42: Yeah we're doing it again here - I definitely think we should use the bounds property if possible

jblake42 (changes requested):
Looking good - behaviour seems correct to me and it saves to the backend. Just some code tidiness suggestions

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/216/commits/9d57f30e2a534d0ed085b0f9dc0ac60015b75ec3

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/216/commits/9d57f30e2a534d0ed085b0f9dc0ac60015b75ec3

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/216/commits/9d57f30e2a534d0ed085b0f9dc0ac60015b75ec3

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/216/commits/9d57f30e2a534d0ed085b0f9dc0ac60015b75ec3

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/216/commits/9d57f30e2a534d0ed085b0f9dc0ac60015b75ec3

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/216/commits/9d57f30e2a534d0ed085b0f9dc0ac60015b75ec3

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/216/commits/9d57f30e2a534d0ed085b0f9dc0ac60015b75ec3

toktagger/ui/src/app/video/components/video-toolbox.tsx
abdullah-ukaea: fixed the styling in commit https://github.com/ukaea/toktagger/pull/216/commits/9ea5a05037401f83b349617cde280fc0cf0c24cc

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: fixed this in commit https://github.com/ukaea/toktagger/pull/216/commits/192c1e3681f82bbbfbe36d2fd4c0a9b185a5cb3b

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
abdullah-ukaea: fixed this in commit https://github.com/ukaea/toktagger/pull/216/commits/192c1e3681f82bbbfbe36d2fd4c0a9b185a5cb3b

abdullah-ukaea:
@jblake42 I have updated the branch based on your review. Could you take a look at the new commits and let me know if any more changes are required. If you are happy and no changes are needed you can merge!

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
jblake42, line 81: I think this function could be cleaned up a significant amount - specifically the input should be `RectangleAnnotation`. This should make all the manipulation of the internal data far easier. This is now possible because of the proper type guarding that has been introduced when this funciton is called.

jblake42: You should also be able to just return a `RectangleAnnotation` as they types are compatible

jblake42: I think a lot of this will not be necessary with the changes recommended above, however if it is still needed we should use the defined types provided by the API - similar to my comments in the last round of reviews

jblake42, line 94: This should also be able to be simplified as you should just be able to alter the object directly and return that ... I think

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
```diff
@@ -134,7 +138,7 @@ function polygonArea(points: [number, number][]): number {
 }
 
 function clampPolygonToImage(
```
jblake42, line 134: Same comments here about this function

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -26,11 +32,23 @@ type UnknownRecord = Record<string, unknown>;
 
 // Our app stores a frame key on target.source (not present in upstream Annotorious types).
 export type VideoImageAnnotation = ImageAnnotation & {
-  target: ImageAnnotation["target"] & {
+  target: ImageAnnotationTarget & {
     source?: string;
   };
 };
 
+export type ImageAnnotationWithSelector<TSelector extends Shape> = Omit<
```
jblake42, line 40: I have cleaned up the typing and templated it - with the previous implementation it was a little messy as the other definition of `target` and `selector` were not omitted meaning there was an unwanted type union. Let me know if you are unhappy with this

jblake42 (changes requested):

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
jblake42: withPolygonGeometry

toktagger/ui/src/app/video/components/overlay-sync-utils.ts
abdullah-ukaea: I think I have done this in commit https://github.com/ukaea/toktagger/pull/216/changes/05f35b5b052be09bf759264b906b707c75c45610

abdullah-ukaea: yep done in commit https://github.com/ukaea/toktagger/pull/216/changes/05f35b5b052be09bf759264b906b707c75c45610

abdullah-ukaea: removed the double cast in commit https://github.com/ukaea/toktagger/pull/216/changes/05f35b5b052be09bf759264b906b707c75c45610

abdullah-ukaea: I don’t think we should change the return here, because the video annotation flow is built around immutable updates. Mutating a.target.selector.geometry directly could bypass setByFrame and cause missed updates or stale state.

abdullah-ukaea:
Can you check my latest commit to see if it fixes the typing issues you mentioned? if it does you can merge @jblake42

jblake42 (approved):
LGTM

Issue #141, opened 2025-12-18
Image Annotation: Add zoom functionality


wk9874:
Would be really useful to be able to zoom into the image so that you can accurately draw boxes around small UFOs

abdullah-ukaea:
closed with https://github.com/ukaea/toktagger/pull/221

Issue #139, opened 2025-12-18
Image Annotation: Add keyboard navigation


wk9874:
Would be good to support keyboard navigation for ease of use, such as:

* +/- keys for moving to next/previous frames
* Delete key to delete an annotation once selected
* Arrow keys to move a bounding box once selected

PR #221, opened 2026-04-09
Video Annotation: Add Zoom functionality

abdullah-ukaea:
### This PR addresses issues:
- https://github.com/ukaea/toktagger/issues/141
- https://github.com/ukaea/toktagger/issues/139  (only partially for zoom related keyboard shortcuts)

### Interaction States

- Default state on load/refresh: Rectangle tool is active.

- Drawing states (Rectangle / Polygon):
   - Users can draw new annotations.
   - Users can select, move, and resize existing annotations.
   
- Drag/Zoom state (via toolbox button):
   - Drawing is disabled.
   - Users can pan the frame and zoom with the mouse wheel.
   
- Temporary Drag/Zoom (via keyboard):
   - Holding Shift temporarily enters Drag/Zoom.
   - Releasing Shift returns to the previous drawing state/tool (Rectangle or Polygon), so annotation flow is not interrupted.

[Screencast from 09-04-26 16:06:24.webm](https://github.com/user-attachments/assets/f2865202-8bc2-4be6-97a0-f508385876ec)


abdullah-ukaea:
Hi @jblake42 @wk9874 This draft is still a work in progress but if possible could you both give this branch a try and tell me if you have any feedback in terms of the UI, functionality and things you want me to add/remove ! 

A keyboard shortcut (e.g. shift or ctrl) for temporarily entering the zoom mode is in the works as Josh mentioned that would be useful and I was thinking of adding a button on the frame view that lets you got to default zoom. 

Let me know if you spot any bugs!

Please don't review the code yet, focus on the functionality for now

abdullah-ukaea:
Commit https://github.com/ukaea/toktagger/pull/221/commits/c7740a10ab31ba0dcfea0fc2bbf6510909a5164d now has added a keyboard shortcut !

abdullah-ukaea:
Commit https://github.com/ukaea/toktagger/pull/221/changes/729acd0dfea6162501c170c4a36cd584e9ff28f6 added a reset zoom button 
image
image

abdullah-ukaea:
Hi @jblake42 This branch is ready for review, the functionality seems to be good for now

jblake42:
I have not looked at the code yet - this is just a comment based on user testing. I think the functionality works great, it seems very robust. I have a couple of comments though:
1. From using this tool vs the tool provided by the openseadragon demo I think I am personally leaning towards the demo version. In the demo version all annotations are click to draw, there is no dragging. This means that dragging can be reserved for panning at all times and zooming is also functional at all times. Even from my small amount of usage I often found needing zoom to be slightly annoying. This could just be a personal thing and other users will prefer the drag functionality, which requires the shift-to-pan
2. This is definitely an issue for a seperate PR as it is not blocking and not to do with zoom, but when drawing a polygon you cannot place a point on another annotation. This may or may not be an issue when annotating, but I think users should be able to have the option to do it. On a similar note, you cannot start an annotation on top of another annotation - maybe this is a required use-case? If it is, we will need to consider if the user has to hold a modifier key to select an annotation to edit?

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -41,167 +63,232 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
-    const TARGET = 900; // "comfortable" tagging width
-    const MAX = 1100; // don't exceed this even on huge screens
-    const MAX_SCALE = 3; // don't upscale more than 3× natural
+    api.viewer.gestureSettingsUnknown.dragToPan = navEnabled;
+    api.viewer.gestureSettingsUnknown.scrollToZoom = navEnabled;
+    api.viewer.gestureSettingsUnknown.clickToZoom = false;
+    api.viewer.gestureSettingsUnknown.dblClickToZoom = false;
 
-    const maxAvailable = Math.min(containerW, MAX);
+    // Critical fix for pan mode: let OSD receive drag/wheel events through the SVG layer.
```
jblake42: This seems like functionality that should be built into OSD - does it not support mouse wheel scrolling out of the box?

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -41,167 +63,232 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
+        api.viewer.element as HTMLElement,
+      );
+      if (nextOverlay) {
+        nextOverlay.style.pointerEvents = "auto";
+      }
+    };
+  }, [api, panMode]);
 
-    const maxScaled = natural.w * MAX_SCALE;
+  useEffect(() => {
```
jblake42: I would personally combine this useEffect with the previous one - it has the same dependencies and handles the same sort of logic

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -41,167 +63,232 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
     }
 
-    if (!isPolygonAnno(annotation)) return null;
-    const polygon = readPolygonGeometry(annotation);
+    const polygon = isPolygonAnno(annotation)
+      ? readPolygonGeometry(annotation)
+      : null;
     if (!polygon) return null;
 
-    const { minX, minY, maxX, maxY } = polygon.bounds;
```
jblake42, line 137: How come we have moved away from using bounds, does it not work?

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -41,167 +63,232 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
+        <ResetViewButton onPress={resetView} />
+
+        <OpenSeadragonAnnotator
           tool={drawingTool}
           drawingEnabled={drawingEnabled}
+          drawingMode="drag"
           autoSave
-          style={(_annotation, state) => ({
+          style={(
+            _annotation,
```
jblake42: Could this be typed to remove the error please - should be `Annotation` I think

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -41,167 +63,232 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
+            state?: { selected?: boolean; hovered?: boolean },
```
jblake42: A pre-built type exists for this - `AnnotationState`

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -41,167 +63,232 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
+          ) => ({
             strokeWidth: state?.selected ? 3 : state?.hovered ? 3 : 2,
```
jblake42: Do you know why the hover state doesn't work? It may be an annotorious bug, the event does not seem to be triggered unless the annotation is selected

toktagger/ui/src/app/video/components/ui_elements.tsx
jblake42, line 1: Just so the comment is there - we should aim to use spectrum here. Happy to ignore since the element is improved upon in later PRs

toktagger/ui/src/app/video/components/video-toolbox.tsx
```diff
@@ -233,20 +233,44 @@ export function VideoToolbox() {
           </div>
           <Grid columns={["1fr", "1fr"]} gap="size-100" marginBottom="size-100">
             <Button
```
jblake42, line 235: Could a `ToggleButtonGroup` be more suitable here?

jblake42 (changes requested):
Apart from the small code comments and the seperate comments this seems pretty close to being ready

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: I looked into this.

Take a look at the commits https://github.com/ukaea/toktagger/pull/221/changes/f8c55dc927fc6ff5b12821270d1bf9bb9fb49b05 and https://github.com/ukaea/toktagger/pull/221/changes/e8b08bb0c554550b728f0455c0042c607fb53162

Annotorious does document AnnotationState.hovered, but in our OSD setup with drawingMode="drag" and drawing enabled, the hover state does not seem to be emitted consistently for existing annotations unless the annotation is already selected/editable.

Since this is only a cursor affordance, I avoided changing annotation behavior and removed the dependency on state.hovered. Instead, we now do a small OSD-coordinate hit-test against the current annotation geometry and set the cursor explicitly: pointer in draw mode, default in drag/zoom mode.

This keeps the drawing/zoom behaviour unchanged and fixes the misleading cursor state.

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/221/changes/967446734861cd826404a3b2e22f322299be1c92

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/221/changes/f60e1b7484f3f86f44ab66376f5f276c0d0168b5

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/221/commits/0992ed5dea726e1032f9ef8c1452d4a6a8603056

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/221/commits/d2f05f85b9861634d730eba73749f02aa16134e3

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: addressed in commit https://github.com/ukaea/toktagger/pull/221/changes/15da4c4480948f98a668cfb379acf09c576a9c95

toktagger/ui/src/app/video/components/ui_elements.tsx
abdullah-ukaea: Agreed but given in future PRs I fix the UI I would prefer to keep this PR focused on the zoom functionality

toktagger/ui/src/app/video/components/video-toolbox.tsx
abdullah-ukaea: Agreed but given in future PRs I fix the UI I would prefer to keep this PR focused on the zoom functionality. I will keep an eye on this.

abdullah-ukaea:
I have finished the review, could you take another look and if you are happy feel free to merge. @jblake42

toktagger/ui/src/app/video/components/ui_elements.tsx
jblake42: Yep agreed

toktagger/ui/src/app/video/components/video-toolbox.tsx
jblake42: Agreed

jblake42:
Let's have a discussion regarding openseadragon typing and gesture settings in our catch-up before merging

toktagger/ui/package.json
```diff
@@ -26,7 +26,7 @@
     "rollup": "^4.59.0",
     "stream-browserify": "^3.0.0",
     "typescript-eslint": "^8.56.1",
-    "vite": "^7.3.2",
+    "vite": "^7.1.4",
```
jblake42: We shouldn't be reverting the vite version here and in the dev dependencies

jblake42 (changes requested):

toktagger/ui/package.json
jblake42: I have pushed a commit to correct this

PR #229, opened 2026-04-13
[WIP] Video Annotation: Keep Propagation Toggle State Across Sample Navigation

abdullah-ukaea:
Do Not Merge yet until PR https://github.com/ukaea/toktagger/pull/221 is merged!

This branch abdullah/fix-propagate-toggle is sourced from branch abdullah/video-zoom

This PR fixes the bug where the propagation toggle switched back on when moving to a different sample. I now keep this toggle in shared in-memory context for the current annotation session. That means the toggle now stays consistent while navigating samples, and it resets naturally on page refresh.


wk9874:
Closes #140

PR #247, opened 2026-04-20
Make clear button unvalidate samples

wk9874:
the Clear annotations button removes all annotations from the sample, but will not yet update these in the database. The user must press save to have those blank annotations saved, to avoid data loss from accidental deletion.

To make this clearer, we will update the database to reflect that the annotations are now unvalidated, as it is unclear whether the user intends to use the db annotations or UI (blank) annotations until the user presses Save.

After pressing clear, the sample is unvalidated, and the user can either:
1. Press save to update the db with 0 annotations, removing previous annotations, and setting back to validated
2. Refresh the page to retrieve the db annotations again, and press save to remark these as validated


praksharma (approved):
It is working as intended. Well at least in a python venv.

wk9874:
Blocked until I can get the CI passing again - unrelated to this PR, but `ray` has got rid of `local_mode` which we were relying on in models tests... see #248

PR #249, opened 2026-04-24
Add Model Training & Prediction Params

wk9874:
### Description
Different types of ML model will require different input parameters from the user at train time, eg number of epochs, batch size, etc. Since users can define ML models themselves, we want them to be able to define the parameters required, their types, and any default values to use. These should then be transmitted to the frontend so that they can be turned into a form for the user to enter values into.

Closes #226 
Closes #243 

#### Backend
* Model developer can now optionally specify a Pydantic model(s) which contains their parameters, types, default values, and any additional validation rules, which is accepted by `train` and `predict` methods of the Model baseclass
* This is added to the ModelRegistry using the same decorator as before
* New methods added to ModelRegistry to support getting the Pydantic class itself, as well as the JSON Schema of the class
* New `/meta/models/{model}` endpoints for train and predict added to get the JSON Schema for a given model from the ModelRegistry
* `/models/train` and `/models/predict` endpoint now accepts an optional `params` argument in the body, which should contain a filled in schema for the model
* Endpoints get the Pydantic class from the ModelRegistry and validates the provided params against it, returning any validation errors found to the frontend
* Parameters passed as a pydantic class instance through to the Ray worker node for use in the Model

#### Frontend
* Added `react-jsonschema-form` as a dependency
* Added custom widgets for each common input type which styles the form as React Spectrum
* Adds new function to core for getting a JSON schema from the backend
* JSON schema passed into the Form and rendered within the model train UI (if required)
* Submit button passed forwards by ref from the model train/predict modal to the form, so that in either case (params or no params), the submit button is in the same place
* When submitting, the form does initial validation against the schema, and returns validation errors without having to pass anything to the backend if invalid entries
* If successfully validated, params are passed to `/models/train` endpoint
* Refactored model predict tool in sample page
#### Other
* Added end to end tests to the models training UI which check both the cases where params are and are not defined
* `react-jsonschema-form` only works with JSON Schema rev 7, while Pydantic returns it as rev 2020. I've tried to convert the most common differences from rev 2020 back to being rev 7 compatible inside the ModelRegistry `get_params_schema` method, but there still may be issues with this... No better workaround as yet

### Testing
To test this PR, you can use the mock disruption project:
* Start toktagger
* Run `python scripts/create_mock_data.py`
* Open new project, click on model train button in top right
* Select `disruption_cnn` from the list

The expected params are defined in `toktagger/api/models/disruption.py` - you can try adding / removing / changing the type / changing default values of the params in here to see if the UI responds properly

PR #252, opened 2026-04-29
(Bugfix) Model save should use file stem

### LINKED ISSUE #239: .model ending in model storage breaks keras
**Describe the bug**
Should just check for any file name beginning with model ID inside get_actor

wk9874:
## Description
For models which save files that do not end in `.model` (eg, `.keras` / `.h5` files from Tensorflow), loading model files will break since the worker looks for this extension. 

## Changes
Provide the 'file stem' to the `save` method, which is just `{cache_dir}/models/{model_id}`, and let the user decide the file suffix. Worker uses `glob` to find the file in the cache dir which have file names starting with the model ID, and returns that to the `load` method.

Closes #239 

**Note**: Don't merge until after https://github.com/ukaea/toktagger/pull/249 is merged


praksharma:
Works just fine.

PR #260, opened 2026-04-30
Video Annotation: Keep Propagation Toggle State Across Sample Navigation

### LINKED ISSUE #140: Image Annotation: Add a `propagate annotations` toggle

abdullah-ukaea:
This PR fixes the bug where the propagation toggle switched back on when moving to a different sample. I now keep this toggle in shared in-memory context for the current annotation session. That means the toggle now stays consistent while navigating samples, and it resets naturally on page refresh.

Closes https://github.com/ukaea/toktagger/issues/140


jblake42 (commented):
I am not a fan of the view logic creeping back into the generic contexts. This stems both from the non-optimal structuring of the contexts, but also the use of conditional logic to change out components. This leads to remounting which affects the contexts. I'll make an issue to track this (or update an exisiting one)

PR #267, opened 2026-05-11
Numpy Array Dataloader for Images

### LINKED ISSUE #233: Support loading numpy arrays for image data
Load .npz files (all frames in one file)

wk9874:
This PR adds support for loading numpy array files (.npy or .npz) in as images into TokTagger.

Note that `.npz` files can contain multiple arrays, while `.npy` files always contain only one array. So for `.npz` files, the user should specify the signal name (name of the array to access in the npz file), or if they dont specify it and only one array is contained within the file, use that.

It should support arrays of shape (num_frames, width, height) for greyscale data (with min-max scaling applied if any numbers are greater than 255), or arrays of shape (num_frames, width, height, 3) for RGB data.

Closes #233


wk9874:
[frames.zip](https://github.com/user-attachments/files/27601510/frames.zip)

Here are some sample npy and npz files you can use for testing - 10000 and 10001 are greyscale, 10002 and 10003 are RGB

toktagger/api/core/data_loaders.py
```diff
@@ -157,6 +160,85 @@ def get_sample(self, sample: Sample, params: ImageParams, **kwargs) -> ImageData
...
+        if params.name != "image":
+            raise DataLoaderError("Must provide image data parameters!")
+        elif params.frame is None:
+            frame_arr = arr[0, ...]
+        elif params.frame > arr.shape[0]:
+            raise DataLoaderError(
+                f"Frame {params.frame} unavailable! Maximum available frame number is {arr.shape[0]}."
+            )
+        else:
+            frame_arr = arr[params.frame - 1, ...]
```
abdullah-ukaea: I think there is a frame indexing bug here. The rest of the video flow appears to treat frame `0` as the first frame, but this loader treats frame `1` as the first frame by using `arr[params.frame - 1]`.

That means if the UI requests frame `0`, this becomes `arr[-1]`, so NumPy returns the last frame instead of raising an error. The response then reports `frame=1` because `params.frame or 1` also converts `0` to `1`.

I think we should either make this loader zero-based to match the existing video flow, or explicitly reject `0` and negative frame numbers if one-based indexing is intended.

potential code improvement suggestion for zero based
```python
if params.name != "image":
    raise DataLoaderError("Must provide image data parameters!")

frame = params.frame if params.frame is not None else 0

if frame < 0 or frame >= arr.shape[0]:
    raise DataLoaderError(
        f"Frame {frame} unavailable! Available frame range is 0 to {arr.shape[0] - 1}."
    )

frame_arr = arr[frame, ...]

return ImageData(
    frame=frame,
    values=base64.b64encode(buffer.getvalue()).decode(),
)
```

toktagger/api/core/data_loaders.py
```diff
@@ -157,6 +160,85 @@ def get_sample(self, sample: Sample, params: ImageParams, **kwargs) -> ImageData
...
+            raise DataLoaderError(
+                f"""Expected array to have three dimensions representing (frame, height, width),
+                or 4 dimensions representing (frame, height, width, RGB),
+                but found {len(arr.shape)} dimensions!"""
+            )
+
+        # If any values > 255, scale arr to be 1-255
+        if np.any(arr > 255):
+            arr = (arr - arr.min()) / (arr.max() - arr.min())
+            arr = (arr * 255).astype(np.uint8)
```
abdullah-ukaea, line 222: Small edge case in the scaling logic: if the array contains values above `255` but all values are the same, then `arr.max() - arr.min()` is `0`, so this will divide by zero.

 Suggested code shape:
```python
  frame_arr = arr[frame, ...]
  max_value = frame_arr.max()

  if max_value > 255 or min_value < 0:
      value_range = max_value - min_value

      if value_range == 0:
          frame_arr = np.zeros_like(frame_arr, dtype=np.uint8)
      else:
          frame_arr = ((frame_arr - min_value) / value_range * 255).astype(np.uint8)
  else:
      frame_arr = frame_arr.astype(np.uint8)
```

  Then keep:
```python
  im = Image.fromarray(frame_arr)
```

For RGB arrays that are already proper uint8 values from 0 to 255, this exact divide-by-zero issue will not happen because the scaling branch is skipped. But the code is not only for RGB arrays; it also supports greyscale arrays and says it scales values above 255.

toktagger/api/schemas/samples.py
```diff
@@ -23,7 +28,9 @@ class ShotData(BaseModel):
     signal_names: Annotated[list[str], Field(min_length=1)]
 
 
-DataTypes = Union[TimeSeriesFileData, ImageFileData, ShotData]
+DataTypes = Union[
+    FileData, TimeSeriesFileData, ImageFileData, ImageArrayFileData, ShotData
+]
```
abdullah-ukaea: I think adding the base `FileData` type here makes sample validation too loose. Because `FileData` only requires `file_name` and `protocol`, a payload with an unsupported `type` could still validate as `FileData` instead of being rejected early.

Unless we need to support a generic file sample, I think this union should only include the concrete file data models

```python
  DataTypes = Union[
      TimeSeriesFileData,
      ImageFileData,
      ImageArrayFileData,
      ShotData,
  ]
```

Even better, if you want stricter validation by type, you could later move toward a discriminated union, but the simple fix is just to remove FileData from DataTypes.

toktagger/ui/src/types.ts
```diff
@@ -198,6 +198,11 @@ export const TimeSeriesFileDataSchema = FileDataSchema.extend({
 });
 export type TimeSeriesFileData = z.infer<typeof TimeSeriesFileDataSchema>;
 
+export const ImageArrayFileDataSchema = FileDataSchema.extend({
+  signal_name: z.string().optional(),
+});
+export type ImageArrayFileData = z.infer<typeof ImageArrayFileDataSchema>;
```
abdullah-ukaea, line 204: can we check the typing here

`ImageArrayFileDataSchema` is defined here, but it is not included in `SampleDataSchema`. Because `SampleDataSchema` currently only uses the broad `FileDataSchema`, array samples can lose the `signal_name` field when parsed as a `Sample`.

  Suggested change:
```JavaScript
  export const SampleDataSchema = z.union([
    TimeSeriesFileDataSchema,
    ImageArrayFileDataSchema,
    FileDataSchema,
    ShotDataSchema,
  ]);
```

toktagger/api/core/data_loaders.py
wk9874: Thanks - I originally had it this way, then swapped it to start at 1 because I thought that was how the others worked for some reason! Have reverted back to 0 based indexing

toktagger/api/core/data_loaders.py
wk9874: Good spot

toktagger/api/schemas/samples.py
wk9874: Dont even know why I did this...

abdullah-ukaea:
LGTM

PR #269, opened 2026-05-15
Use Mongita database in tests

wk9874:
The tests were written before the local pip installation was ready to go, and so uses a MongoDB testcontainer. This means that the tests wont work if someone is running them who does not have docker installed (and leads to instability, especially on Windows, as @jblake42  has found!) To fix this, will use the Mongita database with a temporary file instead.

## Changes
* Removes testcontainers dependency and removes `mongo_container` fixture
* Adds `cache_dir` option to MongoDBClient, and uses this to create a mongita client in conftest fixture
* Adds a `/projects/delete` endpoint to remove all projects and associated items, and changes relevant db utils function to be consistent with samples and annotations
* Adds new `/health` endpoint which returns info like the version, db connection status, model installation status and whether in testing mode
* Adds parameter to `Server` class denoting testing mode (not exposed via init since dont want user to set this)
* Change server creation fixture for end to end tests to also use Mongita, and adds extra checks that it has not accidentally connected to a user's real instance (using testing mode via the health endpoint)


wk9874:
@samueljackson92 could you review this code wise for me please, and @jblake42 could you check to see if it fixes the issues you were seeing on windows?

samueljackson92 (approved):
Changes are sensible and look good to me.

jblake42 (changes requested):
This seems to massively improve robustness on Windows following the addition of `proc.join()` as discussed directly

PR #255, opened 2026-04-29
(Bugfix) Do not allow untrained models to be used for predictions

wk9874:
## Description
Due to the way `get_actor` works, it is possible that the predictions are attempted to be made from a newly initialized ML model (if the weights cannot be found, or failed to be loaded). This could give wildly inaccurate predictions (depending on the setup of the model)

## Changes
Adds wrapped version of each method in `Model`, which check that the model was successfully trained / loaded before attempting to predict with it

Closes #239 

**Note**: Don't merge until after #252  is merged

PR #271, opened 2026-05-18
Add support for model prediction in Video toolbar

### LINKED ISSUE #240: Change model predict/train models to require same params as annotators
Model train and predict should (optionally) accept the dataparams so that ie you know which frame you want to predict on

### LINKED ISSUE #246: Image Annotation: Detect annotation changes in db and update UI
If we are using the `modelPredictSample` tool to create predictions for a given pulse / frame, we will want the UI to update with the annotations generated by that automatically without having to refresh. 

`modelPredictSample` uses `setAnnotations` from the `sampleContext` to update these on the frontend (as well as sending them over the API to the backend), and so can likely just monitor this for changes in a `useEffect` and push new annotations to the UI when necessary

wk9874:
## Description
Adds support for model predictions of single samples from the toolbar of the Video interface, to match what is available in time series.

## Changes
* Adds the modelPredictSample tool to toolbar in video mode as well as time series
* Adds support in the backend (model baseclass and `/samples/{sample_id}/models/{type`) for passing dataParams for the selected sample to the backend (so that, eg, the frame number is available to the model)

## To Test
Use `scripts.setup` to create the local Frame Project, then go into there and 'train' a model with the button in the top right.  Once trained, open the sample, and use the model prediction in the toolbar. Should be an option to either predict on only one frame, or all frames. Predictions will be a box of width 50, height 50 at (0,0) with the id 'test'.

To clear previous predictions, press Clear at the top of the toolbar, then Save, then Clear again
## To Dos
* ~Noticed a formatting issue in the model schema form - checkbox and array element descriptions overfill the toolbar column when using the model predict sample tool, and no combination of widths / minWidth / maxWidth seemed to fix it. Currently have hardcoded the size to 2500, which makes it fit in the toolbar nicely, but means it looks odd in the predict modal. Need some help figuring out what is causing this issue~ Fixed by Josh, setting max width of toolbar
* ~Inside modelPredictSample there is now some video specific logic, which I would like to remove ideally. This is because the `dataParams` stored inside `sampleContext` are not guaranteed to be correct when in video mode, and you need to access the frame from `data` directly to reconstruct the `effectiveDataParams`. Would prefer it to be that the dataParams in sampleContext is the single point of truth which both timeseries and video use~ Fixed by Abdullah
* ~Unlike timeseries, video doesn't seem to update annotations automatically based on when `setAnnotations` is called from sampleContext. This means that once predictions are ready and returned from the backend, `setAnnotations` is called, but you dont see any new boxes until you refresh. Would be good to have it work similarly to time series and auto update if possible~ Fixed by abdullah
* Remove temp model from temp.py and __init__.py once testing complete

**Note** Do not merge until #255 is merged

Closes #240 
Closes #246


wk9874:
I think this branch is ready for review, is that right @abdullah-ukaea ?

wk9874:
Found a bug - import annotations within the video sample page does now work correctly and it loads the annotation as a box on the page without needing a refresh. However it also overwrites all existing annotations in a labelling session. Weirdly it seems to make multiple of the imported annotations - eg if I have three existing annotations, then import a file with a single annotation, the 3 existing annotations disappear, and the imported annotation shows in the annotation list as having 4 instances:

https://github.com/user-attachments/assets/14834b3d-c32e-4d05-96c5-581b6e0a8bc0

wk9874:
If I have a model which creates a prediction over multiple frames, propagation of hand drawn annotations is disabled


https://github.com/user-attachments/assets/cccc924a-e72a-47d7-ab35-f037c041b91e

(not even sure what *should* happen here, so happy to be told that this is either expected behaviour, or is so insignificant that its not worth fixing!)

wk9874:
Model prediction boxes appear to not be selectable for some reason:

https://github.com/user-attachments/assets/85a6c58b-1dde-4ce8-8ee1-b256070963ba

abdullah-ukaea:
> Model prediction boxes appear to not be selectable for some reason:
> 
>  toktagger_prediction_not_selectable.mp4

for some reason I cannot reproduce this 

https://github.com/user-attachments/assets/8493886c-f214-4372-ad0d-8974ec690835

abdullah-ukaea:
> If I have a model which creates a prediction over multiple frames, propagation of hand drawn annotations is disabled
> 
>  model_predict_multiple.mp4 
> (not even sure what _should_ happen here, so happy to be told that this is either expected behaviour, or is so insignificant that its not worth fixing!)

as discussed, this is actually working as expected based on how propagation is currently implemented. However now that model prediction is being integrated into toktagger I need to rethink how to only propagate hand drawn annotations and NOT propagate model predicted annotations.

This will be tackled in a separate follow up PR / needs to be documented in an issue

abdullah-ukaea:
could you give this branch another functional test to help me identify any bugs? I think I resolved the main reported bug now. @jblake42 @wk9874

wk9874:
Yep the instance numbering bug when importing looks fixed now :) other two I think are future work, have raised issues #275  #276 

Found another bug, but I think this is probably a bug in my model Predict tooling implementation - if you ask for a model prediction on one frame, then go to the next frame and ask for a prediction on that one, itll remove the prediction from the first frame... suspect when I wrote the model predict tool only timeseries existed, and so it removes previous annotations from the given model before adding new ones (which ofc means that in the video case, it removes predictions from all frames before writing new predictions for the next requested frame). I'll look into fixing this today

abdullah-ukaea:
> Yep the instance numbering bug when importing looks fixed now :) other two I think are future work, have raised issues #275 #276
> 
> Found another bug, but I think this is probably a bug in my model Predict tooling implementation - if you ask for a model prediction on one frame, then go to the next frame and ask for a prediction on that one, itll remove the prediction from the first frame... suspect when I wrote the model predict tool only timeseries existed, and so it removes previous annotations from the given model before adding new ones (which ofc means that in the video case, it removes predictions from all frames before writing new predictions for the next requested frame). I'll look into fixing this today

sounds good, let me know if I need to change anything on the ui side to fix the prediction bug you mentioned

abdullah-ukaea:
ready for code review @jblake42

toktagger/ui/src/app/components/ui/schemaForm.tsx
```diff
@@ -380,10 +389,16 @@ const SpectrumArrayFieldTemplate = ({
       </Heading>
       {items.map((item) => item)}
       {schema.description && (
-        <Text>
-          {" "}
-          <em>{schema.description}</em>{" "}
-        </Text>
+        // For some reason anything other than hard coding a size here
```
jblake42: @wk9874 - was this fixed by setting the toolbar width. Can we remove the comment etc?

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -41,7 +41,7 @@ interface SampleContextType {
   setAnnotations: (
     updater: (annotations: Annotation[]) => Annotation[] | Annotation[],
   ) => void;
-  setDataParams: (params: DataParams) => void;
+  setDataParams: React.Dispatch<React.SetStateAction<DataParams>>;
```
jblake42: Is this typed like this for a reason or can we be consistent with the other instances below?

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -324,6 +330,16 @@ export function SampleProvider({
           if (typeof frame === "number" && Number.isFinite(frame)) {
             bootstrappedVideoSampleIdRef.current = sampleId;
             lastGoodVideoFrameRef.current = frame;
+            setDataParams((prev) => {
+              if (prev.name === "image" && prev.frame === frame) {
+                return prev;
+              }
+              return {
+                ...prev,
```
jblake42: We should not need to expand the existing previous frame here as you are overwriting all existing data anyway:
```
return {
  name: "image",
  frame,
};
```

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -140,6 +142,28 @@ export function stampLabelAndTrack(
...
+/** Ensure class label, track id, and backend creator bodies exist. */
+function stampLabelTrackAndCreator(
+  a: ImageAnnotation,
+  className: string,
+  trackId: string,
+  createdBy: string | null | undefined,
+): ImageAnnotation {
+  return stampCreator(stampLabelAndTrack(a, className, trackId), createdBy);
+}
+
```
jblake42: The functions are a little messy here imo - you have split functionality into the different types of stamps and then combined them in higher level functions.

I personally think keep them all seperate and then chaining them at the point of use would improve the clarity rather than having multiple functions performing varying levels of stamping

toktagger/ui/src/app/video/components/anno-utils.ts
```diff
@@ -151,6 +175,11 @@ export function getLabelTrack(a: ImageAnnotation): {
   };
 }
 
+/** Read backend creator metadata, falling back to manual at export time. */
```
jblake42: I think the comment is misleading - this function does not do any falling back as such, it just returns `null`. The logic that uses this does the falling back

jblake42: I think falling back to manual is valid, but in that case this function should return `manual` and then you don't have to repeat the line `created_by: getAnnotationCreator(a) ?? "manual",` you could just do `created_by: getAnnotationCreator(a)`

I don't mind either way but whichever you go for be clear in the comment

toktagger/ui/src/app/video/components/video-view.tsx
```diff
@@ -121,6 +125,7 @@ function VideoProvidersInner({ children }: { children: React.ReactNode }) {
         data={data}
         dataParams={dataParams}
         dbAnnotations={annotations ?? []}
+        setSampleAnnotations={setAnnotations}
         propagate={videoPropagate}
         setPropagate={setVideoPropagate}
```
jblake42, line 125: There may be a good reason, but can the `VideoSessionProvider` not just grab the context itself rather than have the inner provider do it for it. It seems a bit redundant to me and bloats the props?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -152,6 +141,191 @@ type FocusRequest = {
...
+          width: item.width,
+          height: item.height,
+          created_by: item.created_by,
+          timestamp: item.timestamp,
+        }),
+      );
+      continue;
+    }
+
+    entries.push(
```
jblake42: Should we be more explicit here and check that the type is as we expect - this will catch if an additional type is added and the logic is not updated

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -152,6 +141,191 @@ type FocusRequest = {
...
+
+    const dbAnno = parsed.data;
+    const key = buildSourceKey({
+      projectId: args.projectId,
+      sampleId: args.sampleId,
+      frame: dbAnno.frame,
+    });
+
+    const anno =
+      dbAnno.type === "video_bounding_box"
```
jblake42: Same here with explicitly checking types

jblake42: There are other instances below - I won't comment them all

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -280,22 +468,50 @@ export function VideoSessionProvider(props: {
     [byFrame],
   );
 
-  const seedFrame = useCallback((f: FrameIndex, list: ImageAnnotation[]) => {
-    setByFrame((prev) => mapSetFrame(prev, f, list));
-    // Seeding should not mark the session dirty.
-  }, []);
+  useEffect(() => {
```
jblake42, line 471: Keep `useEffect`s grouped together

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -433,119 +663,57 @@ export function VideoSessionProvider(props: {
...
+      projectId,
+      sampleId,
+    });
 
-      setByFrame(byF);
-      setDirty(false);
-    },
-    [byFrame.size, dirty, projectId, sampleId],
-  );
+    if (invalid > 0) {
```
jblake42: Do duplicates frequently happen or is there any actions that need to be taken if duplicates happen - if they are frequent or there is no action to be taken we shouldn't flood the log

jblake42 (changes requested):
Mainly focussed on front end due to time constraints

toktagger/ui/src/app/contexts/SampleContext.tsx
abdullah-ukaea: Good spot, there wasn’t a good reason for it to be typed differently. I’ve changed setDataParams to match the other context setters and cleaned up the VideoView call site so the typing is simpler. fixed in commit https://github.com/ukaea/toktagger/commit/ef902a8d36a8423cac4e23b4f99edee35ff24546

toktagger/ui/src/app/contexts/SampleContext.tsx
abdullah-ukaea: good spot, fixed in commit https://github.com/ukaea/toktagger/commit/da4a93caedd519538917b01c1b6899443802df45

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: Agreed, fixed in commit https://github.com/ukaea/toktagger/commit/8dd4d4c8981c10b6805beab675ba5d5565650484

toktagger/ui/src/app/video/components/anno-utils.ts
abdullah-ukaea: I have gone with your second suggestion as I think a fall back to manual rather than null is more correct, so now the comment is actually correct. fixed in commit https://github.com/ukaea/toktagger/commit/21662146aa3d7600fe305c044e9e054d1e366d46

toktagger/ui/src/app/video/components/video-view.tsx
abdullah-ukaea: I agree, there wasn’t a strong reason to pass those through as props since they all come from SampleContext. I’ve moved those context reads into VideoSessionProvider directly, so VideoProvidersInner now only passes the video-specific props it needs to. Fixed in commit https://github.com/ukaea/toktagger/commit/d09152b537c706c0ff362c5c151a3efb58ecc2f2

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/8cc8af1317062bcad02bc51802dd5c10ea99e391

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/8cc8af1317062bcad02bc51802dd5c10ea99e391

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: > There are other instances below - I won't comment them all

could you double check I got all the instances?

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/876420accbee5bb84bde96d1593cb440caa9b52f

Could you double check this commit is what you wanted me to do?

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: Good point, this was just leftover debug-style logging for a handled cleanup path. I removed the warnings and kept the duplicate cleanup behavior unchanged so we do not flood the console. fixed in commit https://github.com/ukaea/toktagger/commit/0d1829331a0681eddad901528a4921f171b0c8fa

abdullah-ukaea:
@jblake42 I have addressed your UI code review now. I think @wk9874 has one pending comment

toktagger/ui/src/app/components/ui/schemaForm.tsx
wk9874: Yep this is fixed, will remove the TODO

abdullah-ukaea:
ready to be merged ? @jblake42

toktagger/ui/src/app/video/components/video-session.tsx
jblake42: Yeah looks good

toktagger/ui/src/app/video/components/video-session.tsx
jblake42: This is a massive component and there are many useCallbacks and useEffects - my comment is that I think it is good practice to group all useEffects after the useCallbacks so developers know where to look. This is not critical, more a general practice we follow

jblake42 (approved):

PR #273, opened 2026-05-22
ML Models: Local Load Support

wk9874:
## Summary
Adds the ability to load pretrained ML model weights into the UI. This is only for local pip installed versions, and should be disabled via an environment variable for production hosted environments (which will have a Hugging Face method instead, but thats future work). You will be asked to provide the path to the weights on your system, which is then copied into the models cache dir. Ray then attempts to load these weights with the provided model, and if successful, saves them back into the cache directory with the correct name (model_id.suffix)

Also adds a bunch of test improvements so that we can test the functionality of toktagger both when the model optional dependencies are and are not installed (as I noticed during this PR that installing without models has been broken for a while)

## Changes
* Adds new model load modal
* Adds new Ray worker node job for loading model weights, which copies weights from a specified path to the cache dir temporarily, attempts to load them into the selected model, then saves them with the correct name and removes the temp copy 
* Adds model load endpoints, which are used to start a load job, and then check the result of it via polling
* Adds health endpoint which can be used to check if models dependencies installed
* Improves handling of model buttons when optional dependencies not installed, by disabling buttons and adding contextualhelp button which tells the user they are missing dependencies
* Refactors tests so that they do not require optional model dependencies to run
* Adds `models_enabled` and `models_disabled` pytest marks, for tests which explicitly require dependencies or not
* Changes CI to run with dependencies both installed and not installed


samueljackson92 (commented):
## Code Review — Local Load Support (PR #273)

Reviewed at high recall (catch every real bug a careful reviewer would catch). 10 findings below, posted as inline comments. Ranked most-severe first.

| # | File | Issue | Severity |
|---|------|-------|----------|
| 1 | `TimeSeriesContext.tsx:242` | `getAnnotation` always returns `null` (`forEach` not `.find()`) | **Critical** |
| 2 | `base.py:172` | `val_annotations = val_test_samples` copy-paste | **High** |
| 3 | `base.py:276` | `get_name` metaclass bug — always returns first model's name | **High** |
| 4 | `routers/models.py:731` | `return HTTPException` instead of `raise` | **High** |
| 5 | `sender.py:31` | Removed non-200 error check; callers ignore return value | **High** |
| 6 | `routers/models.py:29` | `not params` rejects valid `{}` for all-default-field models | **Medium** |
| 7 | `data_loaders.py:124` | `DataParams` default incompatible with `ImageDataLoader` | **Medium** |
| 8 | `TimeSeriesContext.tsx:169` | Fresh `uuidv4()` on every parse — selection state lost on every prediction | **Medium** |
| 9 | `modelLoad.tsx:83` | Polling interval duplicated on model name change mid-poll | **Low** |
| 10 | `worker.py:94` | Double `wrapped_load` when stale weights file exists | **Low** |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

toktagger/api/worker.py
```diff
@@ -63,6 +63,61 @@ def get_actor(project, model):
...
+        return {
+            "project_id": project.id,
+            "model_id": model.id,
+            "message": f"Worker node cannot find weights file at location {weights_path}",
+        }
+
+    model_actor = get_actor(project=project, model=model)
+    # Try loading actor with weights file, catch and reraise any errors
+    try:
+        load_temp_weights_task = model_actor.wrapped_load.remote(str(weights_path))
```
samueljackson92, line 97: **Bug: double `wrapped_load` call**

`get_actor` (line 59) already fires `ml_model.wrapped_load.remote(model_path)` when it finds a matching weights file on disk, and then `load_model` immediately queues a second `wrapped_load.remote` here. When a stale `.model` file exists for this model ID, the actor loads weights twice: first from the cached path, then from the user-supplied path. Between the two calls `_trained=True` but with the wrong weights, and the doubled I/O doubles startup latency.

Fix: either skip the auto-load inside `get_actor` when called from `load_model`, or move the initial-load logic out of `get_actor` entirely so `load_model` owns it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

toktagger/ui/src/app/components/tools/modelLoad.tsx
```diff
@@ -0,0 +1,264 @@
...
+        setMessage(
+          payload === false ? "Model failed to load!" : payload.detail,
+        );
+        setMessageIcon(<Alert aria-label="Failed" color="negative" size="S" />);
+        clearInterval(interval);
+        setIsLoading(false);
+      }
+    }, 1000);
+    return () => clearInterval(interval);
+  }, [project._id, selectedModelName, taskId]);
```
samueljackson92: **Bug: polling interval recreated on every `selectedModelName` change**

`selectedModelName` is in the dependency array (line 137), so if the user changes the model ComboBox selection while a load task is already in flight (`taskId` is set), the effect re-runs and spawns a second `setInterval` that polls against the original `taskId` but with the new model name. Both intervals run concurrently; the original keeps firing until it self-clears, producing spurious status updates. The stale-closure guard on line 95 (`if (selectedModelName == null) return`) does not help here because the new interval captures the new (non-null) name.

Fix: either remove `selectedModelName` from the dependency array and read it via a ref, or gate the effect so it only fires when `taskId` transitions from null to a value.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

toktagger/ui/src/app/components/tools/modelLoad.tsx
```diff
@@ -0,0 +1,264 @@
...
+            <Button
+              variant="accent"
+              onPress={submitLoadJob}
+              isDisabled={!weightsPath || !selectedModelName}
+            >
+              Submit
+            </Button>
+          </ButtonGroup>
+        </Dialog>
+      )}
```
samueljackson92, line 254: **Bug: Submit button permanently disabled on `hugging_face` tab**

`isDisabled={!weightsPath || !selectedModelName}` — `weightsPath` starts as `""` and is only updated by the local-path `TextField`. If the user opens the dialog and switches to the `hugging_face` tab (which has no path input), `!weightsPath` is always `true` so the Submit button stays permanently disabled regardless of model selection.

Fix: gate the `!weightsPath` part on `selectedTab === "local"`:
```tsx
isDisabled={!selectedModelName || (selectedTab === "local" && !weightsPath)}
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

toktagger/api/routers/models.py
```diff
@@ -543,7 +718,9 @@ async def get_sample_predictions(
         prediction_annotations = result.get("annotations_batch")
 
         # Check that annotations contain results for this sample ID
-        if not any(ann.sample_id == sample_id for ann in prediction_annotations):
+        if prediction_annotations and not all(
```
samueljackson92, line 719: **Behaviour change: any → all changes semantics + empty-result handling**

Two differences introduced here vs the base branch:

1. **any → all**: The old check raised 404 only if *no* annotation matched the requested sample. The new check raises 404 if *any* annotation belongs to a different sample. For a single-sample prediction this is fine, but if `get_predictions` ever returns a batch result containing annotations for multiple samples, the `all()` check would incorrectly 404 even though the requested sample's annotations are present.

2. **`prediction_annotations and ...`**: The old code raised 404 when `prediction_annotations` was empty (correct — no results). The new code skips the check entirely and returns an empty list. Is it intentional to let empty results through silently?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

samueljackson92:
**Additional findings in files outside this PR's diff**

These bugs were found while reviewing the full branch (`wk9874/models/local_load_support`). They are in files not changed by this PR, but several are high severity and worth addressing before or alongside this work.

---

**`toktagger/api/routers/models.py:731` (pre-existing, same file as this PR)**
`return HTTPException(...)` instead of `raise HTTPException(...)` — FastAPI sends a 200 with the exception object serialised as the body. Affects the else-branch of `get_sample_predictions` when the predict task is not found.

---

**`toktagger/ui/src/app/contexts/TimeSeriesContext.tsx:242` — Critical**
`getAnnotation` always returns `null`. The inner `return annotation` is inside a `forEach` callback, so it exits only the callback, not the outer function. Replace with `.find()`:
```ts
return annotations.find((annotation) => annotation.id === id) ?? null;
```

---

**`toktagger/api/models/base.py:172` — High**
Copy-paste bug: `self.val_annotations = val_test_samples` should be `self.val_annotations = val_test_annotations`. The validation split ends up with `Sample` objects in `self.val_annotations` instead of `Annotation` objects.

---

**`toktagger/api/models/base.py:276` — High**
`ModelRegistry.get_name` always returns the first registered model's name. The comparison `model_class.__class__.__name__ == model.__class__.__name__` evaluates `'type' == 'type'` for every class object in the registry (they're all instances of the `type` metaclass). Every non-first model actor gets `self.type` set to the wrong name, tagging annotations with the wrong model type.

Fix:
```python
if model_class == model  # direct identity check
```

---

**`toktagger/api/core/sender.py:31` — High**
The non-200 `RuntimeError` was removed from `send_updates`. Callers in `worker.py` (`send_model_updates`, `send_batch_annotations`) never check the return value, so HTTP failures silently drop status updates and annotations with no indication of failure.

---

**`toktagger/ui/src/app/contexts/TimeSeriesContext.tsx:169` — Medium**
`convertRawAnnotationsToTimeSeries` calls `uuidv4()` for every annotation on every parse. Stable backend annotations get new IDs each time `rawAnnotations` changes (e.g. after any prediction lands), invalidating any selected annotation's ID and silently clearing the selection.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

toktagger/api/worker.py
wk9874: `load_model()` is only called directly after a brand new model instance is created inside the database by the load API router. Therefore there will never be cached weights for that model ID, and so `get_actor` will always instantiate and return a new model without trying to load any cached weights

imo this is fine and doesn't need fixing, lmk your thoughts @samueljackson92

toktagger/ui/src/app/components/tools/modelLoad.tsx
wk9874: Good point

toktagger/ui/src/app/components/tools/modelLoad.tsx
wk9874: HF support hasnt been implemented yet, this was just a placeholder so I could check how the tabs looked. Removing any reference to HF for now, above issue can be fixed when HF support is added

toktagger/api/routers/models.py
wk9874: 1. All is correct here - endpoint is for getting predictions for a specific sample via a task ID, and so if that task has made predictions on anything other than the specified sample, something has gone badly wrong (possibly an incorrect task ID used). Should throw an error
2. Yes, it is correct to let empty annotations through. In the case where a model performed predictions but did not add any annotations (eg, detecting UFOs on a single frame of video which does not contain UFOs), then it is valid to return an empty list of annotations.

No change needed

wk9874:
> **`toktagger/api/routers/models.py:731` (pre-existing, same file as this PR)** `return HTTPException(...)` instead of `raise HTTPException(...)` — FastAPI sends a 200 with the exception object serialised as the body. Affects the else-branch of `get_sample_predictions` when the predict task is not found.

Fixed

> **`toktagger/api/models/base.py:172` — High** Copy-paste bug: `self.val_annotations = val_test_samples` should be `self.val_annotations = val_test_annotations`. The validation split ends up with `Sample` objects in `self.val_annotations` instead of `Annotation` objects.

Fixed

> **`toktagger/api/models/base.py:276` — High** `ModelRegistry.get_name` always returns the first registered model's name. The comparison `model_class.__class__.__name__ == model.__class__.__name__` evaluates `'type' == 'type'` for every class object in the registry (they're all instances of the `type` metaclass). Every non-first model actor gets `self.type` set to the wrong name, tagging annotations with the wrong model type.
> 
> Fix:
> 
> ```python
> if model_class == model  # direct identity check
> ```
Fixed as suggested, improved typing. Also highlighted that inside the Model base class, finding the name was using the wrong ModelRegistry. Have decided to add the .type parameter as a class var inside the model instead, which is set inside ModelRegistry, meaning it no longer needs to be assigned in the init within the worker


Other comments raised aren't related to this PR, so have been raised as issues

wk9874:
Closes #292

samueljackson92 (approved):
Gave it a quick whirl this morning. Looks good. Let's get it in dev.

PR #277, opened 2026-05-29
Move docs to zensical

samueljackson92:
Mkdocs and material (our current documentation framework) are having a crisis:  https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/

As suggested, in this post this post moves our docs over to zensical.


zensical.toml
```diff
@@ -0,0 +1,66 @@
...
+Home = "index.md"
+
+[[project.nav]]
+"Creating Projects" = "creating_projects.md"
+
+[[project.nav]]
+"Time Series Labelling" = "time_series.md"
+
+[[project.nav]]
+"2D Profile Labelling" = "profile_2d.md"
```
wk9874: Need to remove this for now as 2D profiles aren't merged yet

wk9874: Currently gives a 404

wk9874 (changes requested):
One small update needed then looks good to me!

Should go through at some point and improve the documentation, but that's probably for a different PR

PR #282, opened 2026-06-05
Swap out getOrInsert

wk9874:
getOrInsert is not supported on the JDC Firefox version since its quite old (2024) - so swap usage of this out for an if statement and set instead


blakeypark:
LGTM

blakeypark (approved):

PR #286, opened 2026-06-08
Add Config File and global Settings

wk9874:
We are starting to get a large number of environment variables which can be set, and will have lots more in the near future as different model options are added. So have added a option to define these settings in either a config toml file, or still via env vars (with slightly different names)

* Adds `pydantic-settings` as a dependency
* Defines new Settings object, which has a global instance accessed by all parts of code
* Updates env vars everywhere else in code to use new settings object
* Env vars inside worker left unchanged as these will not have access to settings object on worker nodea
* Adds tests and docs


samueljackson92 (changes requested):
Looks good, but can you add some example of the TOML file format somewhere in the repo? Perhaps heavily commented with the options?

wk9874:
Committed example toml file to repo, and have added a script which auto generates it based on the Settings schema. This will run as part of the build step of the CI to make sure its always up to date

samueljackson92 (approved):
Looks good to me.

PR #287, opened 2026-06-09
Fix Pages CI

wk9874:
Pages CI seems to be missing a couple of steps and fails to deploy the pages. Fixed according to instructions in: https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages


samueljackson92 (approved):
Looks good to me! Sorry I missed this!

Issue #235, opened 2026-04-14
Image Annotation: Add points


wk9874:
**Is your feature request related to a problem? Please describe.**
For tracking certain objects, we may want to be able to draw a dot, instead of a rectangle or polygon.

**Describe the solution you'd like**
Add a point annotation as an option, and create a new annotation schema on the backend which only expect x and y (or make height and width optional on the current schema)

abdullah-ukaea:
Closed by https://github.com/ukaea/toktagger/pull/288

PR #288, opened 2026-06-12
Video Annotation: Add X,Y points

abdullah-ukaea:
# First draft

**Please give me any feedback on this new feature, report any bugs you find etc**

Implemented video point annotations by reusing Annotorious rectangle shapes as the UI drawing primitive, since Annotorious does not support native points. Reference: https://annotorious.dev/api-reference/shape/

The UI adds a dotted crosshair and center dot over point rectangles to make the keypoint location clear.

This will close issue https://github.com/ukaea/toktagger/issues/235
**This PR needs to be merged first: https://github.com/ukaea/toktagger/pull/271**

# Demo

https://github.com/user-attachments/assets/708c95bb-d54e-4a37-b7ff-73cc4b2151ad


abdullah-ukaea:
# Second Draft

Note: The size of the fixed box can be changed by changing the variable https://github.com/ukaea/toktagger/blob/6c58574cda6a18e8d8850c016a073398a420ee82/toktagger/ui/src/app/video/components/anno-utils.ts#L37

https://github.com/user-attachments/assets/9763bf6f-2d88-41d9-9b5f-01958dd9beff

abdullah-ukaea:
# Update

https://github.com/user-attachments/assets/dd351608-50c0-4704-a647-47533b59de9a

abdullah-ukaea:
ready for review @jblake42

abdullah-ukaea:
@wk9874  could you double check the schema is correct for video point on the backend while Josh reviews the front end? https://github.com/ukaea/toktagger/blob/5c93e9c6bd7050775460d55c2f60f3ef23caa531/toktagger/api/schemas/annotations.py#L70

jblake42 (commented):
The code looks good to me - a couple of points:

- There are references to the points being ellipses or rectangles - does that need to stay as it complicates the code a little?
- I think the point side selection needs some refinement as it seems to reset on refresh and only updates new points
- I am getting a little concerned about the maintainability of the video code due to the scope creep of the tooling and needing to shoe horn in functionality that is not natively supported. We should assess this and see if there is anything we can do to improve this (my guess is there probably isn't too much we can do without forking annotorious

jblake42:
Another thing I have noticed is that points can be very hard to see on white backgrounds - I don't know if there is anything that can be done to improve this

jblake42:
LGTM - merged dev into branch as tests were failing

PR #297, opened 2026-06-18
Update the vite config & build

samueljackson92:
This fixes the static page build for Vite version 8.0


wk9874:
I dont think this has worked, since there is now no file in `static` named `plotly`, so suspect that regex isnt matching

It also looks like this form is deprecated already - https://main.vite.dev/guide/migration#removed-object-form-build-rollupoptions-output-manualchunks-and-deprecate-function-form-one

I think we might want this instead? From https://rolldown.rs/in-depth/manual-code-splitting

```
    rolldownOptions: {
      output: {
        codeSplitting: {
          groups: [
            {
              name: "react",
              test: /node_modules\/react\//,
            },
            {
              name: "plotly",
              test: /plotly/,
            },
          ],
        },
      },
    },
```

samueljackson92:
Closing. Solved in #298

PR #303, opened 2026-06-22
v0.2.1 Release

wk9874:
Prepeare release for v0.2.1

* Added support for ML model predictions from video UI toolbar
* Fixed bug so that video annotations update immediately on importing
* Added support for loading local weights files for ML models
* Fixed bug to correctly update sample validation status when annotations are imported
* Fixed bug where annotation `created_by` field was being overwritten when validated
* Fixed zooming bug in time series UI
* Increased minimum supported Python version to 3.10


wk9874:
Ignore that the CI says its failing, I cancelled all of them because about six CI jobs were running at once

Passing CI here: https://github.com/ukaea/toktagger/actions/runs/27947767030/job/82696761593

samueljackson92 (approved):
CI is passing on the link above. Looks good to me.

PR #304, opened 2026-06-22
Time series event detection models

samueljackson92:
Adds four new time series event detection models to toktagger:

 - DTW motifs - use user generated annotations as motifs and use [Dynamic Time Warping](https://en.wikipedia.org/wiki/Dynamic_time_warping) to detect similar events within the time series trace.
 - Stumpy - use user generated annotations as motifs for the [matrix profile](https://stumpy.readthedocs.io/en/latest/Tutorial_The_Matrix_Profile.html) to find similar motifs in the data.
 - Minirocket - use [minirocket](https://arxiv.org/abs/2012.08791) and a linear probe to classify time series windows.
 - Shapelets - use a [shapelet](https://www.sktime.net/en/stable/api_reference/auto_generated/sktime.classification.shapelet_based.ShapeletTransformClassifier.html) transform to classify time series windows (using sktime's default RotationForest estimator, an ensemble of decision trees).

Each model supports either single or multi-channel inputs.

**To Test:**
 - open a project with a time series view.
 - click on the train models button. 
 - train one of the models
 - once trained, close the training window
 - open the model predictions interface
 - run the model
 - annotations should be generated for this shot and plotted.

Additionally, check the new tests & CI are passing


abdullah-ukaea:
**Known issue: multiple signal inputs**

 As described in https://github.com/ukaea/toktagger/issues/330, the parameter description says that users can provide multiple signals, but the current schema form only renders one input and provides no way to add another. The Pydantic schema supports multiple values, the problem is in the UI’s list handling.

@wk9874  is looking into this. I have also prototyped a possible fix in https://github.com/ukaea/toktagger/blob/slj/ts-event-detection-models-schemaFix/toktagger/ui/src/app/components/ui/schemaForm.tsx

abdullah-ukaea:
The new models trained and predicted successfully in my functional testing, but the training forms contain too much text and require excessive scrolling. Overall it is is not a good UI experience.

Could we show only a short model summary and brief parameter descriptions in the form? The detailed explanations could perhaps live in the Zensical documentation, with a link from the UI if needed.

The user currently sees:
stumpy_motif_ui_model_train_form

stumpy_motif_ui_model_train_form_part2

I think the model summaries could be closer to:
* DTW: “Detects events that have a similar shape to labelled examples using Dynamic Time Warping.”
* STUMPY: “Detects events that are similar to labelled examples using STUMPY.”
* MiniRocket: “Learns to detect labelled events using MiniRocket time-series features.”
* Shapelet: “Learns to detect events using characteristic shapes found in the training signals.”

Parameter descriptions should be similarly concise.

The form text is generated from the model docstrings in these files, so make the changes here:
* toktagger/api/models/shapelet.py
* toktagger/api/models/stumpy_motif.py
* toktagger/api/models/dtw_motif.py
* toktagger/api/models/minirocket.py

toktagger/api/models/event_detection_utils.py
```diff
@@ -0,0 +1,141 @@
...
+    if not positions:
+        return []
+
+    positions = sorted(positions)
+    regions: list[tuple[int, int]] = []
+    start = end = positions[0]
+
+    for pos in positions[1:]:
+        if pos <= end + 1:
+            end = max(end, pos)
```
abdullah-ukaea, line 119: **I think this needs to be improved:**

This only merges detections whose start positions are one sample apart. With a window size of 100 and a step size of 10, positions such as [0, 10, 20, 30] represent overlapping windows but are returned as separate annotations.

Could we merge detections when their windows overlap or touch instead?

```python
if pos <= end + window_size:
    end = max(end, pos)
```

_Optional Improvement: Add a test confirming that overlapping windows with a step size greater than one produce a single annotation._

toktagger/api/models/dtw_motif.py
```diff
@@ -0,0 +1,266 @@
...
+            ta = np.array(data.values[params.signal_names[0]].time)
+            if len(params.signal_names) == 1:
+                va = np.array(data.values[params.signal_names[0]].values, dtype=float)
+            else:
+                va = np.array(
+                    [
+                        np.array(data.values[n].values, dtype=float)
+                        for n in params.signal_names
+                    ]
+                )  # (n_channels, n_samples)
```
abdullah-ukaea, line 108: We need to be careful here, in this PR we claim generic multichannel support, but this code assumes that every selected signal uses the same time points and sampling rate.

Each signal has its own time array, but this code uses the first signal’s times for every channel. Signals sampled at different rates could either fail when stacked or combine values from different times.

Could we resample the selected signals onto a shared time grid before combining them? At a minimum, we should validate that their lengths and time arrays match and return a clear error.

toktagger/api/models/minirocket.py
```diff
@@ -0,0 +1,313 @@
...
+                len(windows), 1, window_size
+            )
+
+        y = np.array(labels)
+
+        self.log_progress(progress=20)
+
+        transformer = MiniRocket(num_kernels=params.num_kernels)
+        transformer.fit(X)
+        X_features = transformer.transform(X)
```
abdullah-ukaea, line 204: Why are we using the univariate MiniRocket transformer for multichannel data? The sktime documentation recommends MiniRocketMultivariate for this case.

Could we select the transformer based on the number of signals?

```python
transformer_class = MiniRocketMultivariate if multivariate else MiniRocket
transformer = transformer_class(num_kernels=params.num_kernels)
```

**Reference**: https://www.sktime.net/docs/examples/transformation/minirocket/#2-Multivariate-Time-Series

_Optional improvement: add a test confirming that multichannel training uses MiniRocketMultivariate._

toktagger/api/models/minirocket.py
```diff
@@ -0,0 +1,313 @@
...
+
+        all_labels = [
+            ann.label
+            for _, _, anns in sample_data
+            for ann in anns
+            if hasattr(ann, "time_min")
+        ]
+        pos_label = (
+            max(set(all_labels), key=all_labels.count) if all_labels else "Event"
+        )
```
abdullah-ukaea, line 132: My understanding is that this is a binary classifier:
* 1 means any annotated event.
* 0 means background.

Currently, all annotation labels are combined into the positive class, and predictions are assigned the most common label. For example, a model trained on both ELM and Sawtooth annotations could label every prediction as ELM.

Could we either let the user select the label to train on or reject training when multiple labels are present? Supporting multiple labels properly would require multiclass classification.

toktagger/api/models/shapelet.py
```diff
@@ -0,0 +1,276 @@
...
+            ann.label
+            for _, _, anns in sample_data
+            for ann in anns
+            if hasattr(ann, "time_min")
+        ]
+        pos_label = (
+            max(set(all_labels_list), key=all_labels_list.count)
+            if all_labels_list
+            else "Event"
+        )
```
abdullah-ukaea, line 136: The Shapelet model has the same issue as MiniRocket regarding the label classification problem.

During training, every annotated window gets:

```python
labels.append(1)
```

This removes the original annotation label. For example, both **ELM** and **Sawtooth** become class `1`, while background windows become class `0`.

There are two reasonable approaches:
1. **Keep the model binary.**
   - Add an `event_label` parameter and train only on annotations with that label.
   - Alternatively, reject training when multiple labels are present.

2. **Make the model multiclass.**
   - Store the original annotation labels in `y` instead of converting them all to `1`.
   - Preserve the classifier's predicted label when creating annotations.
   - Background would also need its own class.

toktagger/api/models/shapelet.py
```diff
@@ -0,0 +1,276 @@
...
+        y = np.array(labels)
+
+        self.log_progress(progress=20)
+
+        classifier = ShapeletTransformClassifier(
+            max_shapelets=params.max_shapelets,
+            n_shapelet_samples=params.n_shapelet_samples,
+            batch_size=params.batch_size,
+        )
+        classifier.fit(X, y)
```
abdullah-ukaea, line 193: The PR description says that Shapelets uses _“a shapelet transform and a linear probe to classify time-series windows.”_

 However, ShapeletTransformClassifier uses RotationForest by default, which is an ensemble of decision trees rather than a linear probe. A linear probe would be something like RidgeClassifierCV, logistic regression, or a linear support vector machine.

Should we provide a linear estimator explicitly, or update the PR description if using RotationForest is intentional?

reference: https://www.sktime.net/docs/api-reference/sktimeclassificationshapelet-basedshapelettransformclassifier/

toktagger/api/models/stumpy_motif.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def predict(
+        self,
+        samples: list[Sample],
+        params: StumpyMotifPredictParams | None = None,
+        data_params: DataParams | None = None,
+    ) -> list[list[AnnotationBase]]:
+        signal_names: list[str] = self.model["signal_names"]
+        multivariate = len(signal_names) > 1
+        threshold_override = params.threshold if params else None
```
abdullah-ukaea, line 164: I think we need to check our threshold override in training and prediction as this looks strange.

The threshold selected during training is saved, but prediction always defaults to 3.0, which replaces the saved value even when the user has not intentionally chosen an override.

Could we either make the prediction threshold optional and fall back to the saved value, or remove it from the training parameters and make it prediction-only? This would make the intended behaviour clearer.

toktagger/api/models/stumpy_motif.py
```diff
@@ -0,0 +1,269 @@
...
+        threshold: float = (
+            threshold_override
+            if threshold_override is not None
+            else self.model["threshold"]
+        )
+        multivariate = signal_vals.ndim == 2
+
+        n = signal_vals.shape[-1] if multivariate else len(signal_vals)
+        if n <= window_size:
+            return []
```
abdullah-ukaea, line 211: This has an edge case?

If the signal and window have the same length, there is still one valid comparison. The current check returns no detections instead.

Could we change this to n < window_size and maybe add a test where the signal length exactly matches the window size?

toktagger/ui/tsconfig.tsbuildinfo
abdullah-ukaea, line 1: please delete this file, we should not be committing this to the remote branch. It does not exist in dev.

toktagger/ui/src/app/components/tools/modelTrain.tsx
```diff
@@ -66,19 +122,94 @@ export function ModelTrainModal({
...
+    })();
+  }, [selectedModelName]);
+
+  // Poll model list while modal is open (Trained Models tab + training spinner)
+  useEffect(() => {
+    if (!modalOpen || !project._id) return;
+
+    const fetchModels = async () => {
+      const response = await getModels(project._id!);
+      if (!response.ok) return;
```
abdullah-ukaea, line 157: Could we handle errors here instead of returning silently? If the initial request fails, models remains null and the UI shows “No trained models yet,” which is misleading.

tests/api/models/test_new_models.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def test_predict_threshold_override_increases_hits(self):
+        """A very low threshold should yield at least as many hits as a high one."""
+        model = self._make_trained(["Ip"], threshold=3.0)
+        sample = make_sample()
+        # Very permissive threshold
+        low_result = model.predict([sample], StumpyMotifPredictParams(threshold=0.001))
+        # Very strict threshold (effectively no detections)
+        high_result = model.predict([sample], StumpyMotifPredictParams(threshold=1e6))
+        assert len(low_result[0]) >= len(high_result[0])
```
abdullah-ukaea, line 192: These thresholds are described the wrong way around. Since a detection requires distance < threshold, 0.001 is strict and 1e6 is permissive.

The assertion can still pass if both results are merged into the same number of annotations. Could we use controlled data with a known difference and check that the permissive threshold accepts more matches than the strict threshold?

tests/api/models/test_new_models.py
```diff
@@ -0,0 +1,325 @@
...
+            [sample], StumpyMotifPredictParams(threshold=training_threshold)
+        )
+        assert len(default_result[0]) == len(explicit_result[0])
+
+    def test_multivariate_train_predict(self):
+        model = self._make_trained(["Ip", "dalpha"])
+        sample = make_sample()
+        result = model.predict([sample])
+        assert len(result) == 1
+        assert isinstance(result[0], list)
```
abdullah-ukaea, line 210: This is a very basic smoke test, the assertions do not prove that multichannel prediction works correctly, we should improve this if the PR states we have multi-channel support.

These tests confirm that multichannel training and prediction do not crash, but they do not prove that the additional channels are used. For example, result = [[]] would pass these assertions even if the second channel were ignored.

Could we use test data where the event pattern exists only in the second channel and check that changing this channel changes the prediction? Shapelet should also have a multichannel test. Different channel lengths or time points should be covered by the shared alignment tests.

abdullah-ukaea (commented):
The code review is not finished.

These files still need to be checked:

* ~toktagger/api/models/__init__.py~
* ~toktagger/api/routers/models.py~
* ~toktagger/api/core/data_loaders.py~
* ~toktagger/api/routers/meta.py~
* ~toktagger/api/schemas/models.py~
* ~toktagger/api/models/base.py~
* ~toktagger/api/main.py~
* ~toktagger/api/worker.py~

toktagger/api/models/__init__.py
```diff
@@ -5,7 +5,9 @@
 
 
 def models_dependencies_installed() -> bool:
-    return importlib.util.find_spec("ray") is not None
+    # ray is a base dependency; check for one of the models-extra packages instead
```
wk9874, line 8: Why is ray a base dependency? It shouldn't be!

During my merge of dev back into this branch, I changed it back to ray being optional. This should be changed back to look for ray

toktagger/api/models/__init__.py
```diff
@@ -17,5 +19,18 @@ def check_models_enabled():
 
 
 if models_dependencies_installed():
-    from toktagger.api.models.disruption import DisruptionCNN as DisruptionCNN
-    from toktagger.api.models.temp import VideoCNN as VideoCNN
+    # These models only need base dependencies (numpy, scipy, scikit-learn, ray)
```
wk9874, line 22: This comment is incorrect, I think we want `ray` to be an optional models dependency, and all of the models to only be imported if all the models dependencies are installed (as they already are below)

The reason for this is the `models_dependencies_enabled` check is used to determine whether the frontend should enable/disable the model training buttons, so all models should either be enabled or disabled at once

If we want to add things which always show up, they should be annotators, not models imo

toktagger/api/models/__init__.py
```diff
@@ -17,5 +19,18 @@ def check_models_enabled():
...
+    from toktagger.api.models.stumpy_motif import StumpyMotifModel as StumpyMotifModel
+    from toktagger.api.models.minirocket import MiniRocketModel as MiniRocketModel
+    from toktagger.api.models.shapelet import (
+        ShapeletTransformModel as ShapeletTransformModel,
+    )
+
+    # Torch-based models are only imported when torch is available, so that Ray
+    # workers (which run in isolated venvs without torch) can still import
+    # toktagger.api.models without hitting an ImportError.
+    if importlib.util.find_spec("torch") is not None:
```
wk9874, line 34: Not needed, torch and ray are optional model dependencies together

toktagger/api/models/base.py
```diff
@@ -9,16 +9,14 @@
...
 from toktagger.api.core.sender import send_model_updates
 from toktagger.api.schemas.models import ModelUpdate
-from toktagger.api.models import models_dependencies_installed
 import pydantic
 import uuid
 from collections import OrderedDict
 import logging
 
-logger = logging.getLogger("ray")
+import ray
```
wk9874, line 17: Change this back to

```
if models_dependencies_installed():
    import ray
```

toktagger/api/models/base.py
```diff
@@ -280,6 +278,21 @@ def get_name(cls, model_class: typing.Type[Model]) -> str:
             name for name, model in cls._registry.items() if model_class == model
         )
 
+    @classmethod
+    def class_refs(cls) -> dict[str, tuple[str, str]]:
```
wk9874, line 282: Do we need this if all models optional dependencies will be available on all ray workers?

toktagger/api/models/base.py
```diff
@@ -312,6 +325,13 @@ def get_params(
             raise ValueError(f"No Model class called '{name}' found in registry!")
         return params
 
+    @classmethod
+    def get_description(cls, name: str) -> str | None:
+        import inspect
```
wk9874, line 330: import not at top level

you need to put this in your agents.md file or something that it shouldnt do this! ;)

toktagger/api/models/base.py
```diff
@@ -354,12 +374,25 @@ def get_params_schema(
...
+        if registered is None:
             raise ValueError(f"No class called '{name}' found in registry!")
+        if (
+            isinstance(registered, tuple)
+            and len(registered) == 2
+            and isinstance(registered[0], str)
+        ):
+            import importlib
+
+            module = importlib.import_module(registered[0])
```
wk9874, line 394: Don't really like this, feels quite unreadable

Don't think we need it if torch and ray are in the same dependency group

Or is the point that it reduces worker startup time if torch isnt needed?

How long does it take to import torch? Surely cant be a big performance hit?

toktagger/api/schemas/models.py
```diff
@@ -12,17 +12,6 @@ class ModelIn(ConfiguredModel):
     score: float
     task_id: Optional[str] | None = None
 
-    @field_validator("type")
-    def check_model_type(cls, value):
```
wk9874, line 16: Why was this removed?

tests/api/models/test_event_detection_utils.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def _make_ann(time_min, time_max):
+    ann = MagicMock()
+    ann.time_min = time_min
+    ann.time_max = time_max
+    return ann
+
+
+class TestZscore:
```
wk9874, line 20: Why are the tests in here inside a class? We don't do that anywhere else, would prefer them to not be in classes for consistency

tests/conftest.py
```diff
@@ -15,7 +15,7 @@
 import importlib
 import pathlib
 
-MODELS_ENABLED = importlib.util.find_spec("ray") is not None
+MODELS_ENABLED = importlib.util.find_spec("dtaidistance") is not None
```
wk9874, line 18: Again would change this back

wk9874 (changes requested):

wk9874:
+ tests have broken since merging into dev, please fix :)

abdullah-ukaea:
We have finished the first review of this branch, feel free to address the changes when you are ready @samueljackson92

samueljackson92:
@abdullah-ukaea this should be ready for re-review now.

abdullah-ukaea:
As discussed, could we make these two usability improvements before I start the next phase of code review? @samueljackson92 

1. In the training form, make Signal Names a dropdown containing the signals available for the current sample, rather than requiring users to type them manually.

2. In the prediction tool, let users select a specific trained model by name. Currently, selecting a model type such as MiniRocket automatically uses its latest version. If users have trained models named test1 and test2, there is no way to choose between them.

model training signal name

model predict no model name option

toktagger/api/schemas/models.py
```diff
@@ -7,12 +7,19 @@
...
     status: Literal["queued", "training", "loading", "failed", "completed", "aborted"]
     progress: Annotated[float, Field(strict=True, ge=0, le=100)]
     score: float
     task_id: Optional[str] | None = None
 
+    @property
+    def annotator_name(self) -> str:
+        """Name recorded against the annotations this model produces. Models loaded
+        from pretrained weights have no name, so fall back to their type."""
+        return self.name or self.type
```
abdullah-ukaea, line 21: Users can give multiple models the same name, but we use that name to identify and delete their predictions. This means predictions from different models could be mixed up or removed.

Could we use the unique model ID for prediction tracking and keep the model name for display only?

Image

toktagger/api/routers/models.py
```diff
@@ -758,16 +772,30 @@ async def create_sample_predictions(
...
 
+    # Remove any unvalidated predictions from a previous run of this model so
+    # that stale annotations don't accumulate in the DB across prediction runs.
+    await utils.delete_annotations(
+        db_client,
+        project_id=project_id,
+        sample_id=sample_id,
+        created_by=model.annotator_name,
+        validated=False,
+    )
```
abdullah-ukaea, line 797: The previous successful predictions are removed as soon as a new job is submitted. If something fails, the user loses the old results and gets no replacement.

Maybe keep the previous predictions until the new inference has completed successfully. Then you can replace the old predictions as part of the successful write, ideally in one database operation

toktagger/api/models/event_detection_utils.py
```diff
@@ -0,0 +1,201 @@
...
+    ref_time = np.asarray(signal_data[ref_idx][0], dtype=float)
+
+    aligned = []
+    for ta, va in signal_data:
+        ta = np.asarray(ta, dtype=float)
+        va = np.asarray(va, dtype=float)
+        if ta.shape == ref_time.shape and np.array_equal(ta, ref_time):
+            aligned.append(va)
+        else:
+            aligned.append(np.interp(ref_time, ta, va))
```
abdullah-ukaea, line 93: I think the shared alignment helper needs more consideration.

This aligns every signal to the longest signal’s full time range, so missing sections are filled with repeated values that could affect training and prediction. It also assumes the time arrays are valid and match their value arrays.

Example: Say you have one signal "Ip" between 0 - 10 seconds and another signal "dalpha" between 2-8 seconds, np.interp fills the missing edges with repeated endpoint values.

Could we validate each signal and align them only over their shared time range, returning a clear error when the data is incompatible?

Examples of potential errors cases:
- Return a clear error for an empty overlap
- Return a clear error for non-increasing or non-finite times
- Return a clear error for mismatched time/value lengths

toktagger/api/models/stumpy_motif.py
```diff
@@ -0,0 +1,266 @@
...
+        logger.info(f"StumpyMotif: inferred window_size={window_size}")
+
+        multivariate = len(params.signal_names) > 1
+        templates: list[tuple[np.ndarray, str]] = []
+        for ta, va, anns in sample_data:
+            for ann in anns:
+                if not (hasattr(ann, "time_min") and hasattr(ann, "time_max")):
+                    continue
+                if params.class_label and ann.label != params.class_label:
+                    continue
```
abdullah-ukaea, line 122: Are you sure other labels should affect the selected class’s window size? This doesn't look right.

The window size is calculated before filtering by class_label, so annotations from other labels can incorrectly affect it. For example, long H-mode annotations could make the window for an ELM model much larger.

Could we calculate the window using only annotations that match class_label, and use all annotations only when class_label is blank?

toktagger/api/models/minirocket.py
```diff
@@ -0,0 +1,311 @@
...
+            for ann in anns:
+                ann_time_pairs.append((ann, ta))
+            sample_data.append((ta, va, anns))
+
+        if not ann_time_pairs:
+            raise ValueError(
+                f"Signals {params.signal_names} not found in any annotated sample."
+            )
+
+        window_size = compute_window_size(ann_time_pairs)
```
abdullah-ukaea, line 120: Similar issue to comment https://github.com/ukaea/toktagger/pull/304/changes#r3960474554 

To summarise I am concerned that other non selected class labels affect the selected class’s window size.

I think we should filter ann_time_pairs by params.class_label before calculating the window size.

toktagger/api/models/shapelet.py
```diff
@@ -0,0 +1,276 @@
...
+            for ann in anns:
+                ann_time_pairs.append((ann, ta))
+            sample_data.append((ta, va, anns))
+
+        if not ann_time_pairs:
+            raise ValueError(
+                f"Signals {params.signal_names} not found in any annotated sample."
+            )
+
+        window_size = compute_window_size(ann_time_pairs)
```
abdullah-ukaea, line 127: Similar issue to comment https://github.com/ukaea/toktagger/pull/304/changes#r3960474554 and https://github.com/ukaea/toktagger/pull/304/changes#r3960539720

To summarise I am concerned that other non selected class labels affect the selected class’s window size.

I think we should filter ann_time_pairs by params.class_label before calculating the window size.

toktagger/api/models/minirocket.py
```diff
@@ -0,0 +1,311 @@
...
+        params: MiniRocketTrainParams,
+    ) -> float:
+        from sktime.transformations.panel.rocket import (
+            MiniRocket,
+            MiniRocketMultivariate,
+        )
+
+        self.log_progress(status="training", progress=0)
+
+        paired = [(s, a) for s, a in zip(samples, annotations) if a]
```
abdullah-ukaea, line 90: Should reviewed background-only shots be ignored here?

This filter removes validated shots with no annotations, even though they are useful negative examples for an event-versus-background classifier.

Could we keep these shots for background sampling and require at least one positive and one negative window before training?

What do you think?

toktagger/api/models/shapelet.py
```diff
@@ -0,0 +1,276 @@
...
+        self,
+        samples: list[Sample],
+        annotations: list[list[Annotation]],
+        params: ShapeletTrainParams,
+    ) -> float:
+        from sktime.classification.shapelet_based import ShapeletTransformClassifier
+
+        self.log_progress(status="training", progress=0)
+
+        paired = [(s, a) for s, a in zip(samples, annotations) if a]
```
abdullah-ukaea, line 95: same potential problem as the comment in minirocket https://github.com/ukaea/toktagger/pull/304/changes#r3960615623  

useful background shots ignored.

toktagger/api/models/minirocket.py
abdullah-ukaea, line 120: Another edge case to be aware of is that compute_window_size permits windows as short as two samples, but sktime's MiniRocket requires at least nine. 

Could be more of an issue if you do the filter comment above I suggested https://github.com/ukaea/toktagger/pull/304/changes#r3960539720. 

Could we check for window_size < 9 and either pad the windows or return a clear error before fitting?

References:
- https://www.sktime.net/docs/examples/transformation/minirocket/
- https://github.com/sktime/sktime/blob/main/sktime/transformations/rocket/_minirocket.py

toktagger/ui/src/app/projects/project_id/samples/sample_id/page.tsx
```diff
@@ -101,6 +105,9 @@ function SamplePageContent(props: { sampleId: string }) {
       <Provider theme={defaultTheme}>
         <ToastContainer placement="top" />
         <SampleDataBreadCrumbs project={project} sample={sample} />
+        <View position="fixed" top="size-100" right="size-100" zIndex={9999}>
+          <ModelTrainModal project={project} isEnabled={modelsEnabled} />
+        </View>
```
abdullah-ukaea, line 110: Should model training really be available from the individual sample page?

Opening the training form from one shot suggests that only this shot will be used, but the model actually trains on all validated samples in the project. Nothing in the modal explains that the current shot is not the only training input, so users can easily train on different data from what they expect. 

This misleads users when doing model training. Adding training here is a non-trivial UI decision that needs some clear functional justification.

Could we keep training on the samples page like before, or if you insist on keeping it here clearly explain in the modal that all validated project samples will be used? 

Personally, I think model training should remain on the samples page like before unless we add a feature to support training on the current shot only. 

Image

Image

toktagger/api/routers/meta.py
```diff
@@ -68,15 +107,62 @@ async def get_model_load_method_allowlist(load_method: LoadMethods) -> str | Non
...
 
 
+@router.get(
+    "/models/{model}",
+    dependencies=[Depends(check_models_enabled)],
+)
+async def get_model_meta(model: str) -> dict[str, typing.Any]:
+    """Get metadata (name, description, tasks) for a specific model type."""
+    description = ModelRegistry.get_description(model)
+    tasks = [str(t) for t in ModelRegistry.tasks(model)]
```
abdullah-ukaea, line 117: I think the api uses tasks values such as "time-series" but we seem to be using "Task.TIME_SERIES" please fix

toktagger/ui/src/app/components/tools/modelTrain.tsx
```diff
@@ -67,18 +125,101 @@ export function ModelTrainModal({
...
+        const errorMessage = await response.json();
+        setMessage(errorMessage.detail);
+        setMessageIcon(<Alert aria-label="Failed" color="negative" size="S" />);
         return;
       }
-      const newSchema: RJSFSchema | null =
-        await getModelTrainSchema(selectedModelName);
-      setSchema(newSchema);
+      const data = (await response.json()) as Model[];
+      setModels(data);
```
abdullah-ukaea, line 168: Bad typing and validation 

This cast tells TypeScript to trust the response but does not check that the API actually returned valid models. The model-names response above has the same issue. 

Could we validate both responses with Zod safeParse and show an error if fetching or validation fails?

toktagger/api/core/data_loaders.py
```diff
@@ -281,26 +282,46 @@ def get_sample(
...
+            elif path.endswith(".feather"):
+                return pd.read_feather(path, columns=item.signal_names)
+            else:
+                raise ValueError("Unsupported file format {}".format(Path(path).suffix))
+
+        if len(matched) == 1:
+            df = _read_file(matched[0])
         else:
-            raise ValueError("Unsupported file format {}".format(file_path.suffix))
+            df = pd.concat([_read_file(p) for p in matched])
```
abdullah-ukaea, line 324: I may be misunderstanding the intended data layout, but I assume this change is meant to support a single shot stored across multiple files. Could you briefly explain the intended use case for changing the tabular data loader.

But regardless, I don’t think this multi-file glob is combining the files correctly as you are expecting.

We first need to define how the matched files are structured, because stacking them as rows can create repeated times and misaligned signals.

Could we define the supported layout and combine the files using an explicit time index?

toktagger/ui/src/app/projects/project_id/components/add_samples.tsx
```diff
@@ -166,6 +175,10 @@ export const AddSamplesEditor = ({
   };
 
   useEffect(() => {
+    // ShotData projects never show the directory fields, so a listing of the
+    // server's working directory must not overwrite the range-derived shot IDs.
```
abdullah-ukaea, line 179: why is this a multi line comment, please run a prompt to change all multi line comments in this branch to single line comments unless the code is so unclear it really justifies that multi line comment.

abdullah-ukaea (changes requested):
## Side Note

In similar spirit to https://github.com/ukaea/toktagger/pull/387#pullrequestreview-5021258023 I think this PR includes a few changes beyond just adding the time-series event-detection models. Changes such as multi-file tabular loading could have been handled in a separate PR.  Smaller PRs make the intended behaviour clearer, are quicker to review and test, and make things easier for both authors and reviewers.

## Summary of review

Thanks for addressing the earlier feedback. 

The main areas I think still need attention before merging are:
-  I really think model training should probably remain on the project samples page. From an individual shot page, most users will not realise that they are training on all validated project samples.  See https://github.com/ukaea/toktagger/pull/304#discussion_r3960788098
- There are still some time series model correctness edge cases around window sizes, background-only shots, MiniRocket’s minimum window length etc
- Model names are not unique but are used to identify predictions and existing predictions are deleted before replacement predictions succeed.
- I am not convinced the multi-file tabular/Parquet loader combines files correctly. Have you functionally tested it?
- There are many verbose multi-line comments that add unnecessary noise to our codebase. Claude in general loves adding comments so we need to tighten claude.md to combat this.
- This branch is quite stale, I recommend solving the minor merge conflict and pulling the latest dev branch.

The core functionality is more or less done with some bugs, but we really need to do some tidying up before this can be merged.

PR #310, opened 2026-06-30
Frontend annotation UUID to backend

jblake42:
This PR passes the UUID of an annotation generated by the frontend to the backend to allow it to remain constant throughout its lifetime. This has two benefits:
- Will make existing frontend tracking of annotations by the front end more consistent as it currently just relies on not fetching updates from the backend
- Will facilitate additional tooling functionality such as multi-click creation as annotations can be tracked across saves


jblake42:
This has been superseded by work in #314

PR #314, opened 2026-07-01
D3-based Polygons

jblake42:
PR for D3-based polygons with the following functionality:

- Click to draw
- Click to add points during edit
- Editable vertices (drag and delete)
- Cancellable during drawing

Assigning an ID to the back end annotations was avoided due to incompatibility of UUID with Mongo Object IDs.

https://github.com/user-attachments/assets/77a65be5-5654-424b-8893-f0832233d2d2


toktagger/ui/src/app/components/ui/annotationsTable.tsx
abdullah-ukaea: There are two BOUNDING_BOX cases in this switch. The second one is duplicate code that can never run, and it references a marker variable that no longer exists.

I think you just delete the second BOUNDING_BOX case. The new first case you just added already handles bounding-box annotations correctly.

toktagger/ui/src/app/contexts/TimeSeriesContext.tsx
```diff
@@ -172,10 +183,27 @@ export const TimeSeriesProvider = ({
...
+    if (project.bounding_box_labels) {
+      project.bounding_box_labels.forEach((label, index) => {
+        const category_id = `${TimeSeriesAnnotationType.POLYGON}_${label}`;
+        timeSeriesCategories.set(category_id, {
+          label,
+          color: randomColor(index),
+          type: TimeSeriesAnnotationType.POLYGON,
+        });
+      });
+    }
```
abdullah-ukaea: I think this is wrong.

Polygon categories are using bounding_box_labels, so the polygon tool may show the wrong labels.

I think you should use project.polygon_labels instead of project.bounding_box_labels in this block.

it should be something like

```ts
if (project.polygon_labels) {
  project.polygon_labels.forEach((label, index) => {
    const category_id = `${TimeSeriesAnnotationType.POLYGON}_${label}`;
    timeSeriesCategories.set(category_id, {
      label,
      color: randomColor(index),
      type: TimeSeriesAnnotationType.POLYGON,
    });
  });
}
```

toktagger/ui/src/types.ts
```diff
@@ -5,6 +5,7 @@ export const BaseAnnotationSchema = z.object({
   project_id: z.string().nullable().default(null),
   sample_id: z.string().nullable().default(null),
   shot_id: z.number().optional(),
+  id: z.string().optional(),
```
abdullah-ukaea: So I think it was decided that the frontend UUID should not be sent to the backend because Mongo expects an ObjectId.

But Since frontend UUIDs are intentionally no longer sent to the backend, is this id field still needed? Backend annotations use _id, and locally converted annotations no longer include id, so annotation.id does not appear to preserve anything in the normal flow.

I think if frontend IDs are now intended to be temporary and frontend-only, remove id from BaseAnnotationSchema and generate the UUID directly. If IDs still need to remain stable, they should be preserved separately in frontend state without sending the UUID to Mongo.

toktagger/ui/src/app/utils.tsx
```diff
@@ -70,7 +73,7 @@ export function convertRawAnnotationsToTimeSeries(
   if (TimeRegionSchema.safeParse(annotation).success) {
     const timeRegion = TimeRegionSchema.parse(annotation);
     return {
-      id: uuidv4(),
+      id: annotation.id || uuidv4(),
```
abdullah-ukaea: depending on your decision on the comment I made in types.ts you may want to change this to just  id: uuidv4()

toktagger/ui/src/app/utils.tsx
```diff
@@ -85,7 +88,7 @@ export function convertRawAnnotationsToTimeSeries(
   if (TimePointSchema.safeParse(annotation).success) {
     const timePoint = TimePointSchema.parse(annotation);
     return {
-      id: uuidv4(),
+      id: annotation.id || uuidv4(),
```
abdullah-ukaea: depending on your decision on the comment I made in types.ts you may want to change this to just id: uuidv4()

toktagger/ui/src/app/utils.tsx
```diff
@@ -97,21 +100,42 @@ export function convertRawAnnotationsToTimeSeries(
   if (BoundingBoxSchema.safeParse(annotation).success) {
     const boundingBox = BoundingBoxSchema.parse(annotation);
     return {
-      id: uuidv4(),
+      id: annotation.id || uuidv4(),
```
abdullah-ukaea: depending on your decision on the comment I made in types.ts you may want to change this to just id: uuidv4()

toktagger/api/schemas/annotations.py
```diff
@@ -60,6 +60,15 @@ class VideoBoundingBox(AnnotationBase):
     y_min: int
 
 
+class Polygon(AnnotationBase):
+    type: Literal["polygon"] = "polygon"
+    segmentation: list[float] = Field(
+        ...,
+        min_length=6,
```
abdullah-ukaea, line 67: I have been wondering if for polygon typing we should check if segmentations contain at least six values AND **have an even length**?

if this is the case it would also have to be changed for video polygon.

But not too important for now feel free to ignore, just an idea.

Examples:

```bash
# Valid
[1, 2, 3, 4, 5, 6]

Three complete points:
(1, 2), (3, 4), (5, 6)

# Invalid
[1, 2, 3, 4, 5, 6, 7]

Final x value (7) has no matching y value.
```

toktagger/ui/src/app/contexts/TimeSeriesContext.tsx
abdullah-ukaea, line 429: The range-selection logic handles time points, time regions, and bounding boxes, but not polygons. Polygons therefore always become unselected when using Plotly’s selection tool.

Add a POLYGON case and decide whether a polygon should be selected when it overlaps the selected range or only when it is completely inside it.

abdullah-ukaea (commented):
Overall, this looks really good. The polygon tooling works well, feels polished, and is performant. I’ve left a few comments. Could you take a look and let me know your thoughts? @jblake42

toktagger/ui/src/app/components/ui/annotationsTable.tsx
jblake42: Good catch, I've deleted the redundant state which was a carry over from before this PR.

toktagger/ui/src/app/contexts/TimeSeriesContext.tsx
jblake42: I agree! I got lucky (or unlucky otherwise I might have noticed) that the categories are the same for the two. Changed

toktagger/ui/src/types.ts
jblake42: I think I will get rid of this for now and just generate a new UUID every time we load as this makes it more clear that the front end is not passing anything back.

As discussed we might change things once a decision is made on what approach will be taken by the back end.

toktagger/ui/src/app/utils.tsx
jblake42: As above

toktagger/ui/src/app/utils.tsx
jblake42: As above

toktagger/ui/src/app/utils.tsx
jblake42: As above

toktagger/api/schemas/annotations.py
jblake42: I have added a field validator. I don't know how important it is, but there is no harm in including it and should flag any underlying issues with the front end logic.

toktagger/ui/src/app/contexts/TimeSeriesContext.tsx
jblake42: Yeah this is a good point. I also need to improve the selection logic of the bounding box too. I'll do that now in this PR as long as it doesn't get too big

jblake42: Done

abdullah-ukaea (approved):

abdullah-ukaea:
Looks good, @jblake42 feel free to merge when ready

PR #315, opened 2026-07-03
Hotfix: Cast float image arrays to ints

wk9874:
Bug spotted by Steve - when using the image array data loader, if the image arrays are floats, then the data loader crashes as it Pillow cannot save the array as an image. The data loader does cast to int, but only inside an if statement where it rescales the image for values > 255. Moved this casting out of the if statement so that it always runs, and added a test to catch this edge case!


SteveAEA (approved):
All looks good and makes sense.

PR #317, opened 2026-07-03
UI Usability Tweaks from Labelling

samueljackson92:
Some tweaks based on my experience trying to label time series data from MAST. The following changes make adjustments to the UI in order to avoid repetitive actions over multiple shots. Also this fixes some edge case issues with data loaders, UI etc.

Video frame data param fix
   - Fixed the sample context frame data param update (setDataParams) inside the video frame fetch handler so that image-mode samples correctly track the current frame. This resolves the batch change type regression.

TimeSeriesContext UX improvements
  - Session persistence: editMode and activeTool are now persisted to sessionStorage per project and restored on mount, so the user's tool selection and edit/view mode survive page reloads.
  - Keyboard shortcut guard: The e key for toggling edit mode now ignores events originating from editable DOM nodes (inputs, textareas, selects) to prevent accidental mode switches while typing.
  - Clearer shortcut semantics: Pressing e now also clears the current annotation tool before toggling edit mode.

Annotation toolbar improvements
 - Category label persistence: Category → label allocations are saved to sessionStorage per project and restored on load. Valid saved labels are checked against the current category list before applying.
 - Help banner persistence: The "first-time edit help" banner now uses localStorage (ts-help-seen) so it genuinely shows only once across sessions.
 - Keyboard shortcut hint: The edit/view mode tooltip now mentions the e key shortcut, and the contextual help text was updated accordingly.

Data loader path resolution
  - ImageDataLoader and ArrayDataLoader now resolve file paths with .resolve() before checking existence, fixing issues when paths are given as relative or symlinked paths.
  - TabularDataLoader refactored to resolve the file path once, then use file_path.suffix instead of repeated str.endswith() checks — cleaner and avoids matching .jsonl against .json.


toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -349,6 +390,7 @@ export function SampleProvider({
         setError(err instanceof Error ? err.message : "An error occurred");
       } finally {
         setIsLoading(false);
+        setInitialLoadComplete(true);
```
abdullah-ukaea: I think this is a little problematic with the new autosave you have introduced.

initialLoadComplete is set to true even when loading fails. This can trigger autosave with an empty annotation list and overwrite annotations already stored for the sample. Successful loads also cause an unnecessary save because lastPersistedRef is not initialized with the loaded annotations.

Only enable autosave after annotations load successfully. At that point, set lastPersistedRef.current = dbAnnotations. Do not enable autosave from the finally block, since it also runs after failed requests.

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -165,8 +171,43 @@ export function SampleProvider({
...
+      `${BACKEND_API_URL}/projects/${projectId}/samples/${sampleId}/annotations`,
+      {
+        method: "PUT",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify(annotations),
+      },
+    ).catch((err) => {
+      console.error("Failed to persist annotations:", err);
+    });
+  }, [projectId, sampleId, annotations]);
```
abdullah-ukaea: **This could cause unintended overwriting of annotations if the user quickly makes changes.**

Each annotation change starts a new PUT request without waiting for the previous one. Because each request replaces all annotations, overlapping requests can save them in the wrong order or leave incorrect data. The code also marks annotations as saved before checking whether the request succeeded, and fetch does not treat HTTP errors as rejected promises.

I think you should allow only one save request at a time and queue the latest changes. Check response.ok, and only update lastPersistedRef after a successful response. Maybe also show a save error in the UI and allow the save to be retried

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -165,8 +171,43 @@ export function SampleProvider({
...
+
+  useEffect(() => {
+    if (!initialLoadComplete) return;
+    saveAnnotations();
+  }, [initialLoadComplete, saveAnnotations]);
```
abdullah-ukaea: I will include a suggestion at the end of the review of how to improve the autosave option so that it can be an optional setting, but the way it is currently done means that the  **Save** and **Save on Navigate** buttons on the left toolbar are bypassed and a user can no longer discard edits, and disabling Save on Navigate no longer means that changes stay unsaved. 

This particular code snippet makes backend persistence unconditional.

I would like to have a chat with you and Josh to discuss this a little bit before this particular feature is merged. I have some ideas on how to handle crash recovery / autosaving.

toktagger/ui/src/app/contexts/SampleContext.tsx
```diff
@@ -165,8 +171,43 @@ export function SampleProvider({
...
+      if (allMatch) return;
+    }
+
+    lastPersistedRef.current = annotations;
+    fetch(
+      `${BACKEND_API_URL}/projects/${projectId}/samples/${sampleId}/annotations`,
+      {
+        method: "PUT",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify(annotations),
```
abdullah-ukaea: One more potential problem with the way the new autosave is currently is done is that I think validation of annotations has not been considered. Currently an annotation should only be validated if a human presses save on the left toolbar (or has save on navigate toggle ON) and it should only be unvalidated if a human presses clear on the left toolbar.

When a user edits a validated sample, autosave stores the changed annotations with validated: false, but the sample can remain marked as validated. The UI may therefore show “Annotations Validated” even though its annotations are now unvalidated.

Mark the sample as unvalidated whenever its annotations change, both in the UI and backend. Only mark the sample and annotations as validated again after the user explicitly saves and that request succeeds.

toktagger/ui/src/app/contexts/TimeSeriesContext.tsx
```diff
@@ -96,22 +107,46 @@ export const TimeSeriesProvider = ({
...
   const [annotations, setAnnotations] = useState<TimeSeriesAnnotation[]>([]);
   const [toolingCallbacks, setToolingCallbacks] = useState<
     Map<TimeSeriesAnnotationType, ToolingCallbacks>
   >(new Map());
   const [activeTool, setActiveTool] = useState<TimeSeriesToolDefinition | null>(
-    null,
+    () => {
+      const saved = sessionStorage.getItem(`ts-active-tool-${projectId}`);
+      return saved ? (JSON.parse(saved) as TimeSeriesToolDefinition) : null;
+    },
```
abdullah-ukaea: The saved tool is parsed and treated as valid without checking it. Invalid JSON can crash the time-series view, while an old tool or label could be restored even if it no longer exists in the project.

I think we should safely parse and validate the saved value against the project’s current tools and labels. If it is invalid or outdated, remove it and default to no active tool.

docs/time_series.md
```diff
@@ -28,7 +28,9 @@ At the top of the interface, you'll find navigation controls to move through you
 - **Previous Button** (◄): Navigate to the previous sample in your project
 - **Next Button** (►): Navigate to the next sample in your project  
 - **Save Button**: Save your current annotations
-- **Delete Button**: Remove selected annotations
+- **Clear Button**: Remove selected annotations, and mark annotations in the database as not validated
+- **Save On Navigate**: Whether to automatically save any annotations to the database when you move to a different sample
+- **Jump to Shot**: Jump to a sample with a given shot ID
```
abdullah-ukaea, line 33: Just wanted to highlight that these bullet points are technically not 100% correct now or are missing some important details. 

As they now contradict the behavior introduced by this PR. Should we also not tell the user that there changes are being automatically updated now?

```markdown
- **Save Button**: Save your current annotations
- **Save On Navigate**: Whether to automatically save any annotations to the database when you move to a different sample
```

Also maybe we should add that the save button validates the samples?

But before changing the docs, lets do this last. First you need to decide what to do with autosave (read comment at the end of the PR) / address some of the code comments I made in this PR.

toktagger/ui/tsconfig.tsbuildinfo
abdullah-ukaea, line 1: I just noticed that this file has been introduced to the UI code.

Can you please delete it? and perhaps add it to the .gitignore for the project or the UI code so that it stops creeping in. 

For context this is TypeScript's generated incremental compiler cache and does not exist in dev.  Committing it adds a very large one-line file that will create unrelated churn.

abdullah-ukaea (commented):
Overall, this is a good PR that introduces useful quality-of-life improvements for people working in the toktagger platform for longer sessions.

My main concern is the new autosave behaviour. It introduces some data-loss and request-ordering risks, changes the meaning of the existing Save button and Save on Navigate option, and is currently mandatory for every user.

I think we should agree on the intended saving and validation workflow in our next meeting. My preference would be to add an **"Automatically Save Changes" toggle** next to Save on Navigate, with a tooltip such as: “Automatically save annotation changes as unvalidated drafts.”

This would support both workflows:
  - Users who want recovery and continuous persistence can enable autosave.
  - Users who prefer explicit saving, or may want to discard their changes, can leave it
    disabled.

Once the issues raised in this review are addressed, the autosave implementation should also be more reliable, with clearer handling of failed or overlapping requests and consistent annotation validation state.

Image

Image

abdullah-ukaea:
**Here is an alternative idea I wanted to suggest. No actions are required for this feel free to ignore.**

 For crash recovery such as a 
- browser error
- accidental navigation
- refresh
 
We could keep a temporary recovery copy of unsaved annotations in the browser instead of continuously writing them to the backend. When the user returns, they could choose to restore or discard those changes, preserving the existing Save workflow.

And again we would serve this as an optional feature the user could opt out from in the UI.

Not sure how feasible this idea is though.

abdullah-ukaea:
This looks merge-ready now. The autosave feature has been removed, preserving the existing Save and Save on Navigate workflows, and stored tools and labels are now properly validated before being restored. The remaining review comments have also been addressed. Approved.

one quick thing before I merge, could you update the PR description now so it is up to date? @samueljackson92

abdullah-ukaea (approved):

PR #319, opened 2026-07-07
Gitlab & Hugging Face load support

wk9874:
Initial commit adding Gitlab & HF Load Support

Not 100% happy with the UI elements, lots of repetition and I'm sure it can be tidied.

Gitlab loading doesn't work on UKAEA instance since its on too old of a version, I hope this will be bumped fairly soon so I can finish that part & test

Hugging Face loading works, added a temporary YOLOv8 implementation into `api/models/temp.py` which can be used to test. This whole file should be removed before this PR is merged

Can take an initial look at the implementation if you get time @samueljackson92 / @abdullah-ukaea , and then I can respond to comments when I'm back from AL

TODOs:
* ~Safe tensors limiting~
* ~Move Gitlab / HF settings to their own subheadings~
* ~Tests~


wk9874:
@praksharma also one for you to look at to see how best to integrate it with your HF model recipes stuff! I'm hoping that effectively the dummy YOLO model implementation I have in `api/models/temp.py` would be replaced by your generalised HF / YOLO models?

samueljackson92:
### Code review

Found 2 issues:

1. **`all()` called with multiple positional arguments instead of an iterable — crashes every Gitlab load attempt.** Python's `all()` takes one iterable; calling it as `all(a, b)` raises `TypeError` unconditionally, before the intended validation can ever run. This breaks the Gitlab load feature outright.

   - https://github.com/ukaea/toktagger/blob/c174d35d5fa94de572fa50c6544d6e04a4c558e3/toktagger/api/routers/models.py#L450-L457 (API endpoint precheck)
   - https://github.com/ukaea/toktagger/blob/c174d35d5fa94de572fa50c6544d6e04a4c558e3/toktagger/api/worker.py#L190-L198 (Ray worker precheck)

   Fix: wrap args in a list, e.g. `all([config.settings.models.gitlab_url, config.settings.models.gitlab_token])`.

2. **`MODELS_GITLAB_URL` / `MODELS_GITLAB_TOKEN` are never forwarded to the Ray worker process, so Gitlab loading fails even when correctly configured.** `_setup_ray()`'s `env_vars` dict only forwards `API_URL`, `MODEL_STORAGE`, and `MODELS_SAFETENSORS_ONLY`. This pattern (explicit propagation into `runtime_env`) was established by an earlier commit specifically because Ray workers can't see the API process's env/settings otherwise — but the two new Gitlab variables read via `os.environ.get(...)` in `worker.py` were never added to the list.

   - https://github.com/ukaea/toktagger/blob/c174d35d5fa94de572fa50c6544d6e04a4c558e3/toktagger/api/main.py#L64-L77 (missing keys in `env_vars`)
   - https://github.com/ukaea/toktagger/blob/c174d35d5fa94de572fa50c6544d6e04a4c558e3/toktagger/api/worker.py#L190-L198 (reads that will always see `None`)

   Fix: add `"MODELS_GITLAB_URL": ...` and `"MODELS_GITLAB_TOKEN": ...` to the `env_vars` dict in `_setup_ray()`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

wk9874:
This is now ready for review @samueljackson92 :)

samueljackson92:
---

### Code review

Found 1 issue:

1. `IndexError` can mask the intended error response in the load-status error handler (this PR changed `f"Load task failed unexpectedly - {e}."` to use `str(e).strip().splitlines()[-1]`, which raises `IndexError` if `str(e)` is empty — e.g. a bare `Exception()`, `asyncio.CancelledError()`, or certain Ray task-failure exceptions — since `"".splitlines()` returns `[]` and indexing `[-1]` on an empty list raises)

https://github.com/ukaea/toktagger/blob/c7eb8cb5330ffb0a8da169b5e17638187cf7def3/toktagger/api/routers/models.py#L577-L584

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

---

tests/model.pt
samueljackson92, line 1: Maybe we should move this to `tests/data` or something?

toktagger/api/core/worker.py
```diff
@@ -0,0 +1,533 @@
...
+import pydantic
+from mlflow import MlflowClient, MlflowException
+from huggingface_hub import hf_hub_download
+from safetensors import safe_open
+import requests
+
+logger = logging.getLogger("ray")
+logger.setLevel("DEBUG")
+
+models_dir_default = pathlib.Path(user_cache_dir("toktagger", "ukaea")).joinpath(
```
samueljackson92: Just checking: does this respect where the user set it in the config?

toktagger/api/core/worker.py
```diff
@@ -0,0 +1,533 @@
...
+    if ml_model:
+        gpu_available = ray.get(ml_model.gpu_available.remote())
+        if use_gpu and not gpu_available:
+            # Stop existing actor
+            logger.info("Existing Actor does not have access to a GPU!")
+            try:
+                ray.get(ml_model.__ray_terminate__.remote())
+            except ActorDiedError:
+                ml_model = None
+                logger.info("Stopped existing Actor, restarting with GPU...")
```
samueljackson92, line 52: This function is quite long. Perhaps this can be refactored into a separate function.

toktagger/api/core/worker.py
```diff
@@ -0,0 +1,533 @@
...
+    send_model_updates(
+        project_id=project.id,
+        model_id=model.id,
+        updates=ModelUpdate(training_status="completed", progress=100),
+    )
+
+    return {"project_id": project.id, "model_id": model.id, "message": None}
+
+
+def load_model_gitlab(
```
samueljackson92, line 94: This function is enormous. Consider refactoring. These loading functions might be better as classes (or the functions instantiate and call classes).

toktagger/ui/src/types.ts
```diff
@@ -275,6 +275,28 @@ export const ModelSchema = z.object({
...
 
 export type Model = z.infer<typeof ModelSchema>;
+
+export const LocalLoadFormSchema = z.object({
+  weights_path: z.string().min(1),
+});
+export type LocalLoadForm = z.infer<typeof LocalLoadFormSchema>;
+
+export const GitlabLoadFormSchema = z.object({
+  model_name: z.string().min(1),
```
samueljackson92: `.nonempty()` is better than `.min(1)`

docs/assets/model_load_gitlab.png
samueljackson92, line 1: I think this needs some improvement:

1. Optional model version should be last.
2. I should be able to just enter a single string like `Ultralytics/YOLOv8` for the repo.
3. I got a timeout with the following settings, but maybe I am using it wrong:

Image

samueljackson92: Then after i try and change the settings I get this error, which persists between closing and reopening the dialog:

Image

tests/model.pt
wk9874: I think we should probably just remove it once we're happy with the functionality of the load methods... Prakhar's PR should implement this properly anyway

It's weird that you get an error with models not installed though, as its import is guarded by a `if models_dependencies_installed()`m and mine works fine... Could you double check with a fresh install?

toktagger/api/core/worker.py
wk9874: Yes,

`os.environ["MODEL_STORAGE"] = os.environ.get("MODEL_STORAGE", str(models_dir_default))`

and that env var is set on the ray worker

However, that does mean it makes a redundant directory if that config option is set. Also the default location is now handled in the config object and the env var is therefore always present. Can just create the dir here

toktagger/api/core/worker.py
wk9874: Is it that long? It's only about 35 lines :)

samueljackson92: Ok not that long, but it feels like it's doing multiple things...

toktagger/api/core/worker.py
wk9874: Agreed, refactoring into classes

docs/assets/model_load_gitlab.png
wk9874: > I think this needs some improvement:
> 
> 1. Optional model version should be last.
> 2. I should be able to just enter a single string like `Ultralytics/YOLOv8` for the repo.
> 3. I got a timeout with the following settings, but maybe I am using it wrong:
> 
> Image

1. Done
2. I intentionally separated the userspace/org out from the project, because I wanted to add the ability to limit the organization which people can pull models from to something trusted for security. For example, you could set `huggingface_userspace = 'UKAEA'` in your toktagger.toml file, and the UI will show this:
image
3. Hmm, surprised this timed out for you, mine worked when I replicated your settings. However the way I've currently got it is that the spinner will poll for a max of 60s, then 'time out' even though the job is correctly submitted and will be executed when a worker is available. Two options here: could either remove this limit, and have the spinner go until the load is complete, or have a 'load job submitted to the queue' message instead, similar to the training modal.

wk9874: > Then after i try and change the settings I get this error, which persists between closing and reopening the dialog:
> 
> Image

That error is expected, since for any given model you should only be able to have one train/load job pending at a time (and your earlier load job is still queued (maybe stuck on something?))

It should probably clear on closing/opening the modal though, will look into that tomorrow

toktagger/api/core/worker.py
wk9874: For now I've left this as-is, because I couldn't think of a nicer way to refactor it... `get_actor` is a function which returns  the model actor for each worker task to use, either one which already exists, or a new one which have been created

wk9874:
> Actually I just spotted that i get the weights file dumped in the local folder where I ran toktagger? Surely it should be in a cache folder?

This was an error in the temp YOLO implementation rather than the actual loading code - `define_model` had `YOLO("yolov8n")`, which means that whenever the model class was instantiated, that ran, and YOLO automatically downloaded the weights to the CWD.

Have fixed it, can you double check it works now pls?

praksharma:
> > Actually I just spotted that i get the weights file dumped in the local folder where I ran toktagger? Surely it should be in a cache folder?
> 
> This was an error in the temp YOLO implementation rather than the actual loading code - `define_model` had `YOLO("yolov8n")`, which means that whenever the model class was instantiated, that ran, and YOLO automatically downloaded the weights to the CWD.
> 
> Have fixed it, can you double check it works now pls?

It looks to similar to https://github.com/ukaea/toktagger/pull/326#discussion_r3705472115. Not sure if you tested it on NVIDIA GPU.

samueljackson92:
Working nicely now. Just merged the dev to get the `uv.lock` conflict fixed. Should be good to go after the CI passes.

Issue #320, opened 2026-07-09
Loading UDA RCO Camera


samueljackson92:
**Describe the bug**
Loading MAST-U's `RCO` camera should already be supported. However, I get the following error:

```
Could not load image signal 'rco' for shot ID '54339': Cannot handle this data type: (1, 1, 3), <u2
```

Quite possibly this is something to do with it being a three channel image.

**To Reproduce**
1. Create a new project with uda_camera loader
2. Add sample `54339` for example
3. Try and open the shot

**Expected behavior**
Camera data should be opened and frame displayed

**Screenshots**

<img width="575" height="254" alt="Image" src="https://github.com/user-attachments/assets/acd906c4-4481-4c24-adb4-f009d94fa1ba" />

abdullah-ukaea:
This issue has been closed by https://github.com/ukaea/toktagger/pull/321

PR #321, opened 2026-07-09
Fix bug with rco camera loading in UDA camera data loader

samueljackson92:
This PR adds better edge case handling for UDA camera data and fixes the bug with loading RCO data.

To Test:
 - Follow the instructions in the #320 issue.

### DIFF
--- a/tests/api/core/test_data_loaders.py
+++ b/tests/api/core/test_data_loaders.py
@@ -18,6 +18,7 @@ from toktagger.api.schemas.data import (
 )
 import pathlib
 import numpy
+import xarray
 from PIL import Image
 import base64
 import io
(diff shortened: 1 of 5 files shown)


toktagger/api/core/data_loaders.py
```diff
@@ -414,6 +414,9 @@ def get_sample(
             if params.frame is None:
                 params.frame = 0  # Default to first frame if not specified
 
+            print(
```
abdullah-ukaea: can we remove this print statement now? I assume it was here for debugging purposes?

toktagger/api/core/data_loaders.py
```diff
@@ -423,6 +426,26 @@ def get_sample(
...
+            # Convert uint16 to uint8 for Pillow compatibility (Pillow doesn't support u2)
+            if image_array.dtype == np.uint16:
+                if np.any(image_array > 255):
+                    val_range = image_array.max() - image_array.min()
+                    image_array = image_array - image_array.min()
+                    if val_range:
+                        image_array = image_array / val_range
+                    image_array = (image_array * 255).astype(np.uint8)
+                else:
+                    image_array = image_array.astype(np.uint8)
```
abdullah-ukaea: This rescales every frame using that frame’s darkest and brightest pixels, so brightness may jump between frames. A frame containing one constant value above 255 would also become completely black. Could we instead use the camera’s declared bit depth for a consistent conversion for RCO, simply clip to 0–255 and convert to uint8?

I think the uda-xarray includes the camera’s bit depth in the data attributes
bit_depth = signal["data"].attrs["depth"]

For shot 54339, RCO reported depth = 8, even though NumPy stores the returned array as uint16. So the code can use that metadata to perform a consistent conversion to uint8.

abdullah-ukaea (commented):
Looks good, please address the 2 comments in the PR.

I also looked at  tests/api/core/test_data_loaders.py and would like to request one more change.

Could we add or extend a data-loader test to cover this regression? The existing UDA camera test only covers RBA’s 2-D uint8 output, whereas issue https://github.com/ukaea/toktagger/issues/320 was caused by RCO returning (1, height, width, 3) uint16 data.

abdullah-ukaea:
Also worth documenting to anyone testing this PR that there is a known bug if you click next very quickly it will break the video view (I will document this in a formal issue when this is merged into dev) that will be fixed in a follow up PR that is caused by the latency of this specific UDA camera source RCO that is not found on rba camera.

We have ideas on how to fix this.

toktagger/api/core/data_loaders.py
samueljackson92: Yes good point.

tests/api/core/test_data_loaders.py
```diff
@@ -165,6 +166,106 @@ def test_uda_camera_loader(uda_env_vars):
...
+        shot_id=54339,
+        data=uda_shot,
+        _id="test",
+        project_id="test",
+        validated_annotations=False,
+    )
+
+    raw_values = numpy.array([[0, 128], [200, 255]], dtype=numpy.uint16)
+    fake_dataset = xarray.Dataset(
+        {"data": xarray.DataArray(raw_values, dims=["x", "y"], attrs={"depth": 8})}
```
abdullah-ukaea, line 186: These tests only use 2-D images, but the original RCO error (documented in https://github.com/ukaea/toktagger/issues/320) happened with a three-channel uint16 image I think. That means the tests check the scaling change, but not the exact image format that previously failed.

Could we add one test using an RCO-like (1, height, width, 3) uint16 image with depth=8,and check that the returned PNG is a (height, width, 3) uint8 image?

abdullah-ukaea (commented):
The data-loader changes look good to me. I just have one final comment on the new tests:

Could we add a test for the three-channel uint16 image shape that caused the original RCO loading problem? 
@samueljackson92

tests/api/core/test_data_loaders.py
abdullah-ukaea: Here is an example of what I mean.

It mirrors the original RCO format: a leading time dimension plus RGB `uint16` pixels with `depth=8`.

```python
def test_uda_camera_loader_rco_rgb_uint16(monkeypatch):
    # RCO returns one time frame of three-channel uint16 image data.
    raw_values = numpy.array(
        [
            [
                [[0, 64, 128], [255, 200, 100]],
                [[10, 20, 30], [40, 50, 60]],
            ]
        ],
        dtype=numpy.uint16,
    )
    fake_dataset = xarray.Dataset(
        {
            "data": xarray.DataArray(
                raw_values,
                dims=["time", "height", "width", "channel"],
                attrs={"depth": 8},
            )
        }
    )
    monkeypatch.setattr(
        data_loaders.xr, "open_dataset", lambda *args, **kwargs: fake_dataset
    )

    sample = Sample(
        shot_id=54339,
        data=ShotData(protocol="uda", signal_names=["rco"]),
        _id="test",
        project_id="test",
        validated_annotations=False,
    )

    image_data = data_loaders.UDACameraDataLoader().get_sample(
        sample, params=ImageParams(name="image", frame=0)
    )
    result = numpy.array(Image.open(io.BytesIO(base64.b64decode(image_data.values))))

    assert result.shape == (2, 2, 3)
    assert result.dtype == numpy.uint8
    assert numpy.array_equal(result, raw_values.squeeze(axis=0).astype(numpy.uint8))
```

samueljackson92:
Updated.

abdullah-ukaea:
LGTM, this is merge ready now

abdullah-ukaea (approved):

PR #322, opened 2026-07-10
Video: xy point improvements 

abdullah-ukaea:
# Summary

Removes the `@annotorious/plugin-tools` dependency and replaces it with a small point editor we own, built for x,y keypoint annotation.

- New `point-editor/` module: a Svelte 5 editor component, a React overlay that draws all point markers, and shared marker style constants.
- Point markers are now drawn in screen pixels: a ring with a centre dot that stays the same size at any zoom, in both view and edit mode.
- Selected points are highlighted (blue ring).
- Crosshair cursor when the point tool is armed.
- Deletes the CSS hacks we used to hide the old editor's resize handles.
- Adds Svelte 5 to the build. Annotorious expects Svelte 4 style editor components and Svelte's own 4 to 5 compatibility helpers are deprecated, so a small adapter (`point-editor-host.svelte.ts`) presents our Svelte 5 editor to Annotorious in the shape it expects

# Demo

https://github.com/user-attachments/assets/4056243c-89b7-42f8-8fd5-d41f403fa32c


abdullah-ukaea:
For now, no code review is required. 

Please can @SteveAEA @jblake42 first functionally test this branch / use the new x,y point annotation. Identify any bugs, make sure the expected features are working.

abdullah-ukaea:
Not ready for code review

wk9874:
I'll leave this to @SteveAEA to functionality test properly, but I've had a quick go and it looks really good! Much better than before :) Couple things:

1. I find it quite hard to click an existing point when fully zoomed out - it looks like this might be because the 'hitbox' is only the very central dot, rather than the whole outer circle? Can this be improved?

https://github.com/user-attachments/assets/07c15dab-9d8a-4057-ae8b-146345db033e

2. Points are only selectable in edit mode, while other items are selectable in both modes
3. Points are rendered over the top of info box popups, which is different to other items
4. Sometimes points seems unselectable, even in edit mode

Example of points 2, 3, 4:

https://github.com/user-attachments/assets/5494f393-d3bb-4f7b-96d8-650193270ffe

(Side note related to 2 - I've just noticed that bounding boxes & polygons can be deleted in View mode. That probably shouldn't be the case - I think the ideal scenario is that the pop up still appears in View mode if the item is clicked, but the delete button is disabled)

abdullah-ukaea:
New updates will come shortly to this PR after some other video PRs are merged to dev.

PR #325, opened 2026-07-14
Adds optional raw image support to ImageDataLoader

praksharma:
`ImageDataLoader` currently returns images as base64-encoded PNGs. This introduces additional overhead for ML training, where images are immediately decoded back into `PIL.Image` objects.

This PR adds an optional `return_raw` flag to ImageParams. When enabled, the loader returns a `RawImageData` containing a `PIL.Image` instead of a base64 string. `return_raw=False` remains the default, so Existing API calls require no modifications.

A couple of questions:
* Consider using a dataclass for `RawImageData`?
* Should `RawImageData` inherit from `Data`? Coz Pydantic doesn't know how to validate a `PIL.Image.Image`.

### DIFF
--- a/toktagger/api/schemas/data.py
+++ b/toktagger/api/schemas/data.py
@@ -38,9 +38,8 @@ class DataParams(ConfiguredModel):
 class ImageParams(DataParams):
     name: Literal["image"] = "image"
     frame: int | None
-    return_raw: bool = (
-        False  # Optional: Return raw encoded image bytes instead of base64.
-    )
+    # Optional: Return raw encoded image bytes instead of base64.
+    return_raw: bool = False
 
 
 DataResponseType = Union[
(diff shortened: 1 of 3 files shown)


toktagger/api/schemas/data.py
samueljackson92: Change this to `str | list[int]` and inherit from `Data`

samueljackson92 (changes requested):

toktagger/api/schemas/data.py
praksharma: I've made the changes.

samueljackson92 (changes requested):
Looks good, but one more thing: can you add a quick test for this in `test_data_loaders.py`

samueljackson92 (approved):

Issue #225, opened 2026-04-13
Image Annotation: Support 'frame' labelling


wk9874:
**Is your feature request related to a problem? Please describe.**
For some image labelling cases, you may not want to annotate something with a bounding box, but rather identify the whole frame as showing something. Eg if I wanted to label 'no UFO'...

**Describe the solution you'd like**
There should be an option to disable bounding box drawing, and instead just associate a frame number with a class label. Eg could be a new option in the taskbar on the right hand side with a tag icon, and if you click anywhere on the image with that selected it labels the whole frame as a given class label. Once annotated, clicking anywhere on the image should probably bring up the information box, which just tells you the class label that frame has. Eg:

<img width="2000" height="993" alt="Image" src="https://github.com/user-attachments/assets/8eea55ea-4bc2-45f0-84e1-ca5db016fa66" />

**Describe alternatives you've considered**
Alternatively could be a whole different Task. Although it might be useful in some cases to have the option within a single sample - eg I might want to annotate a UFO with a bounding box, and then a non-UFO frame without one?

PR #326, opened 2026-07-17
Add Ultralytics video detection models

praksharma:
Adds the support for [Ultralytics](https://docs.ultralytics.com/quickstart)-based object detection for video data.

The implementation uses TokTagger’s native data loader and an in-memory dataset, avoiding Ultralytics’ required on-disk dataset structure.

## Currently implemented

- YOLO video object-detection training and prediction
- Supported pretrained checkpoints:
  - YOLOv8n
  - YOLO11n
  - YOLO26n
  - YOLO26x
- YOLO26 P2 architecture for small-object detection
  - Currently trained from random initialization because a compatible pretrained P2 checkpoint is not available

## TODO
- RT-DETR x and l models
- Unit tests

## Changes

- Adds Ultralytics under `optional-dependencies` in `pyproject.toml`.
- Registers the YOLO video-detection models with TokTagger.
- Adds the implementation under `toktagger/api/models/ultralytics_detection`.

## Model storage

```text
models/
├── pretrained/
│   └── ultralytics/
│       ├── yolo/
│       │   └── <pretrained_checkpoint>.pt
│       └── rtdetr/
│           └── <pretrained_checkpoint>.pt
└── <model_id>/
           ├── args.yaml
           ├── results.csv
           └── weights/
               ├── best.pt
               └── last.pt
```

## Follow-up
~~TokTagger tracks whether a model is usable through the Ray actor’s `_trained` flag. The generic worker normally restores this state by finding a `<model_id>` checkpoint and calling `wrapped_load()`.~~

~~Ultralytics checkpoints use a nested project/model directory instead, so the Ultralytics actor currently finds its checkpoint and restores the trained state itself. Should we think about a more generic checkpoint-discovery hook?~~

Addressed in: https://github.com/ukaea/toktagger/pull/346

## Training-data assumption

This initial implementation assumes that every frame in a validated video sample has been reviewed. Frames without validated bounding boxes are included as negative examples.

Whole-frame labels proposed in #225 should eventually allow explicitly reviewed negative frames to be distinguished from unreviewed frames.

More discussion on this topic: https://github.com/ukaea/toktagger/pull/326#discussion_r3705374823


toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,299 @@
...
+    model = YOLO("yolo26n.pt")
+    b64_string = image_string
+    img_bytes = base64.b64decode(b64_string)
+    img = Image.open(BytesIO(img_bytes)).convert("RGB")
+
+    results = model(img)
+    print(results)
+
+
+# def build_video_frame_manifest(
```
samueljackson92: Please remove commented out code.

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,299 @@
...
+                    "frame": frame_image.frame,
+                    "image": frame_image.values
+                })
+
+            except FileNotFoundError:
+                # print("Can't access the sample. Check with toktagger devs. Read the following error.")
+                print("----------------------")
+                print("End of frames I guess")
+                print("last frame was: ", frame_image.frame)
+                print("----------------------")
```
samueljackson92: Remove print statements. Throw and exception or warnings instead.

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,299 @@
...
+    ) -> float:
+        self.log_progress(training_status="started")
+        if params is None:
+            params = DebugVideoTrainParams()
+
+        print("################ DEBUG VIDEO GET_SAMPLE TRAIN ################")
+        print("num samples:", len(samples))
+        print("Num annotations: ", len(annotations))
+        print("data_loader:", type(self.data_loader), self.data_loader)
+        print("print params: ", params)
```
samueljackson92: Again, remove print statements. Perhaps use debug logging instead?

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,299 @@
...
+            try:
+                # give individual frames
+                # unless we put it in a while True loop we get the first frame of the shot
+                data = self.data_loader.get_sample(
+                    sample,
+                    ImageParams(name="image", frame=frame),
+                )
+
+                # print("get_sample returned:", type(data))
+                # print("sample_debug type:", type(data.values))
```
samueljackson92: Remove commented out code.

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
 models = [
   "torch",
   "ray>=2.49.1",
+  "ultralytics"
```
samueljackson92: 👍

samueljackson92 (changes requested):
Broadly looks good. Please also fix the `ruff` linting warnings.

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+
+        # Validation is intentionally postponed, so no meaningful score is
+        # available yet.
+        return 0.0
+
+    def save(self, file_stem: str) -> None:
+        """Verify that Ultralytics persisted this model's weights.
+        Ultralytics trainer handles the saving.
+        """
+        return None
```
abdullah-ukaea, line 496: Could we save the model weights here? TokTagger calls this function after loading a model so that the weights are available after the Ray actor restarts. Currently, the function returns without saving anything, so a loaded model may stop working after its actor is restarted.

**To be specific:** TokTagger’s load-model workflow calls wrapped_load() followed by wrapped_save() so the weights survive a Ray actor restart. With this implementation, locally loaded YOLO weights only exist at their original path/in the current actor. A recreated actor searches the nested best.pt/last.pt directory and may find nothing, even though MongoDB marks the model as completed.

Please copy _trained_weights_path into this model’s permanent weights directory here, or ,add a generic checkpoint-persistence hook that supports the Ultralytics layout.

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,448 @@
...
+        sample_record_count = 0
+
+        for frame_image in iter_sample_frames(
+            data_loader,
+            sample,
+        ):
+            if isinstance(frame_image.values, str):
+                raise TypeError(
+                    "Expected raw image bytes but received a base64 string."
+                )
```
abdullah-ukaea, line 164: Some video loaders, including ArrayDataLoader and UDACameraDataLoader, return a base64 string even when return_raw=True. This check therefore makes YOLO training and prediction fail for those loaders.

Could we support both raw bytes and base64 images here? Otherwise, we should clearly restrict this model to loaders that support raw image bytes.

**Here is an idea to support both image formats in this model below.**

The model could decode base64 values as well as accepting raw bytes:

```python
if isinstance(frame_image.values, str):
    image_bytes = base64.b64decode(frame_image.values)
else:
    image_bytes = bytes(frame_image.values)
```

This would allow images returned by `ArrayDataLoader` and `UDACameraDataLoader` to reach YOLO.

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,448 @@
...
+            frame_image = data_loader.get_sample(
+                sample,
+                ImageParams(
+                    name="image",
+                    frame=frame_image.frame + 1,
+                    return_raw=True,
+                ),
+            )
+        except FileNotFoundError:
+            break
```
abdullah-ukaea, line 118: This frame loop currently reaches the end of a video cleanly only when using `ImageDataLoader`, because it expects the loader to raise `FileNotFoundError`. 

I was looking into the other valid video loaders and found they behave differently: 
* `ArrayDataLoader` raises `DataLoaderError` for an out-of-range frame
* `UDACameraDataLoader` wraps its loading failures in `DataLoaderError`. 

As a result, YOLO training or prediction can fail for other data loaders instead of stopping after the final frame.

**Below is an idea to give all video loaders the same end-of-video signal.**

Simply catching every `DataLoaderError` would not be safe:

```python
except DataLoaderError:
    break
```

`DataLoaderError` can also mean that a NumPy file is invalid, the data has the wrong shape, or an external data source failed. Catching it here would hide those real errors and incorrectly treat them as the end of the video.

One option would be to introduce a more specific exception in the data-loader layer:

```python
class FrameNotFoundError(DataLoaderError):
    """The requested video frame does not exist."""
```

Each video loader would raise `FrameNotFoundError` only when the requested frame is unavailable. The YOLO iterator could then stop safely without hiding other failures:

```python
except FrameNotFoundError:
    break
```

This approach would require changes in:

- `toktagger/api/core/data_loaders.py`
- `toktagger/api/models/ultralytics_detection/video_detection.py`
- `toktagger/api/models/temp.py`

That shared loader change may be outside the scope of this PR to be fair. If this PR intentionally supports only `ImageDataLoader`, could we validate that explicitly, return a clear error for unsupported loaders and put this in the PR description? A follow up PR could then introduce the common frame @wk9874, what do you think? or should we do it here.

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+        """Return the device assigned to this model actor."""
+        if self.gpu_available() and torch.cuda.is_available():
+            return torch.device("cuda")
+
+        return get_torch_device()
+
+    def define_model(self) -> str:
+        """Resolve the default pretrained checkpoint."""
+        model_path = check_pretrained_model_availability(self.model_name)
+        return str(model_path)
```
abdullah-ukaea, line 353: This downloads the default yolo26n.pt whenever the actor is created, before checking whether trained weights already exist. It also downloads YOLO26n when the user selected a different checkpoint.

Could we only download the selected pretrained checkpoint when new training starts? Existing trained models should be restored without needing this default download or an internet connection.

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,448 @@
...
+    def define_model(self) -> str:
+        """Use the P2 architecture distributed with Ultralytics."""
+        return self.model_name
+
+    def get_training_model(
+        self,
+        params: pydantic.BaseModel,
+    ) -> str:
+        """Always train the P2 architecture instead of a pretrained variant."""
+        return self.model_name
```
abdullah-ukaea, line 448: Ah so hear we need to fix something related to the UI training form that we provide to the user. Remember in your PR description you state _"YOLO26 P2 architecture for small-object detection Currently trained from random initialization because a compatible pretrained P2 checkpoint is not available"_ but the user can get confused currently:

When yolo_ufo_p2 is selected, the form lets the user choose a pretrained checkpoint, but the P2 model ignores that choice and always trains yolo26-p2.yaml from scratch.

Since there is currently no compatible pretrained P2 checkpoint, could P2 use a separate training parameter schema without yolo_size? The generated UI would then stop showing a checkpoint option that has no effect.

Image

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,448 @@
...
+                    # ImageData stores raw encoded bytes as a JSON-compatible
+                    # list of integers. Convert it back into bytes here.
+                    "image": bytes(frame_image.values),
+                    "boxes": boxes,
+                    "classes": classes,
+                    "labels": labels,
+                    "track_ids": track_ids,
+                }
+            )
+            sample_record_count += 1
```
abdullah-ukaea, line 210: This is acceptable for the initial implementation, but I think we should document an important assumption here. **No code changes are required, just think about this and maybe add a comment in the PR description.**

The manifest adds every frame from a validated sample. Frames without bounding boxes are passed to YOLO as negative examples. However, validation currently applies to the whole sample, and in practice an annotator may validate a sample without reviewing every frame. In that case, an unreviewed frame containing an object could incorrectly be treated as background.
 
Are we happy to make the assumption that every frame in a validated sample has been reviewed? If not, a simple short-term alternative would be to include only frames containing validated bounding boxes and discard the remaining frames. The trade-off is that the model would not receive any empty frames as negative examples.

Longer term, we plan to address this through whole-frame labels in https://github.com/ukaea/toktagger/issues/225 as @wk9874  suggested. That would allow annotators to explicitly label frames as, for example, “no UFO”, so the model could distinguish confirmed negative frames from unreviewed frames.

No change is required for this initial PR, but just wanted to make you aware Prakhar, maybe add a small comment at the end of the pr description so it is documented somewhere?

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+
+from .utils import (
+    check_pretrained_model_availability,
+    get_toktagger_cache_dir,
+    get_torch_device,
+)
+
+logger = logging.getLogger(__name__)
+
+DetectionRecord = dict[str, Any]
```
abdullah-ukaea, line 32: I think we should use stricter typing here. 

Could we define DetectionRecord as a TypedDict instead of dict[str, Any]?

Each record is expected to contain specific fields such as image, boxes, classes, shot_id, and frame. Using Any means missing fields or incorrect value types will only fail later while the Ray worker is running.

```python
class DetectionRecord(TypedDict):
    shot_id: int
    frame: int
    image: bytes
    boxes: list[list[float]]
    classes: list[int]
```
  
This would document the expected record structure and allow type checking to catch mistakes earlier. We could also remove labels and track_ids if they are not used by the dataset.

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+            "name": self.id,
+            "exist_ok": True,
+            "save": True,
+            "plots": False,
+            "val": has_validation_data,
+            "close_mosaic": 0,
+        }
+        if learning_rate > 0:
+            overrides["lr0"] = learning_rate
+        return overrides
```
abdullah-ukaea, line 414: Could you do me a favour and try GPU training and prediction, then check git status? After GPU training, I consistently get an untracked yolo26n.pt in the TokTagger repository root:

```text
Untracked files:
    yolo26n.pt
```

It looks like Ultralytics’ AMP check downloads a second copy instead of using the checkpoint already in TokTagger’s cache. Do you see the same behaviour?
  
If so, could we configure Ultralytics to reuse the pretrained checkpoint under MODEL_STORAGE so weight files are not created in the source directory?

Although I comment on this override block the main code block responsible is below on line 461 I think.

```python
trainer = ToktaggerDetectionTrainer(
    overrides=overrides,
    train_dataset=train_dataset,
    val_dataset=validation_dataset,
    class_names=self.class_names,
    progress_callback=self.log_progress,
)
trainer.train()
```

abdullah-ukaea:
**Non-blocking functional feedback:** 

YOLO training currently produces a large amount of terminal output. Would it be worth making Ultralytics quieter by default and keeping the detailed logs behind a verbose/debug setting?

The full output is useful for debugging, but most users will probably monitor training through the UI and only need key progress updates, warnings, and errors in the terminal.

**example of terminal output**

```text
(YoloVideoDetectionP2Model pid=3904) Ultralytics 8.4.98 🚀 Python-3.12.3 torch-2.13.0+cu130 CUDA:0 (NVIDIA GeForce RTX 4090 Laptop GPU, 16376MiB)
INFO:     127.0.0.1:47698 - "GET /projects/6a6fba37aad4c943840dd3f2/models HTTP/1.1" 200 OK
(YoloVideoDetectionP2Model pid=3904) engine/trainer: agnostic_nms=False, amp=True, angle=1.0, augment=False, auto_augment=randaugment, batch=2, bgr=0.0, box=7.5, cache=False, cfg=None, classes=None, close_mosaic=0, cls=0.5, cls_pw=0.0, cls_remap=True, compile=False, conf=None, copy_paste=0.0, copy_paste_mode=flip, cos_lr=False, cutmix=0.0, data=None, degrees=0.0, deterministic=True, device=0, dfl=1.5, dis=6.0, distill_model=None, dnn=False, dropout=0.0, dynamic=False, embed=None, end2end=None, epochs=2, erasing=0.4, exist_ok=True, fliplr=0.5, flipud=0.0, format=torchscript, fraction=1.0, freeze=None, hsv_h=0.015, hsv_s=0.7, hsv_v=0.4, imgsz=1024, iou=0.7, keras=False, kobj=1.0, line_width=None, lr0=0.1, lrf=0.01, mask_ratio=4, max_det=300, mixup=0.0, mode=train, model=yolo26-p2.yaml, momentum=0.937, mosaic=1.0, multi_scale=0.0, name=6a70ac4bafed39bb42331724, nbs=64, nms=False, opset=None, optimize=False, optimizer=auto, overlap_mask=True, patience=100, perspective=0.0, plots=False, pose=12.0, pretrained=True, profile=False, project=/home/gs8173/.cache/toktagger/models/6a6fba37aad4c943840dd3f2/ultralytics/yolo, quantize=None, rect=False, resume=False, retina_masks=False, rle=1.0, save=True, save_conf=False, save_crop=False, save_dir=/home/gs8173/.cache/toktagger/models/6a6fba37aad4c943840dd3f2/ultralytics/yolo/6a70ac4bafed39bb42331724, save_frames=False, save_json=False, save_period=-1, save_txt=False, scale=0.5, seed=0, shear=0.0, show=False, show_boxes=True, show_conf=True, show_labels=True, simplify=True, single_cls=False, source=None, split=val, stream_buffer=False, task=detect, time=None, tracker=tracktrack.yaml, translate=0.1, val=False, verbose=True, vid_stride=1, visualize=False, warmup_bias_lr=0.1, warmup_epochs=3.0, warmup_momentum=0.8, weight_decay=0.0005, workers=0, workspace=None
(YoloVideoDetectionP2Model pid=3904) Overriding model.yaml nc=80 with nc=1
(YoloVideoDetectionP2Model pid=3904) WARNING ⚠️ no model scale passed. Assuming scale='n'.
(YoloVideoDetectionP2Model pid=3904)
(YoloVideoDetectionP2Model pid=3904)                    from  n    params  module                                       arguments
(YoloVideoDetectionP2Model pid=3904)   0                  -1  1       464  ultralytics.nn.modules.conv.Conv             [3, 16, 3, 2]
(YoloVideoDetectionP2Model pid=3904)   1                  -1  1      4672  ultralytics.nn.modules.conv.Conv             [16, 32, 3, 2]
(YoloVideoDetectionP2Model pid=3904)   2                  -1  1      6640  ultralytics.nn.modules.block.C3k2            [32, 64, 1, False, 0.25]
(YoloVideoDetectionP2Model pid=3904)   3                  -1  1     36992  ultralytics.nn.modules.conv.Conv             [64, 64, 3, 2]
(YoloVideoDetectionP2Model pid=3904)   4                  -1  1     26080  ultralytics.nn.modules.block.C3k2            [64, 128, 1, False, 0.25]
(YoloVideoDetectionP2Model pid=3904)   5                  -1  1    147712  ultralytics.nn.modules.conv.Conv             [128, 128, 3, 2]
(YoloVideoDetectionP2Model pid=3904)   6                  -1  1     87040  ultralytics.nn.modules.block.C3k2            [128, 128, 1, True]
(YoloVideoDetectionP2Model pid=3904)   7                  -1  1    295424  ultralytics.nn.modules.conv.Conv             [128, 256, 3, 2]
(YoloVideoDetectionP2Model pid=3904)   8                  -1  1    346112  ultralytics.nn.modules.block.C3k2            [256, 256, 1, True]
(YoloVideoDetectionP2Model pid=3904)   9                  -1  1    164608  ultralytics.nn.modules.block.SPPF            [256, 256, 5, 3, True]
(YoloVideoDetectionP2Model pid=3904)  10                  -1  1    249728  ultralytics.nn.modules.block.C2PSA           [256, 256, 1]
(YoloVideoDetectionP2Model pid=3904)  11                  -1  1         0  torch.nn.modules.upsampling.Upsample         [None, 2, 'nearest']
(YoloVideoDetectionP2Model pid=3904)  12             [-1, 6]  1         0  ultralytics.nn.modules.conv.Concat           [1]
(YoloVideoDetectionP2Model pid=3904)  13                  -1  1    119808  ultralytics.nn.modules.block.C3k2            [384, 128, 1, True]
(YoloVideoDetectionP2Model pid=3904)  14                  -1  1         0  torch.nn.modules.upsampling.Upsample         [None, 2, 'nearest']
(YoloVideoDetectionP2Model pid=3904)  15             [-1, 4]  1         0  ultralytics.nn.modules.conv.Concat           [1]
(YoloVideoDetectionP2Model pid=3904)  16                  -1  1     34304  ultralytics.nn.modules.block.C3k2            [256, 64, 1, True]
(YoloVideoDetectionP2Model pid=3904)  17                  -1  1         0  torch.nn.modules.upsampling.Upsample         [None, 2, 'nearest']
(YoloVideoDetectionP2Model pid=3904)  18             [-1, 2]  1         0  ultralytics.nn.modules.conv.Concat           [1]
(YoloVideoDetectionP2Model pid=3904)  19                  -1  1      8704  ultralytics.nn.modules.block.C3k2            [128, 32, 1, True]
(YoloVideoDetectionP2Model pid=3904)  20                  -1  1      9280  ultralytics.nn.modules.conv.Conv             [32, 32, 3, 2]
(YoloVideoDetectionP2Model pid=3904)  21            [-1, 16]  1         0  ultralytics.nn.modules.conv.Concat           [1]
(YoloVideoDetectionP2Model pid=3904)  22                  -1  1     24064  ultralytics.nn.modules.block.C3k2            [96, 64, 1, True]
(YoloVideoDetectionP2Model pid=3904)  23                  -1  1     36992  ultralytics.nn.modules.conv.Conv             [64, 64, 3, 2]
(YoloVideoDetectionP2Model pid=3904)  24            [-1, 13]  1         0  ultralytics.nn.modules.conv.Concat           [1]
(YoloVideoDetectionP2Model pid=3904)  25                  -1  1     95232  ultralytics.nn.modules.block.C3k2            [192, 128, 1, True]
(YoloVideoDetectionP2Model pid=3904)  26                  -1  1    147712  ultralytics.nn.modules.conv.Conv             [128, 128, 3, 2]
(YoloVideoDetectionP2Model pid=3904)  27            [-1, 10]  1         0  ultralytics.nn.modules.conv.Concat           [1]
(YoloVideoDetectionP2Model pid=3904)  28                  -1  1    463104  ultralytics.nn.modules.block.C3k2            [384, 256, 1, True, 0.5, True]
(YoloVideoDetectionP2Model pid=3904)  29    [19, 22, 25, 28]  1    211304  ultralytics.nn.modules.head.Detect           [1, 1, True, [32, 64, 128, 256]]
(YoloVideoDetectionP2Model pid=3904) YOLO26-p2 summary: 329 layers, 2,515,976 parameters, 2,515,976 gradients, 7.5 GFLOPs
(YoloVideoDetectionP2Model pid=3904)
(YoloVideoDetectionP2Model pid=3904) AMP: running Automatic Mixed Precision (AMP) checks...
(YoloVideoDetectionP2Model pid=3904) AMP: checks passed ✅
(YoloVideoDetectionP2Model pid=3904) optimizer: 'optimizer=auto' found, ignoring 'lr0=0.1' and 'momentum=0.937' and determining best 'optimizer', 'lr0' and 'momentum' automatically...
(YoloVideoDetectionP2Model pid=3904) optimizer: AdamW(lr=0.002, momentum=0.9) with parameter groups 145 weight(decay=0.0), 161 weight(decay=0.0005), 161 bias(decay=0.0)
(YoloVideoDetectionP2Model pid=3904) Image sizes 1024 train, 1024 val
(YoloVideoDetectionP2Model pid=3904) Using 0 dataloader workers
(YoloVideoDetectionP2Model pid=3904) Logging results to /home/gs8173/.cache/toktagger/models/6a6fba37aad4c943840dd3f2/ultralytics/yolo/6a70ac4bafed39bb42331724
(YoloVideoDetectionP2Model pid=3904) Starting training for 2 epochs...
(YoloVideoDetectionP2Model pid=3904)
(YoloVideoDetectionP2Model pid=3904)       Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
INFO:     127.0.0.1:36742 - "GET /projects/6a6fba37aad4c943840dd3f2/models HTTP/1.1" 200 OK
        1/2      1.14G          0      68.36          0          0       1024: 0% ──────────── 0/3  3.3s
        1/2       1.2G          0      69.09          0          0       1024: 33% ━━━━──────── 1/3 1.1s/it 3.6s<2.1s
        1/2       1.2G     0.6244      69.61  0.0009133          1       1024: 66% ━━━━━━━━─        1/2       1.2G     0.6244      69.61  0.0009133          1       1024: 100% ━━━━━━━━━━━━ 3/3 1.4s/it 4.3s
INFO:     127.0.0.1:36752 - "PUT /projects/6a6fba37aad4c943840dd3f2/models/6a70ac4bafed39bb42331724 HTTP/1.1" 200 OK
(YoloVideoDetectionP2Model pid=3904)
(YoloVideoDetectionP2Model pid=3904)       Epoch    GPU_mem   box_loss   cls_loss   dfl_loss  Instances       Size
        2/2      1.21G          0      67.54          0          0       1024: 0% ──────────── 0/3  0.1s
        2/2      1.23G     0.8771      70.25   0.001202          1       1024: 33% ━━━━─────        2/2      1.23G     0.5847      70.18  0.0008012          0       1024: 100% ━━━━━━━━━━━━ 3/3 9.4it/s 0.3s
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 50% ━━━━━━────── 1/2 3.3s/it 1.0s<3.3s
                 Class     Images  Instances      Box(P          R      mAP50  mAP50-95): 100% ━━━━━━━━━━━━ 2/2 1.9it/s 1.0s
(YoloVideoDetectionP2Model pid=3904)                    all          6          1          0          0          0          0
INFO:     127.0.0.1:36756 - "PUT /projects/6a6fba37aad4c943840dd3f2/models/6a70ac4bafed39bb42331724 HTTP/1.1" 200 OK
(YoloVideoDetectionP2Model pid=3904)
(YoloVideoDetectionP2Model pid=3904) 2 epochs completed in 0.002 hours.
```

abdullah-ukaea:
To be clear this is a very good PR ! @praksharma Training and prediction works well with no errors. Just some bugs, edge cases and other thoughts I had to get the yolo training to be more robust or user friendly.

toktagger/api/models/ultralytics_detection/video_detection.py
wk9874: Instead of that, I would just make sure all image-based data loaders have the `return_raw` option implemented...

Potentially we should make another image dataloader base class which implements that logic, which each of `ImageDataLoader`, `ArrayDataLoader`, `UDACameraDataLoader` can inherit from

toktagger/api/models/ultralytics_detection/video_detection.py
wk9874: I agree with implementing `FrameNotFoundError`, probably worth doing it either as part of this PR, or as a separate small PR which is merged before this one

toktagger/api/models/ultralytics_detection/base.py
wk9874: Agree, but could use a Pydantic `BaseModel`, since that's what the rest of the backend uses everywhere :)

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+
+# Callable to pass functions in another function
+from collections.abc import Callable
+from ultralytics.utils import RANK
+
+from toktagger.api.models.base import Model
+from toktagger.api.schemas.annotations import Annotation, AnnotationBase
+from toktagger.api.schemas.samples import Sample
+
+from .utils import (
```
wk9874, line 24: This should probably be a full import path, eg

`from toktagger.api.models.ultralytics_detection.utils import ...`

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+            raise ValueError(
+                f"Could not decode shot {record['shot_id']} frame {record['frame']}"
+            )
+
+        # Ultralytics trains its pretrained models using RGB channel ordering.
+        image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+        original_height, original_width = image.shape[:2]
+
+        letterboxed_image = self.letterbox(image=image)
+        if not isinstance(letterboxed_image, np.ndarray):
```
wk9874, line 76: I don't think this should be necessary, since you are never providing `labels`. I'd just directly type `letterboxed_image` as an array and get rid of this if

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+        # not create that duplicate flat file.
+        #
+        # When Ray recreates this actor, the generic worker will not call
+        # wrapped_load() because it cannot discover the nested checkpoint.
+        # Therefore this Ultralytics-specific actor discovers and loads its own
+        # checkpoint here. Model.wrapped_predict() refuses prediction unless
+        # _trained is True, so it must be restored after a valid checkpoint has
+        # been loaded successfully.
+        if trained_weights is not None:
+            self.load(str(trained_weights))
```
wk9874, line 312: Yeah I'm not a fan of this, would prefer the save and load methods to be implemented as the base class expects.

This PR may help with that? https://github.com/ukaea/toktagger/pull/346

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,507 @@
...
+        }
+        if learning_rate > 0:
+            overrides["lr0"] = learning_rate
+        return overrides
+
+    def train(
+        self,
+        samples: list[Sample],
+        annotations: list[list[Annotation]],
+        params: pydantic.BaseModel,
```
wk9874, line 420: Would be good to narrow the typing here

wk9874 (changes requested):

wk9874:
@praksharma the weights saving / loading has now changed in the models base class & worker, which should hopefully make your life easier (but will definitely require changes!)

Each model is now given a directory named with its model ID, and you can save any number of files in there. There is an optional `weights_filename` parameter passed to `load()`, which should be used if available. This points to the actual weights file to load into the model, and is used for loading pretrained weights into the UI (eg, if I pointed to `/my/user/dir/yolo_weights.pt`, your model's `load()` method would be provided `/my/user/dir` as the results dir and `yolo_weights.pt` to load in

If `weights_filename` is not provided to `load`, there should be some defined fallback behaviour. Eg in your case, it would likely be to load with the file at `results_dir.joinpath("best.pt")` if it exists, else `results_dir.joinpath("last.pt")`

See here for more details: https://github.com/ukaea/toktagger/pull/346

toktagger/api/models/ultralytics_detection/video_detection.py
praksharma: This has been fixed. Now the `YoloVideoDetectionP2Model` class:
*  Builds the P2 architecture and transfer the compatible pretrained weights, depending on `Yolo Size` selected in the UI.
*  New P2-specific layers are randomly initialised.

toktagger/api/models/ultralytics_detection/video_detection.py
praksharma: Good point. I’ve documented this in the PR description.

toktagger/api/models/ultralytics_detection/base.py
praksharma: Thanks to copilot, I was able to find the culprit in thousands of lines of code. For GPU training there is a list of checks being carried out. And this specific line is the culprit:

https://github.com/ultralytics/ultralytics/blob/b3bcdaf963957cd84d373c16abf2c6d11ce04654/ultralytics/utils/checks.py#L1030

INterestingly, Ultralytics has two different settings. `overrides` are per-training-run arguments and `settings` are global Ultralytics configuration, stored in a temp file `settings.json`.

https://github.com/ultralytics/ultralytics/blob/89cb7da4ee647e7247cc2ce4518403d045d94892/tests/test_cli.py#L37

All the settings are defined here:

https://github.com/ultralytics/ultralytics/blob/89cb7da4ee647e7247cc2ce4518403d045d94892/ultralytics/utils/__init__.py#L1357

More specifically, I think we just change the `weights_dir` in the `__init__` method of that class and I hope it works. Unfortunately, I don't have a laptop with GPU to test this. So, I am not sure if the nano model will be created in every `model_id` folder for AMP checks after this commit https://github.com/ukaea/toktagger/pull/326/commits/7036a2888cdfc319a34a8716bcc755946e0f0bf5.

praksharma:
The PR is ready for a review. I have created a [new branch](https://github.com/ukaea/toktagger/tree/standardise-image-based-data-loaders-return-raw-flag) (will create a PR once I commit something) to standardise the image-based dataloaders `return_raw` flag.

abdullah-ukaea:
I understand now that this selection is valid for P2 in this training form. It controls the P2 model size and selects the regular checkpoint used to initialize compatible layers.
 
There are two ways we could make this clearer:

- Simple fix: Just update the existing field description to:  “Pretrained YOLO checkpoint to fine-tune. For P2 models, compatible weights are transferred from this checkpoint, while new P2-specific layers are initialized randomly.” only a docstring change, no code change.
-  Ideal fix: Give the P2 model its own training parameter schema. This would allow the  P2 form to offer only compatible model choices and use a clearer field name such as “Base YOLO checkpoint,” with a P2-specific explanation. (This is related to a comment I made below in the second code review https://github.com/ukaea/toktagger/pull/326#discussion_r3769926062 ) this would involve a minor code change.

Could we also update the PR description? It currently says the P2 model trains from random initialization, but the latest implementation transfers compatible weights from the selected checkpoint.

regular yolo training form

yoloP2 training form

pyproject.toml
abdullah-ukaea, line 1: this is good, but we should add the ultralytics to the uv lock file so users using uv can also have this dependency

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+
+        return all_predictions
+
+
+@ModelRegistry.register(
+    "yolo_ufo_p2",
+    ["video"],
+    YoloTrainParams,
+    YoloPredictParams,
+)
```
abdullah-ukaea, line 521: Was YOLO11n intentionally removed? I suspect it was left out because there is no matching P2 architecture, but that should not prevent users from selecting it for the regular YOLO model.

The normal YOLO and P2 models currently share YoloTrainParams, so they must expose the same checkpoint options. It looks like yolo11n.pt was removed because it has no matching P2 architecture, but this also removes YOLO11n from the normal yolo_ufo form even though the PR description says it is supported.

Could we give P2 its own training parameter schema? The normal schema could continue to offer YOLO11n, while the P2 schema would offer only checkpoints with matching P2 architectures. This would also let us give the P2 checkpoint field the clearer description discussed above. If you disagree and there is a good reason to omit yolo11n and keep the current schema that is fine, at least update the pr description.

other related code:

```python
YoloModelName = Literal[
    "yolov8n.pt",
    "yolo26n.pt",
    "yolo26m.pt",
    "yolo26l.pt",
    "yolo26x.pt",
]
```
..
..
```python
@ModelRegistry.register(
    "yolo_ufo",
    ["video"],
    YoloTrainParams,
    YoloPredictParams,
)
```

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+        ge=1,
+        le=100,
+        description="Maximum number of detections per frame.",
+    )
+
+
+def iter_sample_frames(
+    data_loader: TokTaggerDataLoader,
+    sample: Sample,
+) -> Iterator[ImageData]:
```
abdullah-ukaea, line 143: Add a basic unit test for this?  see this comment for more detail https://github.com/ukaea/toktagger/pull/326#discussion_r3770046488

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+        except FileNotFoundError:
+            break
+
+
+def build_video_frame_manifest(
+    samples: list[Sample],
+    annotations: list[list[Annotation]],
+    class_map: dict[str, int],
+    data_loader: TokTaggerDataLoader,
+) -> list[DetectionRecord]:
```
abdullah-ukaea, line 202: Add a basic unit test for this?  see this comment for more detail https://github.com/ukaea/toktagger/pull/326#discussion_r3770046488

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+        "Created a video manifest containing %s frames.",
+        len(frame_manifest),
+    )
+
+    return frame_manifest
+
+
+def decode_frame_image(frame_image: ImageData) -> np.ndarray:
+    """Decode raw TokTagger image bytes for Ultralytics prediction."""
+    if isinstance(frame_image.values, str):
```
abdullah-ukaea, line 297: Add a basic unit test for this?   see this comment for more detail https://github.com/ukaea/toktagger/pull/326#discussion_r3770046488

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+
+        self._prediction_model = YOLO(str(weights_path))
+        self._trained_weights_path = weights_path
+
+    def predict(
+        self,
+        samples: list[Sample],
+        params: YoloPredictParams,
+        data_params=None,
+    ) -> list[list[AnnotationBase]]:
```
abdullah-ukaea, line 382: Add a basic unit test for this?  see this comment for more detail https://github.com/ukaea/toktagger/pull/326#discussion_r3770046488

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea, line 1: The functions below needs some test coverage:
- iter_sample_frames()
- build_video_frame_manifest()
- decode_frame_image
- predict()

Could we add some focused unit tests for this new functionality? The PR currently has no tests for frame iteration, manifest creation, image decoding or converting YOLO results into bounding-box annotations.

 These tests should not need to train a real model. Small test images and mocked data loaders/YOLO predictions could cover:
 - Reaching the end of a video
 - Frames with no annotations
 - Multiple boxes on one frame
 - Bounding-box coordinate conversion
 - Predictions containing no detections
 - Loading best.pt, with fallback to last.pt
 - Restoring a model after its Ray actor is recreated

toktagger/api/models/ultralytics_detection/utils.py
```diff
@@ -0,0 +1,153 @@
...
+    # partial checkpoint at the final cache path.
+    temporary_path = model_path.with_name(f"{model_path.name}.tmp")
+    temporary_path.unlink(missing_ok=True)
+
+    try:
+        urlretrieve(MODEL_URLS[model_name], temporary_path)
+        temporary_path.replace(model_path)
+    except Exception:
+        temporary_path.unlink(missing_ok=True)
+        raise
```
abdullah-ukaea: Could two training jobs try to download the same checkpoint at the same time? 

Every Ray actor uses the same .tmp path, so one actor could delete or replace another actor’s download. Also, urlretrieve() has no explicit timeout, so a failed connection could leave training waiting for a long time.

Could we use the existing filelock dependency to allow only one download per checkpoint at a time, and add a network timeout? After obtaining the lock, we should check again whether another actor has already completed the download.

Here is a example code snippet of a potential solution:
```python
from filelock import FileLock
from urllib.request import urlopen

  lock_path =
  model_path.with_suffix(".lock")

  with FileLock(lock_path):
      # Another actor may have downloaded it
      while we waited.
      if model_path.exists() and not
      force_download:
          return model_path

      with urlopen(MODEL_URLS[model_name],
      timeout=60) as response:

          temporary_path.write_bytes(respons
          e.read())

      temporary_path.replace(model_path)
```

toktagger/api/models/ultralytics_detection/utils.py
```diff
@@ -0,0 +1,153 @@
...
+
+def prepare_ultralytics_amp_weights() -> None:
+    """Prepare the checkpoint used by Ultralytics' CUDA AMP check.
+
+    Ultralytics performs the check using a hard-coded `yolo26n.pt`.
+    Cache that checkpoint under MODEL_STORAGE and configure Ultralytics
+    to reuse it instead of downloading another copy into the working directory.
+
+    REMOVE BELOW COMMENT AFTER SOMEONE HAS TESTED IT ON A NVIDIA GPU.
+    https://github.com/ukaea/toktagger/pull/326#discussion_r3754183103
```
abdullah-ukaea: Yep you can remove this I tested your branch on a nvidia gpu.

pyproject.toml
abdullah-ukaea: fixed in commits https://github.com/ukaea/toktagger/commit/6f9bdd25e2959f05370d6fcddead7250a4c37e28 and https://github.com/ukaea/toktagger/commit/e189613fe7544eb25cc7ab41f3d7f8565d9a1095

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,546 @@
...
+    get_toktagger_cache_dir,
+    get_torch_device,
+    prepare_ultralytics_amp_weights,
+)
+
+logger = logging.getLogger(__name__)
+
+# TYPE_CHECKING is always false while the application runs.Type checkers treat it as true
+# Othewise this will become a circular import.
+if TYPE_CHECKING:
```
wk9874: This is a bit of a code smell - can we just define YoloTrainParams in this file, and import it inside video_detection instead?

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,546 @@
...
+        ULTRALYTICS_LOGGER.setLevel(previous_level)
+
+
+class DetectionRecord(pydantic.BaseModel):
+    """Frame-level input used by the Ultralytics training dataset."""
+
+    shot_id: int
+    frame: int
+    image: bytes
+    boxes: list[list[float]]
```
wk9874: If this is always x1, x2, y1, y2, could narrow the typing to `list[tuple[float, float, float, float]]`

toktagger/api/models/ultralytics_detection/utils.py
```diff
@@ -0,0 +1,153 @@
...
+from ultralytics import settings
+
+logger = logging.getLogger(__name__)
+
+# Pretrained checkpoints available from the Ultralytics v8.4.0 assets release.
+_ULTRALYTICS_ASSET_BASE_URL = (
+    "https://github.com/ultralytics/assets/releases/download/v8.4.0"
+)
+
+MODEL_URLS = {
```
wk9874: This seems somewhat redundant - isn't it always just `f"{BASE_URL/{file_name}"`?

toktagger/api/models/ultralytics_detection/utils.py
```diff
@@ -0,0 +1,153 @@
...
+def check_pretrained_model_availability(
+    model_name: str,
+    force_download: bool = False,
+) -> Path:
+    """Return the local path to a pretrained model.
+
+    The model is downloaded into TokTagger's cache when it is not already
+    available. Setting ``force_download`` replaces an existing cached model.
+    """
+    if model_name not in MODEL_URLS:
```
wk9874: Feels like you can get rid of MODEL_URLs and just use MODEL_FAMILIES here

wk9874: Otherwise we risk a key being added to one but not the other, leading to a KeyError on line 54

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+    )
+    max_det: int = pydantic.Field(
+        default=5,
+        ge=1,
+        le=100,
+        description="Maximum number of detections per frame.",
+    )
+
+
+def iter_sample_frames(
```
wk9874, line 139: At some point you had that logic which did a coarse search until there was a non black frame, and then iterated from there - should we add that in here?

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+        )
+
+        self._prediction_model = YOLO(str(weights_path))
+        self._trained_weights_path = weights_path
+
+    def predict(
+        self,
+        samples: list[Sample],
+        params: YoloPredictParams,
+        data_params=None,
```
wk9874: Here, you could add a `this_frame_only` parameter to YoloPredictParams. If this is enabled (and data_params is available, which means the prediction was requested from an individual sample page), then take the frame number from data_params and return predictions only for that single frame.

Add in the description of the parameter that 'this is ignored for multi-sample predictions' or something

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,462 @@
...
+                    device=self.get_device().type,
+                    verbose=False,
+                )
+
+                if not results:
+                    continue
+
+                result = results[0]
+
+                if result.boxes is None or len(result.boxes) == 0:
```
wk9874: can this just be `if not result.boxes`?

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/98df223e6d40af28776b82cd13cbe1cca081e863

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/e14550f0c73c8982f2e1e3e729663cf90f9fb72b

toktagger/api/models/ultralytics_detection/utils.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/06860367158e903468ce41d09a33859b93d5179d

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: I have explored this in commit https://github.com/ukaea/toktagger/commit/c03aba1b10ae180966c7fd26b89ebff4948d8bd7 let me know if this is what you meant

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/c1e5baa2c9e40ff968a57e04423b8d437d42c19d

toktagger/api/models/ultralytics_detection/utils.py
abdullah-ukaea: also fixed in commit https://github.com/ukaea/toktagger/commit/06860367158e903468ce41d09a33859b93d5179d

toktagger/api/models/ultralytics_detection/video_detection.py
wk9874: Yep that looks good, thanks!

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: Tested GPU training and prediction successfully: no checkpoint was created in the repository or individual model directory. The AMP check reused yolo26n.pt from the shared TokTagger pretrained-model cache, so this is resolved.

toktagger/api/models/ultralytics_detection/utils.py
abdullah-ukaea: removed in commit https://github.com/ukaea/toktagger/commit/365e4708e0e27d23369fb3605150926a7af36b80

toktagger/api/models/ultralytics_detection/utils.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/a4972801cb97d51dd336466159fa03072a1ca7e8

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: tests added in commit https://github.com/ukaea/toktagger/commit/3e6b94860b572d89e429737cd811878d0c0c4b72

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: tests added in commit https://github.com/ukaea/toktagger/commit/3e6b94860b572d89e429737cd811878d0c0c4b72

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: tests added in commit https://github.com/ukaea/toktagger/commit/3e6b94860b572d89e429737cd811878d0c0c4b72

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: tests added in commit https://github.com/ukaea/toktagger/commit/3e6b94860b572d89e429737cd811878d0c0c4b72

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: tests added in commit https://github.com/ukaea/toktagger/commit/3e6b94860b572d89e429737cd811878d0c0c4b72

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: Done in commit https://github.com/ukaea/toktagger/commit/baed99bf6faf990e29372f163bb0e30adb926d7b.

 I separated the training schemas so regular YOLO offers YOLO11n again, while P2 exposes only checkpoints with compatible P2 architectures.

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: @praksharma Will leave this to you. The core functionality for Ultralytics video detection models is more or less implemented in this branch. We can maybe add this feature in a follow up PR or do this in this branch up to you.

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: done in PR https://github.com/ukaea/toktagger/pull/388

abdullah-ukaea: and applied to this branch via commit https://github.com/ukaea/toktagger/commit/08a94da7a912b27b5c3a687d89288a7ccb1f4bc4

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: resolved by PR https://github.com/ukaea/toktagger/pull/388

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,421 @@
...
+def build_video_frame_manifest(
+    samples: list[Sample],
+    annotations: list[list[Annotation]],
+    class_map: dict[str, int],
+    data_loader: TokTaggerDataLoader,
+) -> list[DetectionRecord]:
+    """
+    Convert validated video samples into frame-level training records.
+    Hash table has been used to speed up things.
+    """
```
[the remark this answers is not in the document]
abdullah-ukaea: resolved by added tests via commit https://github.com/ukaea/toktagger/commit/3e6b94860b572d89e429737cd811878d0c0c4b72 and https://github.com/ukaea/toktagger/commit/08a94da7a912b27b5c3a687d89288a7ccb1f4bc4

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: coarse search done in commits https://github.com/ukaea/toktagger/commit/dcf0fde0709b31dbb968f76d378b9d0dd614a981 and https://github.com/ukaea/toktagger/commit/b5d54dbc01dabd75351ee9abbab9fd8e98a2f643

abdullah-ukaea:
@wk9874 ready for review and can be merged if your happy with it

tests/api/test_ultralytics_video_detection.py
```diff
@@ -0,0 +1,326 @@
+import pytest
+
```
wk9874, line 2: This file should be put inside the tests/api/models directory

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,552 @@
...
+        default=False,
+        description="Predict only the current frame for individual-sample predictions; ignored for multi-sample predictions.",
+    )
+    skip_initial_black_frames: bool = pydantic.Field(
+        default=False,
+        description="Use a coarse-to-fine search to skip initial black frames for full-video prediction; ignored for individual-frame predictions.",
+    )
+
+
+def _is_useful_frame(frame_image: ImageData) -> bool:
```
wk9874, line 77: Should this be in utils.py?

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,552 @@
...
+    image = decode_frame_image(frame_image)
+    return bool(
+        image.mean() > _BLACK_FRAME_MEAN_THRESHOLD
+        or image.max() > _BLACK_FRAME_MAX_THRESHOLD
+        or image.std() > _BLACK_FRAME_STD_THRESHOLD
+    )
+
+
+def _find_first_useful_frame(
```
wk9874, line 86: Same here, utils.py?

tests/api/test_ultralytics_video_detection.py
```diff
@@ -0,0 +1,326 @@
...
+    model = object.__new__(video_detection.YoloVideoDetectionModel)
+    model._trained = False
+    model.wrapped_load(tmp_path)
+
+    expected_path = weights_dir / expected_weight
+    assert loaded_paths == [str(expected_path)]
+    assert model._trained_weights_path == expected_path
+    assert model._prediction_model is sentinel_model
+    assert model._trained is True
```
wk9874, line 326: Should add tests for the coarse search stuff here

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+class YoloTrainParams(pydantic.BaseModel):
+    """Parameters exposed by the model-training form."""
+
+    learning_rate: float = pydantic.Field(
+        default=0,
+        ge=0,
+        description="Initial learning rate. Use 0 for the Ultralytics default.",
+    )
+    epochs: int = pydantic.Field(
+        default=2,
```
wk9874, line 62: Suspect this should be higher

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+        gt=0,
+        description="Number of training epochs.",
+    )
+    yolo_size: YoloModelName = pydantic.Field(
+        default="yolo26n.pt",
+        description="Pretrained YOLO checkpoint to fine-tune.",
+    )
+    # Boolean fields render as checkboxes
+    show_training_output: bool = pydantic.Field(
```
wk9874, line 71: This should maybe be renamed 'enable_debug_logging' or something similar

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+
+
+class YoloP2TrainParams(YoloTrainParams):
+    yolo_size: YoloP2ModelName = pydantic.Field(
+        default="yolo26n.pt",
+        description="Pretrained checkpoint initializes compatible layers, new P2-specific layers are randomly initialized.",
+    )
+
+
+# https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager
```
wk9874, line 88: don't need this comment

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+    classes: list[int]
+
+
+class UltralyticsDetectionDataset(Dataset):
+    """In-memory image dataset for Ultralytics detection training.
+
+    Each record contains encoded image bytes and the bounding boxes associated
+    with that frame. Images remain PNG/JPEG encoded in the manifest and are
+    decoded only when requested by the PyTorch data loader.
+    This can be improved later by loading the images lazily.
```
wk9874, line 123: Could you raise an issue for this so we dont forget about it pls?

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+                dtype=torch.float32,
+            ).view(-1, 1)
+        else:
+            classes = torch.zeros((0, 1), dtype=torch.float32)
+
+        image_tensor = torch.from_numpy(
+            np.ascontiguousarray(letterboxed_image.transpose(2, 0, 1))
+        )
+
+        return {
```
wk9874, line 203: Should this be returned as a pydantic model, or typed dict?

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+        classes = []
+        bboxes = []
+        batch_indices = []
+        image_files = []
+        original_shapes = []
+        resized_shapes = []
+        ratio_pads = []
+
+        for batch_index, sample in enumerate(batch):
+            images.append(sample["img"])
```
wk9874, line 229: This will cause keyerror if this isnt present, assume it requires input from the method above? Pydantic model would help with that

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+                    dtype=torch.int64,
+                )
+            )
+
+            image_files.append(sample["im_file"])
+            original_shapes.append(sample["ori_shape"])
+            resized_shapes.append(sample["resized_shape"])
+            ratio_pads.append(sample["ratio_pad"])
+
+        return {
```
wk9874, line 246: And again maybe return pydantic model

toktagger/api/models/ultralytics_detection/base.py
```diff
@@ -0,0 +1,594 @@
...
+            del self._prediction_model
+
+        self.log_progress(
+            status="completed",
+            progress=100,
+        )
+
+        # Validation is intentionally postponed, so no meaningful score is
+        # available yet.
+        return 0.0
```
wk9874, line 557: We should probably change the models schema to accept `score: float | None` for the score, and have this return None, with the UI showing 'N/A' or something if it gets a None score. This would also be useful for the model loading stuff, where similarly no score is available.

Happy for this to go into an issue tho and tackle later if you dont want to do it now (or want me to do it!)

toktagger/api/models/ultralytics_detection/utils.py
```diff
@@ -0,0 +1,148 @@
...
+                    shutil.copyfileobj(response, output)
+            temporary_path.replace(model_path)
+        except Exception:
+            temporary_path.unlink(missing_ok=True)
+            raise
+
+    return model_path
+
+
+def get_torch_device() -> torch.device:
```
wk9874, line 88: I think this may need renaming to not be confusing, since it will ignore cuda GPUs (intentionally)

Alternatively (and probably preferred) - have this function accept a `use_cuda: bool` flag or something, which basically does the check currently done in base.py L381:
```
if `use_cuda and torch.cuda.is_available():
   return torch.device("cuda")
```

And then remove that check from base.py L381, instead passing the `self.gpu_available` into this func as `use_cuda`.

toktagger/api/models/ultralytics_detection/utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def get_toktagger_cache_dir() -> Path:
+    """Return the model-storage directory supplied to the Ray worker."""
+    cache_dir = Path(os.environ["MODEL_STORAGE"])
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    return cache_dir
+
+
+def resolve_weights_path(
```
wk9874, line 108: The code inside here seems to be duplicated inside the `save` method in base.py? can we consolidate the two?

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,552 @@
...
+        default=0.2,
+        ge=0,
+        le=1,
+        description="Minimum confidence required for a detection.",
+    )
+    iou_threshold: float = pydantic.Field(
+        default=0.2,
+        ge=0,
+        le=1,
+        description="Intersection-over-union threshold.",
```
wk9874, line 59: idk what this means - can we rephrase to be clearer?

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,552 @@
...
+    )
+    max_det: int = pydantic.Field(
```
wk9874, line 61: `max_detections`?

toktagger/api/models/ultralytics_detection/video_detection.py
wk9874: Would be good to add tests for this

toktagger/api/models/ultralytics_detection/video_detection.py
```diff
@@ -0,0 +1,552 @@
...
+                )
+            except FrameNotFoundError:
+                return initial_frame
+
+            if _is_useful_frame(refinement_frame):
+                return refinement_frame
+
+        return fallback_frame
+
+    return initial_frame
```
wk9874, line 136: I thiiiink the logic in this function is correct, but I find it a bit hard to read. Could we maybe refactor it to be more readable, or add comments which explain the logic?

wk9874 (changes requested):

tests/api/models/test_ultralytics_video_detection.py
```diff
@@ -0,0 +1,326 @@
+import pytest
+
```
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/d34d758d1fa15e783d4b07d1dc03151200e90199

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: yep, moved in commit https://github.com/ukaea/toktagger/commit/2662b757c3270b898ff704e17033eec8b38b9018

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: yep, moved in commit https://github.com/ukaea/toktagger/commit/2662b757c3270b898ff704e17033eec8b38b9018 and tests done in commit https://github.com/ukaea/toktagger/commit/51d7c7a3e19a230aeb4754f962f215795b8389a6

tests/api/models/test_ultralytics_video_detection.py
```diff
@@ -0,0 +1,326 @@
...
+    model = object.__new__(video_detection.YoloVideoDetectionModel)
+    model._trained = False
+    model.wrapped_load(tmp_path)
+
+    expected_path = weights_dir / expected_weight
+    assert loaded_paths == [str(expected_path)]
+    assert model._trained_weights_path == expected_path
+    assert model._prediction_model is sentinel_model
+    assert model._trained is True
```
abdullah-ukaea: tests added in commit https://github.com/ukaea/toktagger/commit/51d7c7a3e19a230aeb4754f962f215795b8389a6

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: yep default should be 100 fixed in commit https://github.com/ukaea/toktagger/commit/9b3be848ed311c2efd790319a471433d9250b0f3

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/29a7e30b79db1121197db6aaa0674901c35897ca

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: removed in commit https://github.com/ukaea/toktagger/commit/2d7fffe2865b13dcdec84c90da106693e7c750d2

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/6ba701e9b0c2bb4ddbf7292bbd4a67f0beb88e63

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/9fc9d9e90bbba77c642268022b7826058b179306

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/d4c66c7067f41692d9cfd041b35066171465a009

toktagger/api/models/ultralytics_detection/utils.py
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/81f716da0e28b1a1dae3f0d86b420b14d0090588

toktagger/api/models/ultralytics_detection/utils.py
abdullah-ukaea: attempted to consolidate in https://github.com/ukaea/toktagger/commit/711f00d950deb52bfdb720c47925747141fd2d81

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: worded it better in commit https://github.com/ukaea/toktagger/commit/d540f7fea627a6d57cc8edda229283a2895254c3

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/ab11414b57c2da016cb48980d75a37f30e50d27b

toktagger/api/models/ultralytics_detection/video_detection.py
abdullah-ukaea: I didn't refactor but I did improve variable names and added more comments in commit https://github.com/ukaea/toktagger/commit/0eac05d0b3ddfbaa82d2d5299f6c3b6ebb3e5fbd

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: created issue https://github.com/ukaea/toktagger/issues/404

toktagger/api/models/ultralytics_detection/base.py
abdullah-ukaea: I am happy to do this, but will do it in a follow up PR.

As I looked into how to do it and this one change will modify the below files:
toktagger/api/schemas/models.py
toktagger/api/routers/models.py
toktagger/api/models/base.py
toktagger/api/models/ultralytics_detection/base.py
toktagger/ui/src/types.ts
toktagger/ui/src/app/components/tools/modelPredict.tsx

Created issue https://github.com/ukaea/toktagger/issues/405

PR #331, opened 2026-07-22
Add controls for adding and removing list items in schema forms

wk9874:


toktagger/ui/src/app/components/ui/schemaForm.tsx
```diff
@@ -402,14 +409,38 @@ const SpectrumArrayFieldTemplate = ({
...
 };
 
 const SpectrumArrayFieldItemTemplate = ({
   children,
+  buttonsProps,
 }: ArrayFieldItemTemplateProps) => {
-  return <View marginX={"size-400"}>{children}</View>;
+  return (
+    <Flex marginX="size-400" gap="size-100" alignItems="end">
+      <View UNSAFE_style={{ flex: 1 }}>{children}</View>
```
wk9874: Do we need this unsafe style here? Whats it doing?

toktagger/ui/src/app/components/ui/schemaForm.tsx
```diff
@@ -402,14 +409,38 @@ const SpectrumArrayFieldTemplate = ({
...
+      {buttonsProps.hasRemove && (
+        <ActionButton
+          aria-label={`Remove item ${buttonsProps.index + 1}`}
+          isDisabled={buttonsProps.disabled || buttonsProps.readonly}
```
wk9874, line 439: This allows to remove all input boxes, even if min_items > 1

I suspect there's not much we can do about this, since we rely on `react-jsonschema-form` to tell us when it should be removable?

It also will fail validation with a useful error message if you try to submit the form with no items, so I'm happy to ignore it if there's not an easy fix

wk9874: Similar for type hints in the schema like `tuple[int, int, int]` - it still lets you remove boxes when it shouldn't really allow that

(interestingly, it doesnt let you add any more boxes, which is good!)

toktagger/ui/src/app/components/ui/schemaForm.tsx
```diff
@@ -402,14 +409,38 @@ const SpectrumArrayFieldTemplate = ({
...
 const SpectrumArrayFieldItemTemplate = ({
   children,
+  buttonsProps,
 }: ArrayFieldItemTemplateProps) => {
-  return <View marginX={"size-400"}>{children}</View>;
+  return (
+    <Flex marginX="size-400" gap="size-100" alignItems="end">
+      <View UNSAFE_style={{ flex: 1 }}>{children}</View>
+      {buttonsProps.hasRemove && (
+        <ActionButton
```
wk9874, line 435: Maybe buttons should be slightly smaller

toktagger/ui/src/app/components/ui/schemaForm.tsx
abdullah-ukaea: Good catch, the react spectrum docs say to only use unsafe styling as a last resort. There is a supported style prop I can use instead. Fixed in commit https://github.com/ukaea/toktagger/commit/22308ccc4515406688b813fe7895b829936e348d

Reference: https://react-spectrum.adobe.com/v3/View.html

toktagger/ui/src/app/components/ui/schemaForm.tsx
abdullah-ukaea: changed in commit, let me know if it is fine now https://github.com/ukaea/toktagger/commit/ac4110ddecbda9c51a4e4f3059c243fd334907c1

abdullah-ukaea: before commit 
1_input_box

After commit
smaller_buttons

toktagger/ui/src/app/components/ui/schemaForm.tsx
abdullah-ukaea: Hmm, I had a look and I suggest leaving this as is for now. 

Users can easily add the input again, and RJSF shows a validation error if the form has fewer than the required number of items.

Preventing removal correctly would require extra handling for minItems, optional lists, and fixed tuples, which feels unnecessary for this change. But let me know if you feel strongly that this should be enforced.

abdullah-ukaea:
ready for re review @wk9874

wk9874:
Looks good!

Issue #272, opened 2026-05-19
Align frame and plotly annotation views


jblake42:
There is a disparity between the plotly annotation tool and the frame annotation tool.

We should standardise across these views to ensure the overall product is cohesive. This should be done once the funcitonality of both systems are stable.

PR #333, opened 2026-07-23
Video: Fix race condition during rapid video frame navigation

abdullah-ukaea:
# Summary

Rapid frame navigation could cause stale responses to overwrite newer frames, leaving the video view flickering and unusable. 

This change tracks the latest requested frame, cancels older requests, and only allows the newest response to update the UI. Rapid Next and Previous clicks now resolve to the expected frame.

This will close issue https://github.com/ukaea/toktagger/issues/332

# Current dev branch behaviour

https://github.com/user-attachments/assets/32132e6d-096a-45e0-ba66-1a3a06f62332

# This branch behaviour

https://github.com/user-attachments/assets/589f9edc-e3fb-4d37-9731-49b88b0ec5a6

PR #334, opened 2026-07-23
Video: Harmonise video annotation UI with time series

abdullah-ukaea:
This PR aligns the video annotation UI with the time series view by replacing the floating icon toolbar with a clear right-side annotation toolbar and introducing View and Edit modes.

 * View mode: users can zoom, pan, and select existing annotations.
 * Edit mode: users can also move and reshape existing annotations.
 * Edit mode + Ctrl: users can draw new annotations using the selected tool, including over existing annotations.

Users can switch modes using the toolbar or the E keyboard shortcut. Existing video features such as propagation, hiding annotations, class selection, and instance tracking remain unchanged.

Closes issue https://github.com/ukaea/toktagger/issues/272

# Video view before PR
video_view_before

# Video view after PR
video_view_after

# Time series view 
time_view_now


abdullah-ukaea:
@wk9874 @jblake42 Hi guys, please give this PR a look and let me know what you think. 

Here is the to do list:

- [x] User functional test (Matt)
- [x] UI developer functional test (Josh)
- [ ] Code review (Josh)

abdullah-ukaea:
To Do:

- [x] Get https://github.com/ukaea/toktagger/pull/333 merged into dev
- [x] Merge dev into this branch video/general-ui-improvement
- [x] Add feature to disable annotations during frame loading (when a user clicks next to go to next frame for high latency data loaders)

wk9874:
Overall looks great! Nice to see the two views being harmonized together. A few points from my functionality test:

1. When loading a new frame, the existing frame is no longer overlayed with a grey box along with the loading wheel. Eg compare the video here: https://github.com/ukaea/toktagger/pull/333#issuecomment-5141842487 to the video below. Think it looked nicer before!

https://github.com/user-attachments/assets/b4f79767-c98a-4302-a404-3ea4213dd509

2. I personally think that one of the tools on the right hand side should be auto selected for the user on first load, and its not possible to have no tool selected (basically how it was before harmonization). I think its more intuitive to the user to see something drawn on the screen instead of a popup asking you to select a tool. (But this new way is consistent with the current time series implementation, so happy for this to remain as is and raise an issue to change both later. Or if you / @jblake42 think its better the way it is then happy to leave it)
3. Related to 2 - selecting a tool, then going to View mode, then back to Edit mode, deselects the tool. I can see this getting quite annoying for the user having to constantly reselect the tool every time they switch modes. I think it should just put you back into the mode you were in last time you were in Edit mode. But again this is consistent with time series atm, so happy to raise an issue and tackle separately
4. For 'Add feature to disable annotations during frame loading', while the buttons now do become greyed out, I can still draw:

https://github.com/user-attachments/assets/cc7fcc33-b6df-4f81-b26e-8d00790e8f92

5. When going to next sample, the help box will pop back up if in edit mode. Can this be suppressed? Dont need to be told the info on every navigation :)
6. Bounding boxes & polygons can be deleted in View mode. That shouldn't be the case - I think the ideal scenario is that the pop up still appears in View mode if the item is clicked, but the delete button is disabled, Also mentioned in #322 but should be addressed here
7. This one's slightly more of a feature request, so happy for it to go into an issue if you don't want to tackle it now. I think the 'right click to change mode' feature should also support right clicking an **existing annotation** to change its class label. Ie if I've drawn a bounding box, then realise my class label was set to 'Major UFO' instead of 'Minor UFO', I want to be able to change that label without redrawing the box completely

abdullah-ukaea:
> Overall looks great! Nice to see the two views being harmonized together. A few points from my functionality test:
> 
> 1. When loading a new frame, the existing frame is no longer overlayed with a grey box along with the loading wheel. Eg compare the video here: [Video: Fix race condition during rapid video frame navigation #333 (comment)](https://github.com/ukaea/toktagger/pull/333#issuecomment-5141842487) to the video below. Think it looked nicer before!
> 
>  toktagger_loading.mp4 
> 2. I personally think that one of the tools on the right hand side should be auto selected for the user on first load, and its not possible to have no tool selected (basically how it was before harmonization). I think its more intuitive to the user to see something drawn on the screen instead of a popup asking you to select a tool. (But this new way is consistent with the current time series implementation, so happy for this to remain as is and raise an issue to change both later. Or if you / @jblake42 think its better the way it is then happy to leave it)
> 3. Related to 2 - selecting a tool, then going to View mode, then back to Edit mode, deselects the tool. I can see this getting quite annoying for the user having to constantly reselect the tool every time they switch modes. I think it should just put you back into the mode you were in last time you were in Edit mode. But again this is consistent with time series atm, so happy to raise an issue and tackle separately
> 4. For 'Add feature to disable annotations during frame loading', while the buttons now do become greyed out, I can still draw:
> 
>  toktagger_draw_while_loading.mp4 
> 5. When going to next sample, the help box will pop back up if in edit mode. Can this be suppressed? Dont need to be told the info on every navigation :)
> 6. Bounding boxes & polygons can be deleted in View mode. That shouldn't be the case - I think the ideal scenario is that the pop up still appears in View mode if the item is clicked, but the delete button is disabled, Also mentioned in [Video: xy point improvements  #322](https://github.com/ukaea/toktagger/pull/322) but should be addressed here
> 7. This one's slightly more of a feature request, so happy for it to go into an issue if you don't want to tackle it now. I think the 'right click to change mode' feature should also support right clicking an **existing annotation** to change its class label. Ie if I've drawn a bounding box, then realise my class label was set to 'Major UFO' instead of 'Minor UFO', I want to be able to change that label without redrawing the box completely

Thank you for the functional testing of this branch I appreciate it! @wk9874 

Regarding your points:

1.  I have been experimenting with grey box on / off as it is trivial to change (I just make the div transparent with 1 line) and to be honest I think I disagree with you. I find the grey box to look "glitchy" and not very aesthetic. Currently a user sees a spinner in the centre of the frame to indicate they are still in a frame loading state, I think this is enough. Perhaps @jblake42 can make the final decision here.
2.  Auto selection of tools from the beginning of an annotation session is interesting and I guess my first question is how do you decide what tool to select from the beginning? If you simply hard code to always use bounding box by default, annotators using polygon or point annotations still have to select there tool anyway. Perhaps in the future we can add a "power user" optional setting where you can select what tool to autoselect for every sample in video view? But again to be consistent with time series I will only make this change for video if we do it for time series also. @jblake42  let me know what you think.
3. Agreed and done in this PR
4. Good catch, I have now patched this in this PR
5. Yep I also found the help box popping up to be quite annoying. I have disabled it to automatically popup it now opens via a mouse click manually.
6. I have now fixed this, in this PR. Now in view mode the annotation popup delete button is disabled. It still works the same in edit mode.
7. So I think I know how to do this, but it is non trivial to implement.  I would prefer to create an issue for this if there isn't one already and I will tackle this in a future PR.

wk9874:
Have had another test, all looks good!

For 7, raised #361 

Happy for 1 & 2 to stay as they are

Tested 3, 4, 5, 6 are fixed

Only other comment - should we put the annotation list under the image, in the same format as we have it on Time Series & Profile 2D? Ie should this:
image

Be made to look more like this:
image

I would still want the delete & jump to buttons though - in fact this could be useful to add to the time series & profile2D modes as well... (different PR?)

wk9874:
Raised #362 for that last point

abdullah-ukaea:
This PR is ready for a code review @jblake42

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
```diff
@@ -525,13 +471,13 @@ function Inner({ imageBase64 }: { imageBase64: string }) {
...
-          : ""
-        : isOverAnnotation
+      const cursor = canDrawPoint
+        ? "crosshair"
+        : editMode && isOverAnnotation
           ? "pointer"
-          : "";
+          : isOverAnnotation
+            ? "default"
+            : "";
```
jblake42: I personally believe this to be more readable:
```
const cursor = canDrawPoint
  ? "crosshair"
  : isOverAnnotation
    ? editMode
      ? "pointer"
      : "default"
    : "";
```

toktagger/ui/src/app/video/components/annotation-popup.tsx
```diff
@@ -60,7 +61,8 @@ export function AnnotationPopup(props: {
         <div className="mt-2 flex gap-2">
           <button
             onClick={props.onDeleteBox}
-            className="rounded-md bg-red-500/20 hover:bg-red-500/30 text-red-200 px-2.5 py-1.5 text-xs"
+            disabled={props.deleteDisabled}
+            className="rounded-md bg-red-500/20 enabled:hover:bg-red-500/30 text-red-200 px-2.5 py-1.5 text-xs disabled:opacity-40 disabled:cursor-not-allowed"
```
jblake42: Could we use a variant provided by spectrum here (`negative`) - without testing these look similar to me

toktagger/ui/src/app/video/components/video-annotation-toolbar.tsx
jblake42, line 1: It would be interesting to see how much we can abstract into a shared component - even if it is just a standard toolbar component that provides a way to add tool-specific components to (props or a portal) - I have created an issue

toktagger/ui/src/app/contexts/VideoContext.tsx
jblake42, line 1: Can we move this file into the standardised context folder and rename to be consistent - it is useful to have the contexts in one place as they frequently have to be referred to.

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -469,22 +474,46 @@ export function VideoSessionProvider(props: {
...
+    dataParams.frame !== frame;
+
+  const drawIntent = editMode && ctrlHeld && !hideAnnotations;
+  const canCreate =
+    drawIntent &&
+    drawingTool !== null &&
+    Boolean(selection.className) &&
+    !isFramePending;
+  const canDrawShape = canCreate && drawingTool !== "point";
+  const canDrawPoint = canCreate && drawingTool === "point";
```
jblake42, line 497: I'm a little concerned about the different handling of points vs every other shape despite them both being pretty much native functionality? Could you explain why we need these please?

toktagger/ui/src/app/video/components/video-session.tsx
```diff
@@ -658,20 +739,42 @@ export function VideoSessionProvider(props: {
...
+    if (!canDrawShape) {
       api.cancelDrawing?.();
     }
-  }, [api, drawingTool, hideAnnotations, panMode, selection.className]);
+
+    if (api.viewer && canDrawPoint) {
+      api.viewer.setMouseNavEnabled(false);
+    } else if (api.viewer && !canDrawShape) {
+      api.viewer.setMouseNavEnabled(true);
+    }
```
jblake42, line 768: Is this logic not repeated in `frame-annotator-host`? If so can we try to ensure there is one point of truth

jblake42 (changes requested):
This looks great - functionality is as expected! Only some small suggestions

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: Remember rectangle and polygon are actually native Annotorious tools/shapes, there's no built-in "point" concept at all. 

A "point" here is really a fixed-size ellipse we stamp in programmatically on a single click (auto-assigning class/track id), completely bypassing Annotorious's native drag-to-draw pipeline that rectangle/polygon use, so canDrawPoint has to gate a different, hand-rolled creation path rather than setDrawingEnabled. 

I don't think this needs a code change it's a consequence of us choosing a click-to-draw point and using plugin shapes vs native shapes.

References:
- https://annotorious.dev/api-reference/shape/
- https://github.com/annotorious/plugin-tools

toktagger/ui/src/app/video/components/frame-annotator-host.tsx
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/31cbebf2dde25337bc6f1511f02bf9854f276602

toktagger/ui/src/app/video/components/annotation-popup.tsx
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/6fa8f761140cad2da5d7cc3609345d82fe226699

toktagger/ui/src/app/contexts/VideoContext.tsx
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/60defa9befdc0a123bc037900dfb571ccba727f6

toktagger/ui/src/app/video/components/video-session.tsx
abdullah-ukaea: good catch I didn't realise I was duplicating this. fixed in commit https://github.com/ukaea/toktagger/commit/977bc3fccd5888f830fb57e0018edd9b7f712f51

toktagger/ui/src/app/video/components/video-annotation-toolbar.tsx
abdullah-ukaea: happy to tackle this in a future PR

abdullah-ukaea:
@jblake42 ready for re-review

toktagger/ui/src/app/contexts/VideoContext.tsx
jblake42: I have updated the file name to align to other context files

toktagger/ui/src/app/video/components/video-session.tsx
jblake42: That makes sense!

PR #341, opened 2026-07-28
Use COCO compatible nested polygon segmentation for video and time series

abdullah-ukaea:
This focused change follows the discussion in PR https://github.com/ukaea/toktagger/pull/195 / https://github.com/ukaea/toktagger/pull/195#discussion_r3665645823 and updates video and time series polygons to use COCO compatible nested segmentation [1] across the frontend and backend.  It aligns the schemas, types, and conversion utilities so a single polygon is stored as [[x1, y1, ...]]. 

The derived area and bbox fields are deliberately omitted because they can be calculated from the segmentation by a backend or export layer if needed. These derived values do not belong in the video drawing components, which should remain focused on annotation creation, conversion, and visualisation.

references:

[1] - https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/mask.py#L47-L68


abdullah-ukaea:
@samueljackson92 Feel free to give me any feedback on the polygon schema for video and time series but I strongly believe we should fix it in a separate / concise PR like this. This can then be merged to dev and used for profiles 2d.

toktagger/api/schemas/annotations.py
```diff
@@ -62,35 +62,47 @@ class VideoBoundingBox(AnnotationBase):
...
+            if len(polygon) < 6:
+                raise ValueError("Each polygon must contain at least three points.")
+            if len(polygon) % 2 != 0:
+                raise ValueError(
+                    "Each polygon must contain an even number of coordinates."
+                )
         return v
 
 
 class VideoPolygon(AnnotationBase):
```
samueljackson92, line 85: Why does VideoPolygon not inherit from Polygon?

samueljackson92 (changes requested):
One small comment.

toktagger/api/schemas/annotations.py
abdullah-ukaea: Ah so one interesting thing here is that VideoPolygon has existed before Polygon! https://github.com/ukaea/toktagger/commit/cba0c402c819b5ed680280194bed8c28ad36fa85

I am happy to look at the schema and change this now that D3 based polygons is in but that class did not exist a month ago! it only was introduced in https://github.com/ukaea/toktagger/pull/314/

I will talk to Josh and maybe refactor this now. Perhaps we can have a base polygon class, and then timeseries/spectogram/video if needed can inherit from that? (if that is even needed.) @samueljackson92  let me know how you think we should define the schema now and I will change it in this PR.

samueljackson92: I don't think we need a base class. I think `VideoPolygon` just needs to inherit from `Polygon` now that both exist.

jblake42: When I added these new annotation types I did initially template it all however I believe the version of python we used didn't support default types so I removed this.

Python does not like directly overriding pydantic fields with different types (which is needed if VideoPolygon just directly inherits Polygon) so the only way I could think of having them inherit is something like this (completely untested):

```
T = TypeVar("T")
class Polygon[T](AnnotationBase):
    type: Literal["timeseries_polygon", "video_polygon"]
    segmentation: list[T] = Field(
        ...,
        min_length=6,
        description="COCO polygon segmentation as a flat list: [x1, y1, x2, y2, ...].",
    )

    @field_validator("segmentation")
    @classmethod
    def validate_segmentation(cls, v: list[T]) -> list[T]:
        if len(v) % 2 != 0:
            raise ValueError("Segmentation must contain an even number of coordinates.")
        return v

class TimeSeriesPolygon(Polygon[float]):
    type: Literal["timeseries_polygon", "video_polygon"] = "timeseries_polygon"

class VideoPolygon(Polygon[int]):
    type: Literal["timeseries_polygon", "video_polygon"] = "video_polygon"
    frame: int
    track_id: str
```

I don't know how much clearer this makes things, but it does at least remove some repetition which I believe is what you are going for Sam?

samueljackson92: Ah I see. The issue is the typing. And COCO will require `int` because it expects pixel size, while that does not make sense from time series/profile 2d.

We can either leave as is, or do the above as follow up.

abdullah-ukaea: @samueljackson92 if we are leaving as is, then the PR is ready to be code reviewed (if you havent already) and then you can merge. Maybe just give the video polygon and time series polygon a quick functional test to check there are no bugs with polygon but it looked good to me.

samueljackson92 (approved):

PR #346, opened 2026-08-03
Change save/load to use results dir paths instead of single files

wk9874:
Changes model storage to now make a directory available with name equal to the model ID, so that more than one file can be saved for a given model.

`load()` now accepts two parameters, the storage directory, and (optionally) a weights file name. If the weights file name is provided, it should take precedence, otherwise fallback to what the model's `save()` method will have created.

See example implementation in the `disruption_cnn` model

Note that when loading files via the UI, we will still only support the loading of a single weights file (for now...)

### DIFF
--- a/tests/api/routers/test_models.py
+++ b/tests/api/routers/test_models.py
@@ -738,8 +738,9 @@ async def test_model_load_local_failed(models_api_client, db_client, setup_model
         assert model["training_status"] == "failed"
 
         # Check model has not been saved after completion
-        assert (
-            not pathlib.Path(config.settings.models.cache_dir)
-            .joinpath(f"{model_id}.model")
-            .exists()
+        model_path = (
+            pathlib.Path(config.settings.models.cache_dir)
+            .joinpath(f"{model_id}")
+            .joinpath("weights.model")
         )
+        assert not model_path.exists()
--- a/tests/api/routers/test_models.py
+++ b/tests/api/routers/test_models.py
@@ -163,8 +163,8 @@ async def test_model_predict_missing_weights(
 ):
     # Delete weights
     config.settings.models.cache_dir.joinpath(
-        f"{setup_model_db['model_id_1']}.model"
-    ).unlink()
+        f"{setup_model_db['model_id_1']}"
+    ).joinpath("weights.model").unlink()
     response = await models_api_client.post(
         f"/projects/{setup_model_db['project_id']}/models/mock_disruption_cnn/predict?num_predictions=5&version=1"
     )
@@ -376,7 +376,11 @@ async def test_model_start_training_no_params(
     assert model["score"] == 60  # value returned by train method
 
     # Check model has been saved after completion
-    assert config.settings.models.cache_dir.joinpath(f"{model_id}.model").exists()
+    assert (
+        config.settings.models.cache_dir.joinpath(f"{model_id}")
+        .joinpath("weights.model")
+        .exists()
+    )
 
 
 @pytest.mark.asyncio
@@ -462,7 +466,11 @@ async def test_model_start_training_params(
     assert model["score"] == 50  # value returned from params
 
     # Check model has been saved after completion
-    assert config.settings.models.cache_dir.joinpath(f"{model_id}.model").exists()
+    assert (
+        config.settings.models.cache_dir.joinpath(f"{model_id}")
+        .joinpath("weights.model")
+        .exists()
+    )
 
 
 # Test delete model
@@ -483,17 +491,17 @@ async def test_model_delete_type(models_api_client, db_client, setup_model_db):
 
     # Check for models 1 and 2, their file no longer exists
     assert not config.settings.models.cache_dir.joinpath(
-        f"{setup_model_db['model_id_1']}.model"
+        f"{setup_model_db['model_id_1']}"
     ).exists()
     assert not config.settings.models.cache_dir.joinpath(
-        f"{setup_model_db['model_id_2']}.model"
+        f"{setup_model_db['model_id_2']}"
     ).exists()
     # And for model 3 it does still exist
     assert config.settings.models.cache_dir.joinpath(
-        f"{setup_model_db['model_id_3']}.model"
+        f"{setup_model_db['model_id_3']}"
     ).exists()
     assert config.settings.models.cache_dir.joinpath(
-        f"{setup_model_db['model_id_3']}.model"
+        f"{setup_model_db['model_id_4']}"
     ).exists()
 
 
@@ -518,13 +526,13 @@ async def test_model_delete_type_version(models_api_client, db_client, setup_mod
 
     # Check for model 2, their file no longer exists
     assert not config.settings.models.cache_dir.joinpath(
-        f"{setup_model_db['model_id_2']}.model"
+        f"{setup_model_db['model_id_2']}"
     ).exists()
     # And for models 1, 3 and 4 it does still exist
     assert all(
         (
             config.settings.models.cache_dir.joinpath(
-                f"{setup_model_db[model_id]}.model"
+                f"{setup_model_db[model_id]}"
             ).exists()
         )
         for model_id in ("model_id_1", "model_id_3", "model_id_4")
@@ -647,8 +655,10 @@ async def test_model_load_local(models_api_client, db_client, setup_model_db):
         assert model["progress"] == 100
 
         # Check model has been saved after completion
-        model_path = pathlib.Path(config.settings.models.cache_dir).joinpath(
-            f"{model_id}.model"
+        model_path = (
+            pathlib.Path(config.settings.models.cache_dir)
+            .joinpath(f"{model_id}")
+            .joinpath("weights.model")
         )
         assert model_path.exists()
 
--- a/tests/api/routers/test_models.py
+++ b/tests/api/routers/test_models.py
@@ -471,6 +471,11 @@ async def test_model_start_training_params(
         .joinpath("weights.model")
         .exists()
     )
+    assert (
+        config.settings.models.cache_dir.joinpath(f"{model_id}")
+        .joinpath("config.json")
+        .exists()
+    )
 
 
 # Test delete model
--- a/toktagger/api/models/base.py
+++ b/toktagger/api/models/base.py
@@ -14,6 +14,7 @@ import pydantic
 import uuid
 from collections import OrderedDict
 import logging
+import pathlib
 
 logger = logging.getLogger("ray")
 
@@ -102,14 +103,16 @@ class Model(ABC):
         return self.predict(samples=samples, params=params, data_params=data_params)
 
     @typing.final
-    def wrapped_save(self, file_stem: str) -> None:
+    def wrapped_save(self, results_dir: pathlib.Path) -> None:
         if not self._trained:
             raise RuntimeError("Cannot save a model before it has been trained!")
-        self.save(file_stem=file_stem)
+        self.save(results_dir=results_dir)
 
     @typing.final
-    def wrapped_load(self, file_path: str) -> None:
-        self.load(file_path=file_path)
+    def wrapped_load(
+        self, results_dir: pathlib.Path, weights_filename: str | None = None
+    ) -> None:
+        self.load(results_dir=results_dir, weights_filename=weights_filename)
         self._trained = True
 
     @typing.final
@@ -126,6 +129,17 @@ class Model(ABC):
         progress: float | None = None,
         score: float | None = None,
     ) -> None:
+        """Send progress updates during model training to the TokTagger server.
+
+        Parameters
+        ----------
+        training_status : typing.Literal[ "queued", "started", "failed", "completed", "aborted" ] | None, optional
+            The current stage of model training, by default None
+        progress : float | None, optional
+            How far through training the model currently is (percentage, 0-100), by default None
+        score : float | None, optional
+            A metric of how well the model is performing (such as accuracy or loss), by default None
+        """
         model_update = ModelUpdate(
             training_status=training_status, progress=progress, score=score
         )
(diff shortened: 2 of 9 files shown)


toktagger/api/models/disruption.py
```diff
@@ -364,8 +365,13 @@ def predict(
...
 
-    def load(self, file_path: str):
-        self.model.load_state_dict(torch.load(file_path, map_location="cpu"))
+    def load(self, results_dir: pathlib.Path, weights_filename: str | None = None):
+        if weights_filename:
+            results_path = results_dir.joinpath("weights_filename")
+        else:
+            results_path = results_dir.joinpath("weights.model")
+
+        self.model.load_state_dict(torch.load(results_path, map_location="cpu"))
```
abdullah-ukaea: The "weights_filename" variable is in quotes, so it's treated as a literal filename instead of the variable holding the actual filename passed in. This means the custom filename feature never works it always looks for a file literally called weights_filename and throws FileNotFoundError.

Just remove the quotes so it reads results_dir.joinpath(weights_filename), using the real variable. Also add the missing -> None return type to match save() above it.

toktagger/api/worker.py
```diff
@@ -69,11 +69,9 @@ def get_actor(project: Project, model: Model, use_gpu: bool):
...
-        model_path = next(
-            pathlib.Path(os.environ["MODEL_STORAGE"]).glob(f"{str(model.id)}*"), None
-        )
-        if model_path:
-            ml_model.wrapped_load.remote(model_path)
+        results_dir = pathlib.Path(os.environ["MODEL_STORAGE"]).joinpath(str(model.id))
+        if results_dir.exists():
+            ml_model.wrapped_load.remote(results_dir)
         else:
             logger.debug("No saved weights found, initializing blank model")
```
abdullah-ukaea, line 77: This looks weird as the code now checks "does the directory exist" instead of "does a weights file actually exist," but the directory gets created before the save happens, and the save can fail or crash without ever writing anything into it. That leaves behind an empty directory, which later fools the code into thinking weights exist, causing it to crash instead of falling back to a blank model.

Can we check that the directory actually has contents (results_dir.is_dir() and any(results_dir.iterdir())) rather than just that it exists, and make sure the save completes successfully before marking training as done.

toktagger/api/worker.py
```diff
@@ -154,9 +151,9 @@ def train_model(
         # Wait for train task to complete
         score = ray.get(train_task)
 
-        model_dir = pathlib.Path(os.environ["MODEL_STORAGE"])
-        model_dir.mkdir(exist_ok=True)  # Do i need to do this every time?
-        model_actor.wrapped_save.remote(model_dir.joinpath(str(model.id)))
+        results_dir = pathlib.Path(os.environ["MODEL_STORAGE"]).joinpath(str(model.id))
+        results_dir.mkdir(parents=True)
+        model_actor.wrapped_save.remote(results_dir)
```
abdullah-ukaea: same problem as this comment https://github.com/ukaea/toktagger/pull/346/changes#r3705898262

toktagger/api/models/base.py
```diff
@@ -102,14 +103,16 @@ def wrapped_predict(
...
+        self.save(results_dir=results_dir)
 
     @typing.final
-    def wrapped_load(self, file_path: str) -> None:
-        self.load(file_path=file_path)
+    def wrapped_load(
+        self, results_dir: pathlib.Path, weights_filename: str | None = None
+    ) -> None:
+        self.load(results_dir=results_dir, weights_filename=weights_filename)
         self._trained = True
```
abdullah-ukaea, line 119: The default filename "weights.model" and the fallback logic (use the given filename, otherwise use the default) are copy pasted into every model file instead of living in one place which is exactly how the bug in comment 1 happened. On top of that, each model's save() has to assume its directory already exists, but nothing in the base class guarantees that.

Move this shared logic into the base class once so every model just calls it instead of re implementing it. Also have wrapped_save create the directory (mkdir(parents=True, exist_ok=True)) before calling save(), so individual models and worker.py don't need to handle that themselves.

toktagger/api/worker.py
```diff
@@ -125,10 +121,11 @@ def load_model(
             "message": f"Failed to load weights - {str(e)}",
         }
 
-    # Save the model with the correct file name, delete temporary file
-    save_weights_task = model_actor.wrapped_save.remote(
-        model_dir.joinpath(str(model.id))
-    )
+    # Save the model with the correct dir path
+    results_dir = pathlib.Path(os.environ["MODEL_STORAGE"]).joinpath(str(model.id))
+    results_dir.mkdir(parents=True)
```
abdullah-ukaea: mkdir() is called without exist_ok=True, so it crashes if the directory already exists. 

Add exist_ok=True so it doesn't crash when the folder is already there  or better, remove these mkdir() calls entirely and let wrapped_save handle creating the directory instead as I mentioned in this comment in base.py https://github.com/ukaea/toktagger/pull/346/changes#r3705953069

toktagger/api/worker.py
```diff
@@ -154,9 +151,9 @@ def train_model(
         # Wait for train task to complete
         score = ray.get(train_task)
 
-        model_dir = pathlib.Path(os.environ["MODEL_STORAGE"])
-        model_dir.mkdir(exist_ok=True)  # Do i need to do this every time?
-        model_actor.wrapped_save.remote(model_dir.joinpath(str(model.id)))
+        results_dir = pathlib.Path(os.environ["MODEL_STORAGE"]).joinpath(str(model.id))
+        results_dir.mkdir(parents=True)
```
abdullah-ukaea: mkdir() is called without exist_ok=True, so it crashes if the directory already exists.  Same problem as comment https://github.com/ukaea/toktagger/pull/346/changes#r3705977266 

This is a huge problem as the failure happens after training has completed, so a successful training run is thrown away.

toktagger/api/worker.py
```diff
@@ -110,7 +104,9 @@ def load_model(
     model_actor = get_actor(project=project, model=model, use_gpu=False)
     # Try loading actor with weights file, catch and reraise any errors
     try:
-        load_temp_weights_task = model_actor.wrapped_load.remote(str(weights_path))
+        load_temp_weights_task = model_actor.wrapped_load.remote(
+            weights_path.parent, weights_path.name
+        )
```
abdullah-ukaea, line 110: Can we check this?

results_dir is meant to be a folder in the model cache, but here it's whatever folder the user's uploaded file came from (e.g. Downloads). A model that ignores the filename and assumes the default name will look in the wrong place. The arguments are also passed positionally instead of by keyword, unlike the rest of the codebase, making them easy to mix up.

Maybe use keyword arguments (results_dir=..., weights_filename=...) to prevent mix-ups, and make sure a model can't end up looking in the wrong place either by documenting this case or having the base class resolve the path itself.

toktagger/api/routers/models.py
```diff
@@ -137,9 +138,9 @@ async def delete_models(
         )
 
         # And delete file from storage (if it exists - may not if the job failed)
-        config.settings.models.cache_dir.joinpath(f"{model.id}.model").unlink(
-            missing_ok=True
-        )
+        model_dir = config.settings.models.cache_dir.joinpath(f"{model.id}")
+        if model_dir.exists():
+            shutil.rmtree(model_dir)
```
abdullah-ukaea, line 143: Two small things: 

- f"{model.id}" is an unnecessary f-string wrapping a value that's already a string.
- checking exists() then calling rmtree() separately is slightly redundant when it can be done in one safe step.

To fix simplify to shutil.rmtree(model_dir, ignore_errors=True) and drop the f-string.

Note that existing <model_id>.model files from before this change will never be found or cleaned up, so old models silently appear as blank/untrained so maybe document that somewhere?  e.g. changelog

tests/api/routers/test_models.py
abdullah-ukaea, line 750: Can this be updated?

still checks the old <model_id>.model file path, but that path is never used anymore, so the check always passes regardless of what the code actually does

 Update it to check the new folder path (.joinpath(model_id)) so it actually confirms that a failed load doesn't leave a model directory behind.

toktagger/api/models/base.py
```diff
@@ -195,6 +230,7 @@ def split_data(
 
     @abstractmethod
     def define_model(self) -> None:
+        """Define and return your model architecture here."""
         pass
```
abdullah-ukaea: Small nitpick can you double check this type hint?

The return type says -> None, but the docstring says to return the model and the code actually uses the return value so the type hint is just wrong and should be fixed to something like -> typing.Any.

abdullah-ukaea (commented):
Looks good. I think you need to fix the comment I made in https://github.com/ukaea/toktagger/pull/346/changes#r3705864996 before this can be approved. 

the rest are nice to have, nitpicks and some tests that need to be updated. I also noticed some other things which need to be addressed which are not in the git diff.,

abdullah-ukaea:
Can we fix VideoCNN?

VideoCNN save() and load() (lines https://github.com/ukaea/toktagger/blob/00f2b4b4b12f17220f3a0f0195fed23f529f4170/toktagger/api/models/temp.py#L95 and https://github.com/ukaea/toktagger/blob/00f2b4b4b12f17220f3a0f0195fed23f529f4170/toktagger/api/models/temp.py#L98) were never updated to the new signature, so it'll crash with a TypeError the first time anyone trains this example video model.

update save() and load() here to match this branch's changes, the same way disruption.py was updated?

toktagger/api/models/disruption.py
wk9874: wow, very glad you caught that lol

abdullah-ukaea:
Can we add tests for these gaps? 

No test ever loads a file with a name other than weights.model, so the custom filename logic is never actually tested, and no test checks that a model can save more than one file, even though that's the main point of this change.

Relevant file: https://github.com/ukaea/toktagger/blob/wk9874/models/use_dir_storage/tests/models_definitions.py

These lines are the specific problems.

https://github.com/ukaea/toktagger/blob/00f2b4b4b12f17220f3a0f0195fed23f529f4170/tests/models_definitions.py#L47

and

https://github.com/ukaea/toktagger/blob/00f2b4b4b12f17220f3a0f0195fed23f529f4170/tests/models_definitions.py#L106

Could we add a test that loads a weights file with a different name (e.g. my_weights.pt) and checks the right file was used, and have one mock model save a second file alongside its weights, with a test confirming both files survive a save/load cycle?

toktagger/api/worker.py
wk9874: > make sure the save completes successfully before marking training as done.

Have fixed this to check saving works after training, and if not, deletes the results dir which was created

> which later fools the code into thinking weights exist, causing it to crash instead of falling back to a blank model.

I think for this I think it's now ok - inside `get_actor`, it schedules a `wrapped_load` call, but doesnt wait for that to complete, and so it won't throw an exception if it fails to load. IF it *does* fail to load (which it shouldn't now due to the fix above), it wont mark `trained` as True, and so the model being used is functionally a new model anyway which needs training before usage

toktagger/api/models/base.py
wk9874: > Also have wrapped_save create the directory (mkdir(parents=True, exist_ok=True)) before calling save(),

Done this

> The default filename "weights.model" and the fallback logic (use the given filename, otherwise use the default) are copy pasted into every model file instead of living in one place

This is just because this is the fallback logic my model(s) happen to use. Each other model may want to handle this differently - ie for YOLO models, the `best.pt`, `last.pt`, and YAML config file would be saved to the results dir in the case that there is no filename provided. If there *is* a filename provided, that should be loaded into the YOLO model as the weights file to use

toktagger/api/worker.py
wk9874: Taken care of in base dir

(There *should never be an existing results dir with this model's ID, since they should be unique)

toktagger/api/worker.py
wk9874: > use keyword arguments (results_dir=..., weights_filename=...) 

Done this

> results_dir is meant to be a folder in the model cache, but here it's whatever folder the user's uploaded file came from (e.g. Downloads)

That's correct, this function is used for loading pretrained weights from the user's system. So on the UI, they'll provide a path to a file on their system which they want to load, and this will be used to try and load it into the model. If that is successful, then the loaded model will be saved back into the model cache with the model ID, so that future calls can access it normally

> A model that ignores the filename and assumes the default name will look in the wrong place.

Yeah this is a bit of an issue, I've documented in the docstring of the model base class that if the filename is provided it should take precedence. I'm not sure what else I can do though to enforce this...

toktagger/api/routers/models.py
wk9874: > f"{model.id}" is an unnecessary f-string wrapping a value that's already a string.

Done

> checking exists() then calling rmtree() separately is slightly redundant when it can be done in one safe step.

I think I prefer the way it is, since it won't suppress unexpected errors, while `ignore_errors=True` would

> Note that existing <model_id>.model files from before this change will never be found or cleaned up

True, will add to changelog when we next do a release

wk9874:
> Could we add a test that loads a weights file with a different name (e.g. my_weights.pt) and checks the right file was used

This is covered by `test_model_load_local`

> have one mock model save a second file alongside its weights, with a test confirming both files survive a save/load cycle?
Added to `test_model_start_training_params`

Will add further tests as part of #319 after this PR is merged, since in that PR I make the functions in worker.py bare functions rather than ray remotes, making them easier to test

toktagger/api/worker.py
abdullah-ukaea: sounds good

toktagger/api/models/base.py
abdullah-ukaea: make sense

toktagger/api/worker.py
abdullah-ukaea: Also not sure but that's fine for now since it fails loudly with FileNotFoundError rather than silently loading the wrong fil  the user's upload folder is very unlikely to coincidentally already contain a file called weights.model, so a model that ignores the filename will just error out, not misbehave quietly.

abdullah-ukaea (approved):
ready to be merged, LGTM, nice work on this one.

PR #349, opened 2026-08-04
Profile 2D: reintroduce the spectrogram view as profile 2d [minimal implementation]

abdullah-ukaea:
This pull request introduces the ability to annotate different types of 2D profiles with bounding box and polygon annotations.

- Added the ability to create bounding box and polygon annotations
- Added the ability to manipulate (delete, change type of) annotations on the profile 2d view
- Refactored the threshold annotator to output lists of PolygonAnnotation objects instead of masks

This PR closes https://github.com/ukaea/toktagger/pull/195 as it implements the same core functionality without the stale git history and redundant changes that PR introduced.

Migration Note: The Task enum value changed from spectrogram to profile-2d, so existing projects created with the old value will not load.


abdullah-ukaea:
To Do:

Functional review
- [x] UI developer testing (@jblake42 )
- [x] Backend developer testing (@wk9874 )

Code review
- [x] UI code review in toktagger/ui (@jblake42 )
- [x] Backend code review in toktagger/api (@wk9874 )
- [x] Docs review in docs/ by anyone
- [x] Testing review in tests/ by anyone

**First wave of code review is now done!**

jblake42:
Functionality looks good from my point of view - it aligns to the time series tooling as expected and generation of polygons trough thresholding also works well.

My one usability comment is that the data value popup makes annotation very difficult as it gets in the way. I personally think this should be inhibited at least when drawing is taking place (e.g. ctrl is pressed).

abdullah-ukaea:
> Functionality looks good from my point of view - it aligns to the time series tooling as expected and generation of polygons trough thresholding also works well.
> 
> My one usability comment is that the data value popup makes annotation very difficult as it gets in the way. I personally think this should be inhibited at least when drawing is taking place (e.g. ctrl is pressed).

Done in commit https://github.com/ukaea/toktagger/commit/12d39796acae14ca1051a8fb83aae222e3c60675 !

toktagger/ui/src/app/components/annotators/thresholding.tsx
```diff
@@ -35,109 +70,164 @@ export default function SpectrogramThresholdTool({
...
       return newValue;
     });
   };
 
+  // The range of interest defaults to the full extent of the loaded profile.
   useEffect(() => {
+    if (!isEnabled) return;
+    if (!data || !Profile2DDataSchema.safeParse(data).success) return;
+
+    const profile = Profile2DDataSchema.parse(data);
```
jblake42: From an efficiency point of view we should not be parsing twice - I'd suggest something like this:

```
const result = Profile2DDataSchema.safeParse(data);

if (!result.success) return;

const profile = result.data;
```

toktagger/ui/src/app/components/plots/base-plot.tsx
```diff
@@ -268,18 +274,28 @@ export const BaseTimeSeriesPlot = ({
...
-          },
-        };
-        console.log(selection);
-        findSelectedAnnotations(selection);
+        // Plotly keys the range by axis id, so a selection on a subplot other than the
+        // first reports e.g. y2 rather than y. Resolve the keys rather than assuming x/y.
+        // Plotly's own typing declares only x/y, so widen via unknown.
+        const range = eventData.range as unknown as Record<string, number[]>;
+        const xRange =
+          range[Object.keys(range).find((k) => k[0] === "x") ?? ""];
```
jblake42: I think I would prefer `k.startsWith("x")`

toktagger/ui/src/app/components/plots/base-plot.tsx
```diff
@@ -268,18 +274,28 @@ export const BaseTimeSeriesPlot = ({
...
+        const yRange =
+          range[Object.keys(range).find((k) => k[0] === "y") ?? ""];
```
jblake42: As above

toktagger/ui/src/app/profile2d/components/profile2dViewParamsWidget.tsx
jblake42, line 1: I think this probably belongs in the profile2d components folder

toktagger/ui/src/app/utils.tsx
```diff
@@ -216,3 +232,85 @@ export function convertTimeSeriesToRawAnnotations(
...
+  return sums;
+}
+
+// Plotly supports layout.coloraxis, but @types/plotly.js does not declare it.
+type LayoutWithColorAxis = Partial<Plotly.Layout> & {
+  coloraxis?: { colorbar?: Partial<Plotly.ColorBar> };
+};
+
+// Apply the shared dark mode styling to a Plotly layout.
+export const applyGlobalStyle = (layout: Partial<Plotly.Layout>) => {
```
jblake42: This seems to apply to the 2d profile but not the timeseries plotly graph - additionally it seems to require a refreshing of the page to properly enact the changes. If this is an easy fix I think it should go in here, if not please create an issue and it can be addressed in a new PR after

jblake42 (changes requested):
This is looking good from a UI perspective - I have not reviewed the back end due to time (plus it is probably better Sam or Matt do this anyway)

docs/assets/profile-2d.png
wk9874, line 1: This needs to be updated to the new UI

docs/profile_2d.md
```diff
@@ -0,0 +1,300 @@
...
+- Cividis (default)
+- Viridis
+- Plasma
+- Inferno
+- Magma
+- Cividis
+
+### View Parameters
+
+Access view parameters from the toolbar on the right:
```
wk9874: toolbar on the left

docs/profile_2d.md
```diff
@@ -0,0 +1,300 @@
...
+At the top of the interface, you'll find navigation controls to move through your dataset:
+
+- **Previous Button** (◄): Navigate to the previous sample in your project
+- **Next Button** (►): Navigate to the next sample in your project
+- **Save Button**: Save your current annotations
+- **Delete Button**: Remove selected annotations
+
+**Keyboard Shortcuts:**
+
+- `Shift + ←`: Navigate to previous sample
```
wk9874, line 60: Dont think this is supported

wk9874: (?)

docs/profile_2d.md
```diff
@@ -0,0 +1,300 @@
...
+  the delete control
+
+### Thresholded Regions
+
+The thresholding tool generates polygon annotations automatically, outlining the
+regions whose values exceed a chosen percentile.
+
+**To generate them:**
+
+1. Open the "Threshold" panel in the right toolbar
```
wk9874: left toolbar

docs/profile_2d.md
```diff
@@ -0,0 +1,300 @@
...
+3. For drawing-based annotations, delete and recreate with the desired category
+
+### Deleting Annotations
+
+**For Zones and VSpans:**
+
+1. Right-click on the annotation
+2. Select "Delete" from the context menu
+
+**For Polygons and Bounding Boxes:**
```
wk9874: These are now the same instructions, just combine into one :)

tests/end_to_end/test_profile2d_page.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def reload(project_id=project_id, sample_id=sample_id):
+        page.goto(f"http://localhost:8002/ui/projects/{project_id}/samples/{sample_id}")
+
+    return (project_id, sample_id, reload)
+
+
+def get_x_range(page: Page) -> list[float]:
+    """Read the live Plotly x-axis range from the rendered plot."""
+    return page.evaluate(
```
wk9874: This looks weird - what is this doing? Is there not a way to get this information with built in playwright functions?

tests/end_to_end/test_profile2d_page.py
```diff
@@ -0,0 +1,251 @@
...
+    page.wait_for_timeout(400)
+    zoomed_out = get_x_range(page)
+    zoomed_out_width = zoomed_out[1] - zoomed_out[0]
+    assert zoomed_out_width > zoomed_in_width, "wheel down should zoom the x-axis out"
+
+
+def _annotation_shape_count(page: Page) -> int:
+    """Count the D3 annotation shapes drawn on the heatmap subplot overlay."""
+    return page.evaluate(
+        """(plotId) => {
```
wk9874: Again I'm sure there will be better ways to do this, eg in the timeseries e2e tests we count the vspans or points with:

```
    # One vspan and 3 zones visible
    expect(page.get_by_label("time-point").first).to_be_visible()
    expect(page.get_by_label("time-zone", exact=True)).to_have_count(3)
```

tests/end_to_end/test_profile2d_page.py
```diff
@@ -0,0 +1,251 @@
...
+            return overlay.querySelectorAll("polygon, path").length;
+        }""",
+        PLOT_ID,
+    )
+
+
+def _annotation_mark_count(page: Page) -> int:
+    """Count committed annotation marks (bounding boxes + polygons) on the overlay."""
+    return page.evaluate(
+        """(plotId) => {
```
wk9874: Same comment as above

tests/end_to_end/test_profile2d_page.py
```diff
@@ -0,0 +1,251 @@
...
+
+    add_bounding_box(page)
+    page.wait_for_timeout(500)
+    manual = _annotation_mark_count(page)
+    assert manual >= 1, "manual bounding box should be drawn"
+
+    # Enabling thresholding adds polygons but must keep the manual annotation.
+    page.get_by_role("button", name="Threshold").click()
+    page.get_by_role("switch", name="Thresholding").click(force=True)
+    page.wait_for_timeout(2500)
```
wk9874: Ideally we should avoid timeouts if possible, as these can be flakey on the CI. Would prefer to do a `expect(page.get_by_role(...)).to_be_visible)` or similar on something you expect to pop up once the thresholding is done (but sometimes this may not be possible!)

toktagger/api/core/annotators.py
```diff
@@ -141,18 +150,57 @@ def downsample_time_series(
...
+
+
+def compute_stft(
+    data: TimeSeriesData, nfft: int = 256, nperseg: int = 256, noverlap: int = 128
+) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
     time = np.array(data.time)
     values = np.array(data.values)
 
-    sample_rate = 1 / (time[1] - time[0])
+    sample_rate = 1 / np.abs(np.median(np.diff(time)))
```
wk9874: Could this give a div by zero in some situations?

toktagger/api/core/annotators.py
```diff
@@ -548,50 +596,111 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
-            created_by=AnnotatorTypes.SPECTROGRAM_THRESHOLD,
-        )
+        if isinstance(signal, TimeSeriesData):
+            dim_1, time, values = compute_stft(signal)
+        elif isinstance(signal, Profile2DData):
+            dim_1 = np.array(signal.dim_1)
+            time = np.array(signal.time)
+            values = np.array(signal.values).T
+
+        values = np.nan_to_num(values, 1e-6).clip(1e-6)
```
wk9874: Second argument to `nan_to_num` is `copy: bool`: https://numpy.org/doc/2.3/reference/generated/numpy.nan_to_num.html

Did you mean `np.nan_to_num(values, nan=1e-6).clip(1e-6)`?

toktagger/api/core/annotators.py
```diff
@@ -548,50 +596,111 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
-        threshold_value = np.percentile(values, self.params.percentile)
-        threshold_mask = values > threshold_value
-        return SpectrogramMask(
-            label="SpectrogramMask",
-            values=threshold_mask.tolist(),
-            created_by=AnnotatorTypes.SPECTROGRAM_THRESHOLD,
-        )
+        if isinstance(signal, TimeSeriesData):
+            dim_1, time, values = compute_stft(signal)
+        elif isinstance(signal, Profile2DData):
```
wk9874, line 647: Should we add a `else:` for if it is not an expected type, maybe raising an error?

toktagger/api/core/annotators.py
```diff
@@ -548,50 +596,111 @@ def predict(self, data: MultiVariateTimeSeriesData) -> list[TimeRegion]:
...
+                        np.linspace(0, len(contour) - 1, self.params.max_polygon_points)
+                    ).astype(int)
+                    contour = contour[indices]
+                # Contour coordinates are (row, col), convert to (x, y)
+                contour_coords = [
+                    (int(np.round(c[1])), int(np.round(c[0]))) for c in contour
+                ]
+                contour_coords = [(time[x], dim_1[y]) for x, y in contour_coords]
+                poly = Polygon(contour_coords)
+                # if poly.is_valid and poly.area > 0:
```
wk9874: Should this be commented out?

wk9874 (changes requested):

wk9874:
Plus a comment from functional testing - it's not ideal that the way you close a polygon in video versus profile-2D is different:
* Profile 2D: Click back on starting point
* Video: Double click on the final point

Can we make it so that the profile 2D UI also requires double click to close polygon?

wk9874:
Also can view parameters such as the signal being viewed, colormap, whether to apply log scaling etc be persisted over refresh / moving between samples?

toktagger/ui/src/app/components/annotators/thresholding.tsx
abdullah-ukaea: good catch, fixed in commit https://github.com/ukaea/toktagger/commit/cc989dcb2eaaff4c8fafc2c195a14f5c98496496

toktagger/ui/src/app/components/plots/base-plot.tsx
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/840c39c9f7d8c7b0c19c990c33810b5d93eed04c

toktagger/ui/src/app/components/plots/base-plot.tsx
abdullah-ukaea: also fixed in same commit as above https://github.com/ukaea/toktagger/commit/840c39c9f7d8c7b0c19c990c33810b5d93eed04c

toktagger/ui/src/app/profile2d/components/profile2dViewParamsWidget.tsx
abdullah-ukaea: agreed, moved in commit https://github.com/ukaea/toktagger/commit/81afaa8e0d8ab30b565a1eb676c3e6b7858a8260

toktagger/ui/src/app/utils.tsx
abdullah-ukaea: I looked into this and it wasn't too difficult. So I made the changes now in this PR, I didn't think the changes justified an entire issue and separate PR to be created.

I updated profiles2d and utils in this commit https://github.com/ukaea/toktagger/commit/48d815c32ef9fc9a83e08c7bfd0e5a3a24aa60ef

Now while this PR is focused on profiles2D I also fixed it in time series view in this commit https://github.com/ukaea/toktagger/commit/6672aafac8f664d59c32e596011845fbe6bbaead let me know what you think.

toktagger/api/core/annotators.py
abdullah-ukaea: Good catch ! Duplicate/repeated timestamps (or fewer than 2 samples) made the median time step zero, and numpy divided silently instead of raising, so it only actually
crashed several lines later with a confusing OverflowError.
 
Added an explicit check that raises a clearer ValueError instead. Fixed in https://github.com/ukaea/toktagger/commit/1622ac4de52630b77d0e76f9a04c56daefc7456c

toktagger/api/core/annotators.py
abdullah-ukaea: yep good catch, fixed in commit https://github.com/ukaea/toktagger/commit/adfe2e86a47491a0df484208eac6dc325584e347

toktagger/api/core/annotators.py
abdullah-ukaea: yep added an else statement that raises a runtime error in commit https://github.com/ukaea/toktagger/commit/9b711d2a4d58bd163e629bdebec0ebc0b03c646b

toktagger/api/core/annotators.py
abdullah-ukaea: Good spot, just copied it over from the original profile2D branch, think this was commented out during debugging if I had to guess.

Without the check every polygon is kept, including broken ones. But with it we only keep valid, non-zero area shapes.

I readded it because rounding contour points to pixels regularly produces self intersecting or zero area polygons that the schema doesn't catch.

fixed in commit https://github.com/ukaea/toktagger/commit/0976f76f8ac78093a079e5232b2ae5d6c9264812

abdullah-ukaea:
> Also can view parameters such as the signal being viewed, colormap, whether to apply log scaling etc be persisted over refresh / moving between samples?

done in commit https://github.com/ukaea/toktagger/commit/f3091d4bb2ef22c85a5ecc65da06937106b692ea @jblake42 can you double check this code change is correct as it touches ui code

abdullah-ukaea:
> Plus a comment from functional testing - it's not ideal that the way you close a polygon in video versus profile-2D is different:
> 
> * Profile 2D: Click back on starting point
> * Video: Double click on the final point
> 
> Can we make it so that the profile 2D UI also requires double click to close polygon?

agreed, we will create and issue and tackle this in a different PR. As the polygon tool is a generic component not related to profile 2D view. Profile 2D just simply loads the tool.

docs/profile_2d.md
abdullah-ukaea: I also was surprised but I just tried and it worked so.....

docs/profile_2d.md
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/5f22cf97dda00b2629ceec9de5dfe2b8ca745c22

docs/profile_2d.md
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/5f22cf97dda00b2629ceec9de5dfe2b8ca745c22

docs/profile_2d.md
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/5f22cf97dda00b2629ceec9de5dfe2b8ca745c22

docs/assets/profile-2d.png
abdullah-ukaea: updated in commit https://github.com/ukaea/toktagger/commit/912175ec7787f7f9e8776e5cfea5cb0c32d8d83c

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: Updated in commit https://github.com/ukaea/toktagger/commit/5a7b061ea7a365815c8d924faadeb0fc131e1749 swapped all the custom page.evaluate counting and fixed timeouts, same as the time-series tests use. Let me know if this is closer to what you had in mind!

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/5a7b061ea7a365815c8d924faadeb0fc131e1749

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/5a7b061ea7a365815c8d924faadeb0fc131e1749

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: refactored the end to end test https://github.com/ukaea/toktagger/commit/5a7b061ea7a365815c8d924faadeb0fc131e1749

abdullah-ukaea:
Ready for second round of code review

- [x] UI @jblake42 
- [x] Backend @wk9874 
- [x] Docs (anyone)
- [x] Tests (anyone)

toktagger/ui/src/app/contexts/SampleContext.tsx
jblake42, line 1: Making use of the session is a good solution - I would prefer us to have a context that persists enough for us to use that rather than rely on the session as it is far easier to keep that type safe and understand what is going on. No need to change anything currently

toktagger/ui/src/app/profile2d/components/profile2dViewParamsWidget.tsx
```diff
@@ -6,9 +6,20 @@ import { ComboBox, Flex, Item, Switch } from "@adobe/react-spectrum";
...
 export function Profile2DViewParamsWidget() {
-  const { sample, setViewParams } = useSample();
-  const [selectedSignal, setSelectedSignal] = useState<string | null>(null);
-  const [logScale, setLogScale] = useState<boolean>(false);
+  const { sample, viewParams, setViewParams } = useSample();
+  // Seed from context (which may already hold a restored value - see
+  // SampleContext's view params persistence) rather than always starting blank,
+  // otherwise the effect below would immediately overwrite it with the default.
+  const initialParams =
+    viewParams.name === "profile_2d"
```
jblake42, line 12: I do have concerns about us using this strings to provide logic - not just limited to 2D profiles. We probably should enumerate these to allow them to be changed if need in the future. Added issue

jblake42 (approved):
All looks good. 

As a general request, could you look to slim down comments in the future as much as possible. While it is beneficial to use comments for complex or obscure code it can be a code smell as we should be defaulting to using descriptive code rather than comment essays. Numerous, multi-line comments also have the opposite effect and can make code harder to follow.

I have a feeling this could be AI-related as agents tend to go a bit crazy with comments, but I could be wrong.

samueljackson92:
Amazing work all. I will give this a test next week! 🚀

abdullah-ukaea:
> All looks good.
> 
> As a general request, could you look to slim down comments in the future as much as possible. While it is beneficial to use comments for complex or obscure code it can be a code smell as we should be defaulting to using descriptive code rather than comment essays. Numerous, multi-line comments also have the opposite effect and can make code harder to follow.
> 
> I have a feeling this could be AI-related as agents tend to go a bit crazy with comments, but I could be wrong.

Thank you for the review and spotting the build up of multi line comments introduced by this PR. That is for sure not something I want to introduce into the dev branch of toktagger for the reasons you mentioned. So in commit https://github.com/ukaea/toktagger/commit/2986011157ee1a8c43de65b822fc6a559ef6f0ab I combed through all the multi line comments that this PR touches and changed them to 1 or 2 lines. 

Not a single line of code was touched just the comments @jblake42

docs/profile_2d.md
```diff
@@ -0,0 +1,293 @@
...
+The main visualization consists of two vertically stacked plots:
+
+- **Heatmap Plot (Top)**: Displays the 2D data with color intensity representing signal amplitude or power. The color scale automatically adjusts based on the data range and can be toggled between linear and logarithmic scaling.
+- **Integrated Plot (Bottom)**: Shows the time-integrated values, summed over the spatial dimension. This 1D plot helps identify significant temporal events that may be less obvious in the 2D view.
+
+Both plots share a common time axis, enabling synchronized navigation and analysis.
+
+### Color Scale
+
+The heatmap uses a continuous color scale to represent data values. The colorbar on the right side shows the mapping between colors and values:
```
wk9874, line 28: left side

(sorry missed one in my first review)

docs/profile_2d.md
```diff
@@ -0,0 +1,293 @@
...
+- **Log Scale Toggle**: Switch between linear and logarithmic color scaling for better visualization of different dynamic ranges
+
+### Navigation Controls
+
+At the top of the interface, you'll find navigation controls to move through your dataset:
+
+- **Previous Button** (◄): Navigate to the previous sample in your project
+- **Next Button** (►): Navigate to the next sample in your project
+- **Save Button**: Save your current annotations
+- **Delete Button**: Remove selected annotations
```
wk9874, line 56: 'Clear button'

docs/profile_2d.md
wk9874: Oh yeah you're right, I was trying shift and the + key

Maybe worth making it a bit clearer here, like `Shift` & `←` ?

docs/profile_2d.md
```diff
@@ -0,0 +1,293 @@
...
+
+- `Shift + ←`: Navigate to previous sample
+- `Shift + →`: Navigate to next sample
+
+### Plot Toolbar
+
+The Plotly toolbar at the top of the plot provides navigation tools:
+
+- **Zoom**: Box zoom to a region of interest
+- **Select**: Rubber-band select to pick annotations
```
wk9874, line 68: This is now Box Select, not rubber band select

docs/profile_2d.md
```diff
@@ -0,0 +1,293 @@
...
+1. Open the "Threshold" panel in the left toolbar
+2. Enable the tool with the **Thresholding** switch
+3. Adjust the parameters (see
+   [Automated Annotation Tools](#automated-annotation-tools) below)
+
+Results refresh automatically whenever a parameter changes - there is no apply
+step. The heatmap is faded while the tool is active so the generated outlines
+stand out against the data.
+
+**Visual appearance:** The regions appear as ordinary polygon annotations, and can
```
wk9874, line 194: Just a note - these polygons generated by the thresholding tool don't show up in the table at the bottom of the page until a user attaches a label to them. This may be intentional? Or should they show up in the bottom table with an 'unknown' label or something:?

docs/profile_2d.md
```diff
@@ -0,0 +1,293 @@
...
+be edited, relabelled and deleted exactly like hand-drawn ones.
+
+## Modifying Annotations
+
+### Selecting Annotations
```
wk9874, line 199: Should mention the box select tool in this section

docs/profile_2d.md
```diff
@@ -0,0 +1,293 @@
...
+1. Right-click on the annotation (for Zones and VSpans)
+2. Select "Change Type" and choose the new category
+3. For drawing-based annotations, delete and recreate with the desired category
+
+### Deleting Annotations
+
+1. Right-click on the annotation to open the context menu
+2. Select "Delete"
+
+Alternatively, select any annotation and click the Delete button in the navigation controls.
```
wk9874, line 232: Don't think this button exists anymore

tests/end_to_end/test_profile2d_page.py
```diff
@@ -0,0 +1,142 @@
...
+    expect(page.get_by_label("polygon").first).to_be_visible()
+    expect(page.get_by_label("bounding-box")).to_have_count(1)
+
+    # Disabling removes the generated polygons but keeps the manual annotation.
+    page.get_by_role("switch", name="Thresholding").click(force=True)
+    expect(page.get_by_label("polygon")).to_have_count(0)
+    expect(page.get_by_label("bounding-box")).to_have_count(1)
+
+
+def test_profile2d_threshold_toggle_removes_annotations(server_setup, page: Page):
```
wk9874, line 98: I think this test is redundant, and full covered by the test above

tests/end_to_end/test_profile2d_page.py
```diff
@@ -0,0 +1,142 @@
+import pathlib
```
wk9874, line 1: Very nicely refactored tests, much better than before!

I think we may want to add tests for the following cases:

1. Check you can manually draw time region, time point, and that saving sends them to the backend as correct annotation types
2. Annotations cannot be drawn in View Mode
3. Annotations cannot be selected / deleted in View mode
4. Annotations can have their labels changed & can be deleted in Edit mode
5. Should check that annotations show up correctly in table at bottom of page (this can just be added to an existing test)

toktagger/api/core/data_loaders.py
```diff
@@ -53,21 +54,25 @@ class DataLoaderError(Exception):
...
+    def __init__(self):
+        pass
+
     @classmethod
     @abstractmethod
     def sample_data_type(
         cls,
-    ) -> Type[DataTypes]:
+    ) -> Type[
+        ShotData | ImageFileData | FileData | TimeSeriesFileData | ImageArrayFileData
```
wk9874, line 65: Why change this away from the `DataTypes` union?

wk9874 (changes requested):
Sorry, couple more small docs / test recommendations :)

wk9874:
> > Plus a comment from functional testing - it's not ideal that the way you close a polygon in video versus profile-2D is different:
> > 
> > * Profile 2D: Click back on starting point
> > * Video: Double click on the final point
> > 
> > Can we make it so that the profile 2D UI also requires double click to close polygon?
> 
> agreed, we will create and issue and tackle this in a different PR. As the polygon tool is a generic component not related to profile 2D view. Profile 2D just simply loads the tool.

Raised #360

docs/profile_2d.md
abdullah-ukaea: cleaned up docs in commit https://github.com/ukaea/toktagger/commit/604e32510bacea3bd243690ad047061c8639660c

docs/profile_2d.md
abdullah-ukaea: cleaned up docs in commit https://github.com/ukaea/toktagger/commit/604e32510bacea3bd243690ad047061c8639660c

docs/profile_2d.md
abdullah-ukaea: good catch, updated that line in commit https://github.com/ukaea/toktagger/commit/604e32510bacea3bd243690ad047061c8639660c

docs/profile_2d.md
abdullah-ukaea: done in commit https://github.com/ukaea/toktagger/commit/604e32510bacea3bd243690ad047061c8639660c

docs/profile_2d.md
abdullah-ukaea: yep good catch, done in commit https://github.com/ukaea/toktagger/commit/604e32510bacea3bd243690ad047061c8639660c

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: agreed, removed this test in commit https://github.com/ukaea/toktagger/commit/0252c137c69ff24eff5648a9100e1582f4c658f1

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: I have tried to do this in commit https://github.com/ukaea/toktagger/commit/0252c137c69ff24eff5648a9100e1582f4c658f1 let me know if this looks okay or if I need to change something.

toktagger/api/core/data_loaders.py
abdullah-ukaea: Okay so this is one more redundant change I copied over from the original branch. Good catch, I also noticed a few other things which didn't need changing in the backend. Fixed in commit https://github.com/ukaea/toktagger/commit/5aa6c290a9d9e8a212beed67ccb651711545435c

abdullah-ukaea: backend diff should look more cleaner now I think, please double check

docs/profile_2d.md
abdullah-ukaea: Yep from my understanding in Sam's branch the thresholding tool created annotations did not show up in the annotation table but they are sent to the backend and you can export them. I think the only caveat is that they do not have a class label? but they do have a created by tag. I kept that same behaviour in this branch.

I was not sure how to approach that in this branch, @samueljackson92 could you let me know what your thoughts are on this?

samueljackson92: Yes I believe the threshold tool created annotations without a class label or "unknown" class label. I think in future (different PR) maybe we should give the user an option to select as part of the annotation tool (the same change needs to be made for all annotators, .e.g in time series view as well). 

To be clear: The polygons returned from the thresholding tool should be visible in the annotations table, with class label "unknown".

toktagger/api/core/annotators.py
```diff
@@ -169,24 +169,7 @@ def _coords_to_flat_list(coords: Iterable[Tuple[float, float]]) -> List[float]:
     return flat
 
 
-def smooth_binary_mask(mask: np.ndarray, radius: int = 3) -> np.ndarray:
```
wk9874: Agree with removing this since it's unused, but might want to check with @samueljackson92 if that's intentional? Or if this is meant to be called somewhere?

toktagger/api/core/data_loaders.py
```diff
@@ -54,25 +56,21 @@ class DataLoaderError(Exception):
...
         pass
 
     @abstractmethod
     def get_sample(
         self,
         sample: Sample,
+        params: DataParamTypes = DataParams(),
         **kwargs,
-    ) -> DataResponseType:
+    ) -> Data:
```
wk9874: This should probably still be DataResponseType since we wouldn't support any user defined data class inheriting from `Data`

wk9874 (changes requested):

docs/profile_2d.md
samueljackson92: Added issue: https://github.com/ukaea/toktagger/issues/369 to follow up on this.

tests/end_to_end/test_profile2d_page.py
```diff
@@ -68,6 +69,37 @@ def add_bounding_box(page: Page, label: str = "NTM", offset: int = 150) -> None:
...
+    page.mouse.move(start_x, start_y)
+    page.keyboard.down("Control")
+    page.mouse.down()
+    page.mouse.move(start_x + offset, start_y, steps=20)
+    page.mouse.up()
+    page.keyboard.up("Control")
+
+    page.get_by_role("button", name="Edit Mode").click()
+
+    time.sleep(1)
```
wk9874: Shouldn't use time.sleep() in playwright tests, can either wait for something to be visible (expect(page.get_by_role(...)).to_be_visible()) - which is the preferred way, or use playwright.wait_for_timeout(...) (discouraged, also flaky)

wk9874: Alternatively in this case, do neither of these in this function itself, and just make sure theres some kind of expect(...) after the function is called in the test (since the number of time regions existing after calling this func may be different on a test to test basis)

tests/end_to_end/test_profile2d_page.py
```diff
@@ -140,3 +158,99 @@ def test_profile2d_saved_threshold_annotations_persist(server_setup, page: Page)
...
+def test_profile2d_tools_disabled_in_view_mode(server_setup, page: Page):
+    """Annotation tools can't be activated - and so nothing can be drawn - in View Mode."""
+    setup_project(page)
+
+    expect(page.get_by_role("button", name="View Mode")).to_be_enabled()
+    expect(page.get_by_role("button", name="TIME REGION")).to_be_disabled()
+    expect(page.get_by_role("button", name="TIME POINT")).to_be_disabled()
+    expect(page.get_by_role("button", name="BOUNDING BOX")).to_be_disabled()
+    expect(page.get_by_role("button", name="POLYGON")).to_be_disabled()
+
```
wk9874, line 173: May also want to try adding an annotation anyway by holding control and dragging, and checking nothing is drawn, and that the popup telling you to enter edit mode shows up

tests/end_to_end/test_profile2d_page.py
```diff
@@ -140,3 +158,99 @@ def test_profile2d_saved_threshold_annotations_persist(server_setup, page: Page)
...
+
+def test_profile2d_save_time_annotations(server_setup, page: Page):
+    """Manually drawn time regions/points save as the correct annotation types, and
+    show up in the annotations table."""
+    project_id, sample_id, _reload = setup_project(page)
+
+    add_time_annotation(page, "TIME REGION", "NTM")
+    add_time_annotation(page, "TIME POINT", "Disruption", offset=-150)
+
```
wk9874, line 213: Eg here, could add expect(page.get_by_label("time-zone")).to_have_count(1), and same for time region

tests/end_to_end/test_profile2d_page.py
```diff
@@ -140,3 +158,99 @@ def test_profile2d_saved_threshold_annotations_persist(server_setup, page: Page)
...
+    time_point = next(a for a in annotations if a["type"] == "time_point")
+    assert time_point["label"] == "Disruption"
+
+
+def test_profile2d_annotations_locked_in_view_mode(server_setup, page: Page):
+    """Selecting or deleting an annotation is an edit-mode-only action."""
+    setup_project(page)
+
+    add_time_annotation(page, "TIME REGION", "NTM")
+    # add_time_annotation leaves us back in View Mode.
```
wk9874, line 247: I'd add a expect(page.get_by_role("button", name="View Mode")).to_be_enabled() here to be safe

tests/end_to_end/test_profile2d_page.py
```diff
@@ -140,3 +158,99 @@ def test_profile2d_saved_threshold_annotations_persist(server_setup, page: Page)
...
+
+    # Annotations have pointer events disabled in View Mode, so a right-click can't
+    # reach one to open its context menu.
+    page.get_by_label("time-zone").first.click(button="right", force=True)
+    expect(page.get_by_role("menuitem", name="Delete")).to_have_count(0)
+    expect(page.get_by_label("time-zone")).to_have_count(1)
+
```
wk9874, line 272: Can you try selecting and dragging it, and check it doesn't move? (Unless thats too difficult to do because there is nothing to select in View mode..)

wk9874 (changes requested):

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/387d1a31674d32ba734b168b0e73bbf7220095ae

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/387d1a31674d32ba734b168b0e73bbf7220095ae

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/387d1a31674d32ba734b168b0e73bbf7220095ae

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/387d1a31674d32ba734b168b0e73bbf7220095ae

tests/end_to_end/test_profile2d_page.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/387d1a31674d32ba734b168b0e73bbf7220095ae

toktagger/api/core/data_loaders.py
abdullah-ukaea: fixed in commit https://github.com/ukaea/toktagger/commit/3eaceb5eeddf549a000814ca2d4f613e5ad2be20

abdullah-ukaea:
pending a functional test from @samueljackson92 and addressing this one review comment https://github.com/ukaea/toktagger/pull/349#discussion_r3747940147 I think we are ready to merge? @wk9874

wk9874 (approved):
Looks good to me! Pending functional test from Sam

samueljackson92:
Gave it a quick test drive with Thomson profiles. Works nicely.

image

Also tested loading spectorgrams over UDA.

LGTM

toktagger/api/core/annotators.py
samueljackson92: Removing this is fine I believe.

Issue #350, opened 2026-08-04
Releasing Ctrl mid drawing leaves a polygon stuck


abdullah-ukaea:
**Describe the bug**

If you let go of Ctrl part way through drawing a polygon, the polygon gets stuck. The last corner keeps following your mouse around the plot and you can't finish the shape. The only way out is to hold Ctrl again and click back on the first corner, or press Esc to throw it away.

This happens in the time series view on dev and in the profile 2D view, as both use the same polygon tool. No other tool is affected, since time regions, points and bounding boxes are all drawn in one press-drag-release.

**To Reproduce**

1. Go to a time series project and open a sample (e.g. Local ELM Project)
2. Click View Mode to switch into edit mode
3. Click the POLYGON tool and choose a label from the dropdown
4. Hold Ctrl and click 2–3 points on the plot to start a polygon, but do not click back on the starting vertex to close it
5. Release Ctrl
6. Move the mouse around the plot the polygon's open edge keeps tracking the cursor, and there is no way to finish the shape without pressing Ctrl again or Esc.

**Expected behavior**

Letting go of Ctrl should end the drawing. Happy to discuss which of these we want:

- Option 1: finish the polygon. Keep the corners already placed and drop the two temporary ones. If there are fewer than three real corners it isn't a valid shape, so throw it away. This is what Sam did in slj/profile2d (https://github.com/ukaea/toktagger/blob/slj/profile2d/toktagger/ui/src/app/components/tools/polygon.tsx#L129). Keeps the user's work and handles a slip of the key.
- Option 2 cancel the polygon. Throw the shape away, assuming the user changed their mind. Simpler, but it's the same as Esc, and then the only way to finish a polygon is clicking exactly on the first corner.
- Option 3 just stop it following the cursor. Freeze the shape when Ctrl is released and carry on when it's pressed again. Smallest change, but it leaves ongoingAction set, so it may not fully unstick things.

**Screenshots**

video example of time series view, single plot

https://github.com/user-attachments/assets/a09998ef-5baf-4841-9aeb-2fa7fbba3bcf

video example of time series view, multi plot

https://github.com/user-attachments/assets/ec648e99-34c9-4f5e-befc-bbb708b582b3

abdullah-ukaea:
Closed by https://github.com/ukaea/toktagger/pull/352

Issue #351, opened 2026-08-04
Shot labels are deleted when any annotation is drawn or edited


abdullah-ukaea:
**Describe the bug**
Ticking a shot label and then drawing, moving or deleting any annotation makes the shot label disappear. It's not just a display problem the label is removed from the sample's annotations, so saving writes the loss to the database. Since "Save on Navigate" is on by default, simply moving to the next sample is enough to persist it.

The cause is that shot labels are stored in the same annotation list as time regions, points, boxes and polygons, but the time series view doesn't know how to handle them, and overwrites the whole list whenever anything changes.

**To Reproduce**
1. Open a sample in a time series project (e.g. Local ELM Project)
2. In the Toolbox, open Shot Labels and tick one (e.g. Valid)
3. Click View Mode to switch into edit mode
4. Pick any tool and a label, then Ctrl+drag on the plot to draw one
5. Look back at Shot Labels the tick has gone
6. Click Save, reload the page, and check the sample's annotations. The class_label annotation is no longer there

**Expected behaviour**

Shot labels should survive drawing and editing annotations. More generally, the time series view should only replace the annotations it manages, and leave any other type alone rather than deleting it.

**Screenshots**

https://github.com/user-attachments/assets/ee787881-3eb6-4599-880d-871ad038e122

abdullah-ukaea:
Closed by https://github.com/ukaea/toktagger/pull/352

PR #352, opened 2026-08-04
Time series: bug fixes for polygon and shot label.

abdullah-ukaea:
Fixes two annotation bugs:

- Releasing Ctrl mid polygon now commits the shape (matching how other tools finish on mouse-up) instead of leaving the tool stuck, unless fewer than 3 points were drawn, in which case it's discarded. 
- Ticking a shot label and then drawing/editing/deleting another annotation no longer deletes the shot label, since the time series view now merges its own changes into the annotation list instead of overwriting the whole thing.

Closes issues https://github.com/ukaea/toktagger/issues/350 and https://github.com/ukaea/toktagger/issues/351


jblake42 (approved):
Nice, LGTM

Issue #355, opened 2026-08-05
Improve tool selection in UI


wk9874:
**Is your feature request related to a problem? Please describe.**
In the time series and (newly harmonized) video UI, when you first load into a sample, the tool is unselected. This means if you try to draw, you get a 'select tool' message at the top. 

If you then select a tool, but enter View mode and then go back to Edit mode, it'll have deselected your tool again and you have to select it again. I think this may lead to a lot of redundant clicks & annoy the user

**Describe the solution you'd like**
Previously with video annotation, it loaded with a tool auto selected, and it wasn't possible to have a tool deselected (the buttons acted more like a switch). Personally I think this was preferrable

**Describe alternatives you've considered**
Could leave it as unselected initally when you load the sample with the popup. But the deselection on flicking between Edit & View mode should definitely be changed :)

abdullah-ukaea:
To keep track of which views have had there tool selection fixed:
- [x] Time series (https://github.com/ukaea/toktagger/pull/357)
- [x] Video (https://github.com/ukaea/toktagger/pull/334)
- [x] Profile2D (https://github.com/ukaea/toktagger/pull/349)

PR #357, opened 2026-08-06
Time series: hide polygon/bounding box tools + persist tool selection across mode switches

abdullah-ukaea:
Two small UX fixes to the time series view:

- Hide the D3-based polygon and bounding box tooling from the time series toolbar they have no use case in time series view. Components are left untouched, they'll still to be used by profile2D.
-  Revert the "clear tool on edit/view mode toggle" behaviour introduced in [#317](https://github.com/ukaea/toktagger/pull/317). In practice this caused redundant re-selecting of the tool everytime a user toggled modes, adding friction to annotating. Tools now persist across mode switches (matches pre PR 317 behaviour).

This partially addresses issue https://github.com/ukaea/toktagger/issues/355

https://github.com/user-attachments/assets/483f390b-de71-451f-8918-b05484db3ff7


abdullah-ukaea:
@wk9874 Does this PR do what you expect now for time series?

wk9874:
Yep I definitely prefer this - may be worth checking with @samueljackson92 if he's happy with it? 

I think the selected tool persisting across View/Edit mode toggles will prevent a lot of unnecessary clicks if you're labelling a project which eg only uses time points, otherwise itll get really annoying to have to keep reenabling the same tool!

And I can't think of any case where bounding boxes or polygons would be useful on a time series plot, so makes sense to remove them

jblake42 (approved):
LGTM - pretty sure I saw the tooling change in the profile 2d PR anyway

wk9874 (approved):

PR #367, opened 2026-08-10
Add claude & agents MD

samueljackson92:
As in title.


abdullah-ukaea:
Hi @samueljackson92 

I rewrote the CLAUDE.md and AGENTS.md in this PR as a lot of the content wasn't actually true for our codebase (e.g. it described Next.js and things from other PRs), and both files were too bloated with things like a full directory tree, which Claude's own best practices explicitly say to leave out since it can just read the code itself. 

New versions are shorter (to not bloat the context window before it even starts a task) and only include commands/conventions an agent can't infer on its own, per the linked guides.

I also added a lot of detail on the coding best practices our backend and frontend should actually follow, based on patterns I pulled from past PR reviews.

References: 
- https://code.claude.com/docs/en/best-practices#write-an-effective-claude-md
- https://learn.chatgpt.com/guides/best-practices
- https://learn.chatgpt.com/docs/agent-configuration/agents-md

abdullah-ukaea:
Merge when you are ready @samueljackson92

PR #370, opened 2026-08-10
Profile2D: fix flaky end to end test

abdullah-ukaea:
The test ( test_profile2d_edit_mode_relabel_and_delete) was flaky because Playwright’s physical right-click could be intercepted by Plotly’s overlapping drag layer instead of reaching the annotation.

It now dispatches the context-menu mouse event directly on the annotation SVG, reliably testing the relabel and delete workflow.

For context this test was testing if: Annotations can have their labels changed & can be deleted in Edit mode


abdullah-ukaea:
not merge ready, root cause not fixed

samueljackson92:
Looks like it's passing now. Is this ready for merge?

wk9874:
This has now passed 3 times in a row, so I think we can merge it in, but keep an eye on the e2e tests and will revisit if necessary

wk9874:
(if either @samueljackson92  or @abdullah-ukaea can check the updates I made to the tests and make sure they're happy!)

